PR #7, opened 2016-08-31
German localization & Typos

JamesCullum:
Please verify format, QT crashed and i had to edit it via Notepad++.
Note: Data could be cleaned up, still some CSS and HTML in it and a lot of duplicates


JamesCullum:
Wait, why does it only show one commit? Dang it, i'll try again.

PR #8, opened 2016-08-31
German localization & typos

JamesCullum:
Please verify format, QT crashed and i had to edit it via Notepad++.
Note: Data could be cleaned up, still some CSS and HTML in it and a lot of duplicates

P.S.: Seems like it works now, sorry for the inconvenience


fluffypony:
reviewed

PR #12, opened 2016-09-03
readme: section on installing from package

radfish:
Also, structure readme sections consistently with bitmonerod, to make the concise.
Also, moved license so that packages can fetch and install it directly.


fluffypony:
Reviewed

PR #13, opened 2016-09-03
qmake project: option for static build and libunwind

radfish:
Distribution packages should link dynamically (static libs for dependencies are not available for most packages), so let's add an option for choosing static build when needed but also allowing dynamic build.

Also, libunwind was needed to link on Arch Linux, so added it.


fluffypony:
Reviewed

PR #14, opened 2016-09-03
German translations fix, libunbound link dependency fix, updated MacOS documentation

mbg033:
this PR depends on https://github.com/monero-project/monero/pull/1038

### DIFF
--- a/BasicPanel.qml
+++ b/BasicPanel.qml
@@ -131,7 +131,7 @@ Rectangle {
                 horizontalAlignment: Text.AlignLeft
                 verticalAlignment: Text.AlignBottom
                 color: "#535353"
-                text: qsTr("Availible Balance:")
+                text: qsTr("Available Balance:")
             }
 
             Text {
--- a/README.md
+++ b/README.md
@@ -117,10 +117,33 @@ Process: TODO
 `./release/bin/monero-core`
 
 ### On OS X:
-1. install homebrew
-2. install dependencies: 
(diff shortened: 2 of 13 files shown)


radfish:
After #1040 goes in, please update this PR (#14) to also set INSTALL_VENDORED_LIBUNBOUND=ON in that .sh script, in addition to BUILD_GUI_DEPS=ON.

PR #1040 is necessary to keep this software packageable. Sorry for creating inconvenience here.

mbg033:
I don't have fresh experience with creating packages, just to make sure I understand correctly: is this only for linux where we can (should?) use dynamically linked libs that come from repository rather than build static libunbound from bitmonero, right?

radfish:
Yes, on Linux, it should be possible to build against shared libraries distributed with the distribution instead of against vendored dependencies (that is, library sources copy-pasted into the source tree of the software that depends on those libraries). The Arch Linux package is already built exclusively against shared libraries shipped with the distribution.

Also, vendored dependencies are not going to go away, so let them be used, but do not install them into the system -- that conflicts with distribution's library packages and creates a mess for the user.

Also, if we have to vendor, each software should vendor its own dependencies. One software should not be using dependencies vendored in another software. So, if you need libunbound here, I would vendor it (that is, copy-paste it into this repo). The fact that bitmonerod also happens to depend (and vendor) libunbound is irrelevant.

Thanks.

PR #23, opened 2016-09-19
Bitmonero renamings

mbg033:
changes regarding bitmonero->monero renamings + disabled link to libunwind for Linux build (https://github.com/monero-project/monero/issues/1090).

### DIFF
--- a/get_libwallet_api.sh
+++ b/get_libwallet_api.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 
-BITMONERO_URL=https://github.com/monero-project/bitmonero.git
-BITMONERO_BRANCH=master
+MONERO_URL=https://github.com/monero-project/monero.git
+MONERO_BRANCH=master
 # thanks to SO: http://stackoverflow.com/a/20283965/4118915
 CPU_CORE_COUNT=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.ncpu)
 pushd $(pwd)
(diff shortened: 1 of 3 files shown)


get_libwallet_api.sh
```diff
@@ -1,60 +1,66 @@
...
-popd
+# for Linux, we use libunbound shipped with the system, so we don't need to build it
+
+if [ "$PLATFORM" != "Linux" ]; then
+    echo "Building libunbound..."
+    pushd $MONERO_DIR/build/release/external/unbound
+    make -j$CPU_CORE_COUNT
+    make install -j$CPU_CORE_COUNT
+    popd
+fi
```
radfish: Don't the cmake and make commands above (lines 32-50) already build libunbound? The INSTALL_VENDORED_LIBUNBOUND flag tells cmake to copy libunbound.a to lib/ folder, so it must also build it. Unless I'm missing something, lines 53-63 should not be needed on any platform, no?

get_libwallet_api.sh
```diff
@@ -1,60 +1,66 @@
...
 
 if [ "$(uname)" == "Darwin" ]; then
     # Do something under Mac OS X platform        
-    cmake -D CMAKE_BUILD_TYPE=Release -D STATIC=ON -D BUILD_GUI_DEPS=ON -D CMAKE_INSTALL_PREFIX="$BITMONERO_DIR"  ../..
+    cmake -D CMAKE_BUILD_TYPE=Release -D STATIC=ON -D BUILD_GUI_DEPS=ON -D INSTALL_VENDORED_LIBUNBOUND=ON -D CMAKE_INSTALL_PREFIX="$MONERO_DIR"  ../..
 elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
     # Do something under GNU/Linux platform
-    cmake -D CMAKE_BUILD_TYPE=Release -D STATIC=ON -D BUILD_GUI_DEPS=ON -D CMAKE_INSTALL_PREFIX="$BITMONERO_DIR"  ../..
+    PLATFORM="Linux"
+    cmake -D CMAKE_BUILD_TYPE=Release -D STATIC=ON -D BUILD_GUI_DEPS=ON -D CMAKE_INSTALL_PREFIX="$MONERO_DIR"  ../..
```
radfish: Btw, just to clarify, when I said we should be using distro's packaged libraries, I didn't refer to what this script should do. Any reasonable packaging script for monero-core shouldn't be using this script at all (it should rely on a separate package for monerod and/or libwallet -- which is what Arch set of packages do already). Just wanted to clarify this, in case you have preferences regarding how you want this script to work -- I have zero input with regard to this script.

monero-core.pro
```diff
@@ -86,7 +86,8 @@ linux {
         -lssl \
         -lcrypto \
         -Wl,-Bdynamic \
-        -lunwind \
+        # currently monero has an issue with "static" build and linunwind-dev
+        # -lunwind \
```
radfish: Only for one of the distros, though... I think this will break the Arch package. That was the reason for 010a215. Hm.. ok, how about we add another CONFIG switchs for `libunwind_on` and `libunwind_off`, turn it on by default, so so near the top
```
CONFIG += libunwind_on
```
and then change the above lines to
```
CONFIG(libunwind_off) {
   message(Building without libunwind)
} else {
    message(Building with libunwind)
    LIBS += -Wl,-Bdynamic -lunwind
}
```
It might read a bit oddly with two flags, only because I'm not sure how to do it with one flag. It would need to be set by default, but we must also be able to remove it via command line -- is that possible? (maybe somehow, like `CONFIG -= libunwind_on`). In any case the above should do the job.

Then on Ubuntu, update readme instructions to pass `libunwind_off`when building on Ubuntu -- until this static linking problem is fixed. Distros that are not broken will continue to work by default without any extra flags, as desired.

get_libwallet_api.sh
mbg033: Yes, most likely lines 32-50 will build libunbound (due it's a dependency for some library within libwallet_merged), but it will not copy it into "lib/" until ```make install``` invoked

monero-core.pro
mbg033: added this option and created PR https://github.com/monero-project/monero-core/pull/30

PR #36, opened 2016-10-05
Update README.md with fixed typo in path to QT

dternyak:
There was a typo in the mac instructions, here's the fix.


dternyak:
@fluffypony Bumpity bump. Small fix, could save people building on mac some unnecessary debugging.

fluffypony:
@dternyak PRs need to be reviewed, so I try leaving them open for ~a week so that I'm not the only one reviewing:)

PR #37, opened 2016-10-05
critical: broken wallet after crash

Jaqueeee:
Fixes a critical bug that could leave a wallet in broken state after a crash (god forbid).
The bug only affects newly created wallets where wallet is synced, has transactions but isn't closed properly. When opening the crashed wallet the fast sync method will be used which will miss any transactions made from the wallet and therefore leaving it in broken state. 
Consider moving this logic to libwallet in future.

Fixes https://github.com/mbg033/monero-core/issues/59

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -53,6 +53,7 @@ ApplicationWindow {
     property var transaction;
     property alias password : passwordDialog.password
     property int splashCounter: 0
+    property bool isNewWallet: false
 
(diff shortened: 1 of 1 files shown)


medusadigital:
after creating a fresh wallet i killed the monero-core process to cause wallet corruption. the wallet didnt take any damage and could be opened normally after. no block header scanning happened when reopening the wallet

PR #43, opened 2016-10-10
Continue wallet recover if sync is not finished

Jaqueeee:
To be reviewed by @medusadigital before merge.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -794,4 +794,8 @@ ApplicationWindow {
             }
         }
     }
+    onClosing: {
+        walletManager.closeWallet(currentWallet);
+        console.log("onClosing called");
+    }
 }
--- a/main.qml
+++ b/main.qml
@@ -158,7 +158,6 @@ ApplicationWindow {
             }
 
             console.log("using wizard wallet")
-
             connectWallet(wizard.settings['wallet'])
 
             isNewWallet = true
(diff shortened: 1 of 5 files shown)


main.qml
```diff
@@ -794,4 +794,8 @@ ApplicationWindow {
             }
         }
     }
+    onClosing: {
+        walletManager.closeWallet(currentWallet);
+        console.log("onClosing called");
+    }
```
Jaqueeee: @mbg033: FYI. Noticed that Wallet::~Wallet() wasn't called on some occasions when application was closed. Added these lines to make sure the wallet is saved properly. Have you noticed this behavior too?

mbg033: @Jaqueeee that must be a bug - missing "cleanup" code.  Wallet lifetime managed only by WalletManager, so only the way to properly destruct the wallet object is WalletManager::closeWallet() call. Probably make sense to implement some ref counting within WalletManager and close wallet (wallets) by e.g. QCoreApplication::aboutToQuit() signal

Jaqueeee: @mbg033 yes. that makes sense.
Also this one could probably be solved cleaner. 
https://github.com/monero-project/monero-core/blob/master/main.qml#L147
The previous strict comparison evaluates true when a wallet is opened that has previously crashed or has been killed by the user. Application crashes after password dialog when closeWallet is being called on that wallet.

PR #44, opened 2016-10-10
Critical: restore-height fallback & corrupt wallet fix

Jaqueeee:


medusadigital:
tested both in Jaquees continue-recover branch.

- password protected wallets dont get corrupted anymore if the process crashes
- restore from seed fallback (if no block height is provided) now works again too

PR #45, opened 2016-10-10
History: display blockheight, sort by blockheight

mbg033:


medusadigital:
looks nice here. can you possibly add descending sorting by height by default ? 
anyway, even without it, transaction history seems ready for beta!

mbg033:
set default sorting by blockheight, descending

fluffypony:
Reviewed

PR #46, opened 2016-10-10
various password issues fixed

Jaqueeee:
- store pw in session to be able to view seed on settings page immediately after new wallet is created
- empty pw field after form submission
- strict comparison


medusadigital:
worked fine, tested in Jaquees pw-issues branch

PR #66, opened 2016-10-16
Transfer: disable Send button if the payment id is invalid

moneromooo-monero:
Needs https://github.com/monero-project/monero/pull/1229

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -239,6 +239,19 @@ Rectangle {
         anchors.topMargin: 5
     }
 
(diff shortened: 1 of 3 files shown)


Jaqueeee:
Works with https://github.com/monero-project/monero/pull/1229. But imo we need to inform the user why the button is disabled.  Also needed for other validation failures so i'll create a new issue for that.

moneromooo-monero:
Branch updatd, since it'd conflct. The new commit adds address validation, and depends on https://github.com/monero-project/monero/pull/1230

Issue #62, opened 2016-10-15
Repo improvements


anonimal:
I know we are _all_ pressed for time but I think the organization of this repo could use a little more work:
1. ~~build instructions are out of date (one apparently only needs only run `build.sh`)~~
2. we desperately need a CMake recipe (we don't need all these build scripts)
3. ~~we should consider using a git submodule for the `monero` repo~~
4. ~~these scripts shouldn't be commandeering all of my cpu cores~~ Thanks @moneromooo-monero #106 

These are just a few issues off the top of my head. I look forward to contributing to monero-core so if these issues aren't resolved by then, I'll be happy to make fixes.

moneromooo-monero:
The most trivial of these is fixed :D https://github.com/monero-project/monero-core/pull/106

xmr-eric:
@anonimal Can this issue be closed?

anonimal:
@xmr-eric step 2 is unresolved. I would advise to keep this issue open until that is addressed.

TheCharlatan:
Pull Request for cmake is open: https://github.com/monero-project/monero-gui/pull/2404

selsta:
#2404 is now merged.

+resolved

PR #80, opened 2016-10-20
make build script exit with non-zero status on error

danrmiller:
I'm trying to integrate with some things like buildbot, and there are many different ways the build.sh script errors but still exits with status 0. This should help that a bit by making it exit with a non-zero status when there is an error from any of the commands.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+set -o pipefail
 source ./utils.sh
 pushd $(pwd)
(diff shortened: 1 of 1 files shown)


fluffypony:
Plz rebase

danrmiller:
I'm going to instead look at the approach mentioned in #62 .

anonimal:
Hi @danrmiller, I got caught up with kovri but also wanted to see where Ilya would go with other in-progress build decisions. I'll get back on #62 soon (if no one else gets to it first); probably when I start tackling *builds* in general.

Issue #83, opened 2016-10-23
Possible error in build instructions in ReadMe.md


taushet:
Following the instructions to the letter, this happened:

    minertuxalpha@minertuxalpha-desktop:~/monero-core$ qmake
    Project MESSAGE: Building with libunwind
    WARNING: DESTDIR: Cannot access directory 'release/bin'
    WARNING: langrel: Cannot access directory '/home/minertuxalpha/monero-core/release/bin/translations'
    WARNING: langrel: Cannot access directory '/home/minertuxalpha/monero-core/release/bin/translations'
    WARNING: langrel: Cannot access directory '/home/minertuxalpha/monero-core/release/bin/translations'
    WARNING: langrel: Cannot access directory '/home/minertuxalpha/monero-core/release/bin/translations'
    WARNING: langrel: Cannot access directory '/home/minertuxalpha/monero-core/release/bin/translations'
    Failure to open file: /home/minertuxalpha/monero-core/Makefile
    Unable to generate makefile for: /home/minertuxalpha/monero-core/monero-core.pro

However when  `sudo qmake:`

    minertuxalpha@minertuxalpha-desktop:~/monero-core$ sudo qmake
    Project MESSAGE: Building with libunwind

...it works fine. Just me, or everyone?

jwinterm:
Did you clone the repo as root or something? You should be able to access directories in your home folder.

taushet:
I know it seems strange, but no. I made a new terminal and followed the instructions to the letter. Is there any *downside* to doing it with sudo?

taushet:
Fresh install of Ubuntu 12.04.1 LTS.

mbg033:
There's definitely nothing that requires root privileges.

mbg033:
Also, alternatively you can just invoke build.sh in the fresh (just cloned) project directory

dternyak:
Definitely just you, seems like a permission error. Fix with chmod

PR #84, opened 2016-10-23
Change make and qmake to sudo make and sudo qmake

taushet:
See error #83


fluffypony:
This is definitely non-standard - I'd suggest we figure out why it's not working for you, rather than blanket the build instructions:)

PR #89, opened 2016-10-27
Fix for black window on windows build

Jaqueeee:


Jaqueeee:
@mbg033 
see bottom http://doc.qt.io/qt-5/qml-qtgraphicaleffects-desaturate.html
"Note: It is not supported to let the effect include itself, for instance by setting source to the effect's parent."

Didn't throw any errors but caused a black screen when using ANGLE.

PR #93, opened 2016-10-30
Wallet picker

Jaqueeee:
- Added close wallet button on settings page which fires up the wizard
- added "Open from file" option in wizard
- Changed default wallet path to $HOME[/Documents]/Monero/wallets (Documents on windows)
- Changed default account name to system env user account
- various bug fixes and improvements


medusadigital:
Couldnt test on windows yet, but all in all very fine. 
"opening last used wallet on startup" now also works correctly

thank you!!

PR #94, opened 2016-10-31
Change default mixin to 4 & add mixin # to confirmation window

### LINKED ISSUE #74: Mixin should default to 4
It now defaults to 2. And it seems to be such an ludicrous pain in the butt to initialize stuff to more than a constant that my attempt to scan the index-to-mixin failed, so I'll have to leave that to someone who knows QML.

dEBRUYNE-1:
Closes #74.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -372,11 +372,12 @@ ApplicationWindow {
             // here we show confirmation popup;
 
             transactionConfirmationPopup.title = qsTr("Confirmation") + translationManager.emptyString
-            transactionConfirmationPopup.text  = qsTr("Please confirm transaction:\n\n")
+            transactionConfirmationPopup.text  = qsTr("Please confirm transaction:\n")
                         + qsTr("\nAddress: ") + address
                         + qsTr("\nPayment ID: ") + paymentId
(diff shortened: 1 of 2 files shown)


main.qml
```diff
@@ -372,11 +372,12 @@ ApplicationWindow {
             // here we show confirmation popup;
 
             transactionConfirmationPopup.title = qsTr("Confirmation") + translationManager.emptyString
-            transactionConfirmationPopup.text  = qsTr("Please confirm transaction:\n\n")
+            transactionConfirmationPopup.text  = qsTr("Please confirm transaction:\n")
                         + qsTr("\nAddress: ") + address
                         + qsTr("\nPayment ID: ") + paymentId
-                        + qsTr("\nAmount: ") + walletManager.displayAmount(transaction.amount)
+                        + qsTr("\n\nAmount: ") + walletManager.displayAmount(transaction.amount)
```
dternyak: @dEBRUYNE-1 I think you made a mistake here.

dEBRUYNE-1: That adds a line between Payment ID and Amount, which makes it, in my opinion, more clear. Thus, with this change you have:

Address:
Payment ID:

Amount:
Fee:

Mixin:

PR #95, opened 2016-11-01
Custom password dialog. fixes issue with standard dialog

mbg033:
image
Fixes issue https://github.com/monero-project/monero-core/issues/86

### DIFF
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -77,7 +77,7 @@ Window {
 
             TextField {
                 id : passwordInput
-                focus:true
+                focus: true
                 Layout.fillWidth: true
                 Layout.alignment: Qt.AlignHCenter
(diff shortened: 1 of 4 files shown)


dEBRUYNE-1:
Looks nice!

M5M400:
looking good here!
screenshot from 2016-11-02 09-12-20

would be nice if one could just hit "enter" to confirm password though. Currently you have to confirm using the mouse to click the OK button.

mbg033:
@M5M400 improved, now it accepted with Enter key, rejected with Esc. Also Tab changes the keyboard focus.

taushet:
Works for me. No issues to report, other than that the OK/cancel buttons are in the wrong order by convention (the OK should be on the right)

M5M400:
@mbg033 works for me as well, nice!

medusadigital:
worked here

PR #96, opened 2016-11-01
Change project name to Monero

Jaqueeee:
Fixes https://github.com/monero-project/monero-core/issues/60

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -53,7 +53,7 @@ int main(int argc, char *argv[])
 
     app.setApplicationName("monero-core");
     app.setOrganizationDomain("getmonero.org");
-    app.setOrganizationName("The Monero Project");
+    app.setOrganizationName("Monero");
 
     filter *eventFilter = new filter;
     app.installEventFilter(eventFilter);


fluffypony:
I'm confused about this - organisation name *should* be the Monero Project, and the application is Monero Core. Does this appear somewhere that's making it confusing?

Jaqueeee:
OK. I'm not sure if it appears in any other places except config dir mentioned in https://github.com/monero-project/monero-core/issues/60
Let me see if i can find any alternative ways of setting that dir name without changing the organization name.

Jaqueeee:
Ok. It seems there's no other trivial ways of changing that config folder name. So it's up to you @fluffypony if you wan't to close this and continue using "The Monero Project" as folder name.

fluffypony:
Let's close this for now, it could come in handy later on (having a The Monero Project folder I mean)

moneromooo-monero:
Looking into this, it seems it's Qt being silly.
The "organization" name is to organize config based on who wrote the program, so it'll write ~/.config/organization/program.conf. So orgnization doesn't actually mean author (though they use company names in their examples). Normally you'd have ~/.config/program/X with X being whatever file(s) you use there, but they seem to have this hardcoded to use the above.

peronero:
Updated originating ticket:

https://github.com/monero-project/monero-core/issues/60#issuecomment-258996349

Jaqueeee:
@fluffypony Please reconsider this one.  See https://github.com/monero-project/monero-core/pull/96

PR #100, opened 2016-11-02
Switch button position on password modal

taushet:
Buttons are the wrong way around (against convention)

Not tested.


Jaqueeee:
Nice. But Ilya has implemented new password dialog recently https://github.com/monero-project/monero-core/pull/95

taushet:
Sure, but the point stands.

fluffypony:
You'll have to rebase it against the new dialog box

taushet:
Done (I think...)

PR #103, opened 2016-11-03
Restyles the wizard's finish page

maitscha:
I have restyled the wizard's finish page, as I think the labels should be aligned. For this, I used a `<table>` (see screenshot).

image


medusadigital:
it does look good, but together with this PR https://github.com/monero-project/monero-core/pull/93 there seems to be an issue: the PR from Jaquee removes the wizzard page regarding smart mining/auto donation. both features are not implemented yet and should be set to false per default. (otherwise the users is confused, since he hasnt even been asked if he wants to mine or not). If this should be part of yours or of Jaquees PR im not sure (cant read too much code)

maitscha:
Commented out auto-donation and background-mining in 1036eef.

PR #104, opened 2016-11-03
Fix sort by date

mbg033:
Sort by date actually sorts by datetime, so transactions sorted properly within single day


medusadigital:
the fix works, but the initial sorting is still by Transaction Time and not by timestamp. (first all incoming transactions, later all outgoing transcactions)

taushet:
Works here. I agree with @M5M400, though. Conventions are key, and people get funny when programs act in ways they are not used to.

PR #106, opened 2016-11-05
Allow user to set number of parallel jobs

moneromooo-monero:


taushet:
On a side note, there will be many users who will get quite confused with this sort of thing. Once the dust settles, we (I?) should get the tooltips sorted. Do we want long and informative tooltips or short phrases?

PR #108, opened 2016-11-05
history: support old payment ids and make fields selectable

Jaqueeee:
fixes https://github.com/monero-project/monero-core/issues/107
+ makes paymentid, tx id, block height selectable


medusadigital:
works nice and does the job for now, so 64 char paymentIds are fully visible

PR #109, opened 2016-11-05
Switch OK/Cancel button positions

taushet:
As per conventions, the 'OK' button should be rightmost.


medusadigital:
worked here

PR #110, opened 2016-11-05
Fix erroneous DateTime & small README change

dEBRUYNE-1:
1. `firstDateTime` was set to current date, whereas `lastDateTime` was set to the genesis block. This should be the other way around. [This](https://i.imgur.com/4kwb8Bh.png) is how it looked before the PR. 

2. The GUI connects to mainnet by default. Thus, we should remove the `--rpc-bind-port 38081` (testnet) from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ TODO
 
 7. Before running the GUI, it's recommended that you have the Monero daemon running in the background.
 
-	`./monerod --rpc-bind-port 38081`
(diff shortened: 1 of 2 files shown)


medusadigital:
works fine

moneromooo-monero:
Not my call, but since those two things are totally independent, they should be two separate commits (and the README one should have a commit message that's more explicit than "small README change").

PR #115, opened 2016-11-05
Write support for tx notes

moneromooo-monero:
Support still needs adding for displaying them in the history,
but at least they're saved in the cache now, and not ignored.


moneromooo-monero:
Needs https://github.com/monero-project/monero/pull/1298

medusadigital:
tested, works

PR #117, opened 2016-11-06
Badfields

moneromooo-monero:


pages/Transfer.qml
```diff
@@ -268,7 +282,7 @@ Rectangle {
         shadowPressedColor: "#B32D00"
         releasedColor: "#FF6C3C"
         pressedColor: "#FF4304"
-        enabled : amountLine.text.length > 0 && checkAddressAndPaymentID(addressLine.text.trim(), paymentIdLine.text.trim(), appWindow.persistentSettings.testnet)
+        enabled : checkInformation(amountLine.text, addressLine.text, paymentIdLine.text, appWindow.persistentSettings.testnet)
```
Jaqueeee: Spam alert: Gives "Invalid address format" 6 times in log.

Jaqueeee: 4 of them seems to be from this line. the other 2 is generated from somewhere else.

moneromooo-monero:
Might be because QML idiotically calls the function several times though it's called once. Happened in a previous patch. Alternatively, it's because I don't know how QML works. Or maybe the enabled property is evaluated more than needed. A log in the check function would tell.

moneromooo-monero:
I'm seeing just two such messages when typing in the address field.

Jaqueeee:
I got six of them when starting the app. It's because the enabled state is checked when object is created.  I'm proposing the following change:
https://github.com/Jaqueeee/monero-core/commit/06f2dd59a0fdf8c5ac594a98bdfcb0346b8bbe00

moneromooo-monero:
Wouldn't work on mouse paste. Besides, it's not like it's taking a lot of time, it feels instant here.

Jaqueeee:
Fair enough. I would prefer the fileds not being marked att erroneous on app start. But such ux issues can be fine tuned later on.

moneromooo-monero:
Well, it's a log. People using the GUI are trying to not look at it :P

PR #118, opened 2016-11-06
Fix small amounts from being conflated with zero

moneromooo-monero:
Something like 0.000000000012 was converted to 1.2e-11
by the conversion to double, and that was rejected by
the Cryptonote parser.


fluffypony:
Please rebase

moneromooo-monero:
done

PR #120, opened 2016-11-06
History: Fix filter validation

mbg033:
Date & amount filter validation; Payment ID input validated with hexadecimal regex


fluffypony:
Please rebase

mbg033:
rebased

PR #127, opened 2016-11-07
refresh history after daemon address change

Jaqueeee:
Fixes https://github.com/monero-project/monero-core/issues/126


mbg033:
Thanks, confirmed, can be closed after merged

PR #134, opened 2016-11-08
asynchronous transaction

Jaqueeee:
+ show waiting... message when waiting for transaction to finish. 
fixes https://github.com/monero-project/monero-core/issues/130 and https://github.com/monero-project/monero-core/issues/116

### DIFF
--- a/src/libwalletqt/Wallet.h
+++ b/src/libwalletqt/Wallet.h
@@ -194,7 +194,6 @@ private:
     mutable QTime   m_daemonBlockChainTargetHeightTime;
     mutable quint64 m_daemonBlockChainTargetHeight;
     int     m_daemonBlockChainTargetHeightTtl;
-    QObject * m_walletManager;
 };
 
 
--- a/src/libwalletqt/Wallet.h
+++ b/src/libwalletqt/Wallet.h
@@ -171,7 +171,7 @@ signals:
     void historyModelChanged() const;
 
     // emitted when transaction is created async
-    void transactionCreated(PendingTransaction * transaction);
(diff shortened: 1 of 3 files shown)


src/libwalletqt/Wallet.h
```diff
@@ -175,6 +181,7 @@ class Wallet : public QObject
     mutable QTime   m_daemonBlockChainTargetHeightTime;
     mutable quint64 m_daemonBlockChainTargetHeight;
     int     m_daemonBlockChainTargetHeightTtl;
+    QObject * m_walletManager;
```
mbg033: not clear why do you need this?  WalletManager is a singleton currently, it's not supposed to be deleted. So PendingTransaction will never be deleted from "parent" object.

Jaqueeee: Thanks for feedback. Removed. 
OK. I got the dreaded "QObject: Cannot create children for a parent that is in a different thread." when i used Wallet as parent. Changed to creating PendingTransaction without parent and warning is gone. Since PT is deleted in disposeTransaction() i hope that is an OK solution?

mbg033: Thanks. Yes, thats good, I would even remove second "parent" parameter from PT's ctor

medusadigital:
worked here

fluffypony:
Please rebase

Jaqueeee:
Rebased
Needs rebasing again if you merge https://github.com/monero-project/monero-core/pull/119 https://github.com/monero-project/monero-core/pull/120 https://github.com/monero-project/monero-core/pull/135 @fluffypony 
@moneromooo-monero @mbg033

fluffypony:
@Jaqueeee git can handle most merges, even if multiple commits touched the same file:)

PR #135, opened 2016-11-08
Add a button to sweep unmixable outputs

moneromooo-monero:
Needs https://github.com/monero-project/monero/pull/1310 and https://github.com/monero-project/monero-core/pull/129 and https://github.com/monero-project/monero/pull/1307

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -157,6 +157,7 @@ ApplicationWindow {
             middlePanel.checkPaymentClicked.disconnect(handleCheckPayment);
         }
         middlePanel.paymentClicked.connect(handlePayment);
+        middlePanel.sweepUnmixableClicked.connect(handleSweepUnmixable);
         // basicPanel.paymentClicked.connect(handlePayment);
         middlePanel.checkPaymentClicked.connect(handleCheckPayment);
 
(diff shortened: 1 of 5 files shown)


main.qml
```diff
@@ -414,6 +415,44 @@ ApplicationWindow {
...
+            informationPopup.onCloseCallback = null
+            informationPopup.open();
+            // deleting transaction object, we don't want memleaks
+            currentWallet.disposeTransaction(transaction);
+
+        } else if (transaction.txCount == 0) {
+            informationPopup.title = qsTr("No unmixable outputs to sweep") + translationManager.emptyString
+            informationPopup.text  = qsTr("No unmixable outputs to sweep") + translationManager.emptyString
+            informationPopup.icon = StandardIcon.Information
+            informationPopup.open()
```
Jaqueeee: You need to reset the onCloseCallback here. 
`informationPopup.onCloseCallback = null`

moneromooo-monero:
Thanks, fixed.
Also with sweep_all now.

src/libwalletqt/Wallet.cpp
```diff
@@ -215,6 +215,16 @@ PendingTransaction *Wallet::createTransaction(const QString &dst_addr, const QSt
     return result;
 }
 
+PendingTransaction *Wallet::createTransactionAll(const QString &dst_addr, const QString &payment_id,
+                                                 quint32 mixin_count, PendingTransaction::Priority priority)
+{
+    Bitmonero::PendingTransaction * ptImpl = m_walletImpl->createTransaction(
+                dst_addr.toStdString(), payment_id.toStdString(), boost::optional<uint64_t>(), mixin_count,
```
mbg033: Can it be done without boost? Something like passing 0 here? Initially idea was to not introduce anything except stl into libwallet_api interface. boost currently isn't compile time dependency for GUI (only libs), so at least Windows build will break. Otherwise boost include path needs to be added to monero-core.pro

moneromooo-monero: I don't get that point, since src/wallet/api already uses boost in many places. Why does it make a difference ?
And if it does, passing 0 could work, yes. I just thought it a bit iffy and I remembered boost::optional, but if that's a problem I'll use 0. I'd like to understand why this is a problem even though other code there uses boost first though.

moneromooo-monero: I think I see. You want to be able to build monero-core, but not monero ? as in, copy binary libs from machine to machine ?

fluffypony:
Please rebase

src/libwalletqt/Wallet.cpp
moneromooo-monero: I'll add a poor man's optional implementation. I saw there's one in std:: but it's only from c++17...

moneromooo-monero:
done

src/libwalletqt/Wallet.cpp
mbg033: Yes, I just wanted to keep interface without any additional dependencies until this really needed. src/wallet/api - yes it uses boost, but not in public interface.

fluffypony:
Please rebase...again... (sorry about that:)

moneromooo-monero:
done, still needs 129 though

moneromooo-monero:
rebased, and ported to async txes

PR #138, opened 2016-11-09
Adds openssl and boost lib includes for OSX build

maitscha:
This adds the paths to the openssl and boost libraries (installed with brew) for the MacOS X
build. Furthermore the MacOS X build instructions in the README.md has been improved (formatting).


maitscha:
By the way this hopefully also fixes the broken MacOS build: https://build.getmonero.org/builders/monero-core-osx-10.11/builds/62/steps/compile/logs/stdio

PR #160, opened 2016-11-13
get_libwallet_api.sh: fix version.h generation timeline

moneromooo-monero:


dEBRUYNE-1:
Resolves my compilation issue.

PR #161, opened 2016-11-13
Do not open wallet from default location automatically

Jaqueeee:
Fixes https://github.com/monero-project/monero-core/issues/153

Automatic opening of wallet from default location is not needed anymore when we have the wallet picker. (Note: Not to be mistaken of automatic opening of last opened wallet which is still working)


dEBRUYNE-1:
Can confirm this works.

taushet:
Works here.

PR #163, opened 2016-11-13
Refresh progress ux

Jaqueeee:
- make it possible to close popup and exit program
- ensure correct message is shown in popup


dEBRUYNE-1:
Tested, works properly.

taushet:
Works here.

M5M400:
works, but uses standard window decoration. is there no nicer way?

PR #168, opened 2016-11-15
Rename 'Transfer' tab to 'Send'

xmr-eric:
See issue 159: https://github.com/monero-project/monero-core/issues/159

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -228,7 +228,7 @@ Rectangle {
                 id: transferButton
                 anchors.left: parent.left
                 anchors.right: parent.right
-                text: qsTr("Transfer") + translationManager.emptyString
+                text: qsTr("Send") + translationManager.emptyString
                 symbol: qsTr("T") + translationManager.emptyString
                 dotColor: "#FF6C3C"
                 onClicked: {


expez:
Won't all the translation files have to be updated too?

dternyak:
@expez Yes, but that can be done independently imo.

expez:
@dternyak You can't be serious?  This change is a net loss if it makes a trivial word change but then breaks the wallet (aesthetically if they get no text / error msg on the button, or functionally if the button becomes invisible because the empty string is used) for all other languages.

dternyak:
@expez I wasn't aware that it becomes broken for other languages

dternyak:
@expez Can you confirm that this would still have caused a breaking change to the GUI? It is now merged, so it would be very helpful if you could open an issue if you can produce a bug. I still don't believe that his PR could have caused that, but I am also not able to build the GUI currently.

expez:
I don't know for sure what's going to happen on other languages, but when you ask for a translation and it cannot find one, you probably get one of these as fallbacks:

* Exception is thrown (startup, runtime, compile time?)
* You get the empty string
* You get `NULL`
* You get the default as a fallback.

If I wrote this you'd get the last one, and this will look weird but not be completely broken.  Let's hope they agreed with that fallback.

I don't have time to fix the build and test either.

PR #169, opened 2016-11-15
Add 32/64-bit cache error msg

Jaqueeee:
Fixes https://github.com/monero-project/monero-core/issues/149
Copied error msg from cli-wallet.


Jaqueeee:
Closing until we have a better solution.

PR #183, opened 2016-11-18
Fix mixin being wrong in sweep all transactions

moneromooo-monero:


dEBRUYNE-1:
Tested, works properly now.

medusadigital:
tested, also fixes https://github.com/monero-project/monero-core/issues/182

taushet:
Works here.

PR #186, opened 2016-11-18
Wallet: fix history leak on destruction

moneromooo-monero:


taushet:
As with #187, works here,, but not really sure how to test this change specifically.

PR #187, opened 2016-11-18
MiddlePanel: fix null pointer dereference for history model

moneromooo-monero:


taushet:
Not entirely sure what this actually does/did, but this works here (I just did a few transactions then played around with the panes and everything worked as normal - was there anything specific to test?)

Jaqueeee:
@taushet Nothing specific to test here. Just removes an error from console/log.

PR #188, opened 2016-11-19
Remove useless black bar buttons

moneromooo-monero:
Except the yellow layout change one, which is actually useful

In particular, the close button closed without saving the wallet
cache, contrary to what closing using the window manager close
button did. Which isn't nice if you want to keep your tx keys.

### DIFF
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -150,7 +150,7 @@ Rectangle {
     ColumnLayout {
         anchors.fill: parent
         anchors.margins: 2
-        anchors.topMargin: 30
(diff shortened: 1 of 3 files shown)


fluffypony:
Please rebase

fluffypony:
Also I just realised 189 was closed in favour of this, but then re-opened, so I merged it. My bad...

moneromooo-monero:
I'll wait a bit before doing that, since there was a discussion and people seemed to prefer the other one slightly (even though half the buttons don't work). So I might end up closing it instead.

PR #193, opened 2016-11-21
Adjust height for smaller screens

Jaqueeee:


medusadigital:
tested, does the job

taushet:
Works here.

PR #196, opened 2016-11-23
Daemon progress bar redesign

Jaqueeee:
redesigned to fit smaller screens. 
https://github.com/monero-project/monero-core/pull/193

### DIFF
--- a/components/DaemonProgress.qml
+++ b/components/DaemonProgress.qml
@@ -26,7 +26,7 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-import QtQuick 2.6
+import QtQuick 2.0
 
 Item {
     id: item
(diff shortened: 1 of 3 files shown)


components/DaemonProgress.qml
```diff
@@ -26,13 +26,13 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-import QtQuick 2.0
+import QtQuick 2.6
```
dEBRUYNE-1: @mbg033 can we use anything over 2.0 for QtQuick?

Jaqueeee: 2.6 is available in QT 5.6 and above. Can @mbg033 confirm if QtQuick > 2.0 is an issue?

mbg033:
it depends. if app needs to be built/packaged for e.g. Ubuntu 16.04 - it only provides Qt 5.5.1 in repository.

Jaqueeee:
ok. tnx @mbg033 
I'll find a workaround that doesnt involve > 2.0 then.

Jaqueeee:
@dEBRUYNE-1 qtquick dependency removed.

dternyak:
@Jaqueeee Looks like you have it covered on both counts!

PR #205, opened 2016-11-25
Change OrganizationName to MoneroProject

hyc:
Get rid of spaces, makes for ugly .config path on POSIX systems
Obsoletes #96

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -54,7 +54,7 @@ int main(int argc, char *argv[])
 
     app.setApplicationName("monero-core");
     app.setOrganizationDomain("getmonero.org");
-    app.setOrganizationName("The Monero Project");
+    app.setOrganizationName("MoneroProject");
 
     filter *eventFilter = new filter;
     app.installEventFilter(eventFilter);


dEBRUYNE-1:
Wouldn't "Monero" be a more elegant solution?

maitscha:
+1 for "Monero"

hyc:
I used "MoneroProject" since the github org name is also "monero-project" and this was close enough without using special characters.

taushet:
+1 'Monero'.

xmr-eric:
+1 Monero

moneromooo-monero:
I kinda like Monero better, but since this is a directory, it'll possibly end up having a kovri.conf file in it too at some point, and MoneroProject might make more sense in that light.

hyc:
@moneromooo-monero exactly. This is an organization name and it shouldn't be so specific since the Monero Project has other software, not just Monero itself.

fluffypony:
+1 for @hyc's suggestion. I've given up wanting lowercase folders on everything, and have grown to accept the occasional title-cased folder name:)

peronero:
If the idea is that the .config directory serves as an umbrella "vendor" location for Monero Project projects, and the reasoning behind 'MoneroProject' is to maintain consistency with Github, then isn't 'monero-project' a better solution? Is Qt's Organization Name exposed to end-users anywhere?

moneromooo-monero:
You know what ? We totally should bypass Qt's silly system and write to ~/.monerorc like any sane software.

fluffypony:
Can we get some consensus and figure out how to move forward with this? Thanks:)

hyc:
At this point all I really care is
* no embedded spaces
* the name should reflect the Project, not just "monero"

monero-project is fine with me.

Jaqueeee:
monero-project is ok.

moneromooo-monero:
Seems fine to me.

hyc:
obsoleted by #267

PR #208, opened 2016-11-26
Add fee multiplier info to priority levels

moneromooo-monero:
Users are unlikely to know the link

### DIFF
--- a/components/PrivacyLevelSmall.qml
+++ b/components/PrivacyLevelSmall.qml
@@ -99,7 +99,7 @@ Item {
             font.bold: true
             color: "#000000"
             x: row.x + (row.positions[0] !== undefined ? row.positions[0].currentX - 3 : 0) - width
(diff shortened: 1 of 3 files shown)


taushet:
Looks good and obviously builds fine. 

Is there any way of estimating (however roughly) what this actually means to the user in terms of transaction time? I would suggest that currently the difference is minimal, but should there be a  marked increase in transaction volume, I imagine it would.

moneromooo-monero:
There probably is, though it also depends on the volume of transactions coming *after* that tx, so then it becomes a bit of a moving target, and our initial estimation might go wrong. Especiually since other people may start using larger fees later, if there's a glut, which is exactly when you want to know. So... maybe. Feel free to add a bug about it, but I don't think this will be done soon, but good to remember.

taushet:
I won't clog the issues list with it for now, but as per our discussion on IRC, I'll add it after the dust settles following release.

PR #209, opened 2016-11-26
Print wallet filename in password dialog where approrpiate

moneromooo-monero:


moneromooo-monero:
needs https://github.com/monero-project/monero/pull/1380

PR #210, opened 2016-11-26
Receive: use atomic units for amount in the QR code encoded URI

moneromooo-monero:
This is what the spec mandates


moneromooo-monero:
It was right, after all.

PR #235, opened 2016-12-03
Show destinations in tx details popup

### LINKED ISSUE #229: [Feature] Add recipient's address to transaction details in transaction history
This is consistent with the behavior of `show_transfers` in `monero-wallet-cli`. In addition, this makes it more convenient for users to verify their payment.

Jaqueeee:
Fixes #229

Also added Transfer class for later use.

### DIFF
--- a/src/libwalletqt/TransactionInfo.cpp
+++ b/src/libwalletqt/TransactionInfo.cpp
@@ -1,7 +1,8 @@
 #include "TransactionInfo.h"
 #include "WalletManager.h"
-
+#include "Transfer.h"
 #include <QDateTime>
+#include <QDebug>
 
 TransactionInfo::Direction TransactionInfo::direction() const
 {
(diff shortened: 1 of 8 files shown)


src/libwalletqt/Transfer.h
```diff
@@ -0,0 +1,28 @@
...
+class Transfer : public QObject
+{
+    Q_OBJECT
+    Q_PROPERTY(quint64 amount READ amount)
+    Q_PROPERTY(QString address READ address)
+private:
+    explicit Transfer(uint64_t _amount, const QString &_address,  QObject *parent = 0): m_amount(_amount), m_address(_address), QObject(parent) {};
+private:
+    friend class TransactionInfo;
+    qint64 m_amount;
```
moneromooo-monero: Is there a good reason why this is signed ?

src/libwalletqt/Transfer.h
```diff
@@ -0,0 +1,28 @@
...
+    QString m_address;
+
+public:
+    qint64 amount(){ return m_amount; }
```
moneromooo-monero: Same about signed, also const

src/libwalletqt/TransactionInfo.cpp
```diff
@@ -71,6 +71,27 @@ QString TransactionInfo::paymentId() const
...
 }
 
+QString TransactionInfo::destinations_formatted() const
+{
+    QString destinations;
+    for (auto const& t: transfers()) {
+        if (!destinations.isEmpty())
+          destinations += "<br> ";
+        destinations +=  WalletManager::instance()->displayAmount(t->amount()) + ": " + t->address();
+    }
```
moneromooo-monero: Looks like this leaks the transfers which are newed below

Jaqueeee: Fixed. Gets deleted with parent TransactionInfo

src/libwalletqt/Transfer.h
Jaqueeee: no. fixed.

src/libwalletqt/Transfer.h
Jaqueeee: fixed

PR #236, opened 2016-12-04
added japanese translation

kenshi84:
I tried my best to convey the meanings, but I was confused by this sentence:

    Payment ID&lt;/b&gt;&lt;br/&gt;&lt;br/&gt;A unique user name used in&lt;br/&gt;the address book. It is not a&lt;br/&gt;transfer of information sent&lt;br/&gt;during the transfer

I guess the address book feature is yet to be implemented, but I believe Payment ID does get transferred and stored in the blockchain (either with or without encryption). What's the intent of this text?

Another confusing sentence: `Kickstart the Monero blockchain?` - what does it mean to kickstart a blockchain?

### DIFF
--- a/translations/monero-core_ja.ts
+++ b/translations/monero-core_ja.ts
@@ -1,0 +1,1014 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="ja_JP">
+<context>
+    <name>AddressBook</name>
+    <message>
+        <location filename="../pages/AddressBook.qml" line="47"/>
+        <source>Add new entry</source>
+        <translation>新規項目の追加</translation>
+    </message>
+    <message>
+        <location filename="../pages/AddressBook.qml" line="56"/>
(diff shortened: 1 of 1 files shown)


dternyak:
Thank you for your hard work on this!

Jaqueeee:
@kenshi84 Nice work! "Payment ID" is used on send page, History and in the Address book which i'm currently working on. The description of the payment ID you refer to is supposed to be shown when hovering the payment ID field. But that's not implemented yet. The "kickstart monero blockchain" is currently hidden and will most likely be removed. Unfortunately there are some strings in the codebase that aren't used.

kenshi84:
Still, it sounds strange to me that payment ID is a "unique user name used in the address book". Payment ID is supposed to be unique to a given transaction, not a user. In the yet-to-be-implemented address book, I'd imagine user name displayed with its associated wallet address. 
Am I missing something?

Jaqueeee:
You're not missing anything. That description is wrong and will not be used. Address book consists of three fields. Address, Payment ID and Name

PR #237, opened 2016-12-04
Fix random crashes on wallet close/open

Jaqueeee:
Should fix https://github.com/monero-project/monero-core/issues/230


Jaqueeee:
Need to work more on this one. closing

PR #239, opened 2016-12-04
win32: use static boost

Jaqueeee:
boost on build machine is built with Qt's mingw toolchain

### DIFF
--- a/monero-core.pro
+++ b/monero-core.pro
@@ -76,6 +76,9 @@ win32 {
         MSYS_PATH=c:/msys32/mingw32
     }
 
+    # boost root path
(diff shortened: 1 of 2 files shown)


monero-core.pro
```diff
@@ -87,17 +87,18 @@ win32 {
     }
 
     LIBS+=-L$$MSYS_PATH/lib
+    LIBS+=-L/c/Qt/Qt5.7.0/Tools/mingw530_32/boost/lib
```
danrmiller: Maybe make this path a variable?

Jaqueeee: fixed

PR #249, opened 2016-12-08
Fix random crashes on wallet close/open

### LINKED ISSUE #230: Rare crash when loading password protected wallet
This can happen 1 in 20ish runs on my macbook running osx.

I don't know what is causing this but my best guess is that it has something to do with the async wallet close design. The application will first try to open the wallet with blank password. This fails, async wallet cleanup is initiated (can this be blocked by modal password dialog?), and password dialog gets generated. When correct password is entered the wallet gets loaded, and here the rare crash might occur. If the wallet has no password it will just open it straight away, minimizing risk with far less back and forth logic going on.

Jaqueeee:
Should fix #230

- added mutex lock on open/close wallet
- Make sure only one wallet can be open at same time
- make sure wallet doesn't get closed multiple times-> crash
- removed commented out code from main.qml

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -109,30 +109,8 @@ ApplicationWindow {
             ctrlPressed = false
     }
 
-    function mousePressed(obj, mouseX, mouseY) {
-//        if(obj.objectName === "appWindow")
-//            obj = rootItem
-
-//        var tmp = rootItem.mapFromItem(obj, mouseX, mouseY)
-//        if(tmp !== undefined) {
-//            mouseX = tmp.x
-//            mouseY = tmp.y
-//        }
-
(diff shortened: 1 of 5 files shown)


fluffypony:
Please rebase

Jaqueeee:
Rebased. More testing + review needed before merge. ping @medusadigital @dEBRUYNE-1 
https://build.getmonero.org/downloads/monero-core-d568343-win32.zip

mochaccinuh:
Works good, no crash with my testing.

Test steps: open application, load password protected wallet, close application.
Ran this 200 times by use of automatic keyboard/mouse record playback tool.

Jaqueeee:
rebased again

Issue #242, opened 2016-12-04
Remove broken address book link for GUI beta release


xmr-eric:
Speaking particularly to the link on the main "Send" page. It's in the way, and probably not necessary for GUI beta unless some dev is able to easily implement it.

PR #253, opened 2016-12-09
Remove link to Address Book from Send page

xmr-eric:
GUI seems releasable without Address Book functionality. 

This PR proposes removing the Address Book link for the initial GUI release, with a goal to reimplement it later when functionality is complete.

See #242


xmr-eric:
Closed because of #268

PR #260, opened 2016-12-09
Optional custom bar

moneromooo-monero:


xmr-eric:
On OSX 10.12.1

Wallet installs with custom title bar, which looks great. The question mark symbol doesn't seem to do anything (can we remove it?), but everything else works really well.

Unfortunately if you go into Settings and deselect "Custom decorations" the wallet should restore whatever title bar is default for the OS. But on OSX the title bar simply disappears completely, and theres no way to minimize the window or exit out of the program.

xmr-eric:
@moneromooo-monero https://youtu.be/hoKH2ViAgdY

moneromooo-monero:
I am unable to test this, as the WM I use is not running on the same VM as the GUI. I will try to fix this blind, but I will need someone to test for me.

xmr-eric:
I can test

moneromooo-monero:
Does https://github.com/moneromooo-monero/monero-core/tree/optional-custom-bar2 help ?

moneromooo-monero:
If not, does https://github.com/moneromooo-monero/monero-core/tree/optional-custom-bar3 help ?

fluffypony:
Plz rebase

xmr-eric:
Still not working on my OSX 10.12.2 :\

moneromooo-monero:
Are you running fullscreen ? medusa reports that it breaks when fullscreen on Mac.
If you can compile Qt, there's a sample called "windowflags" which you could try in the same settings to see if it also happens with it, or not.

PR #263, opened 2016-12-10
Remove fee multiplier info from low/high slider

moneromooo-monero:
This was a braino on my part, this is only applicable for the
list based low/high selector.


taushet:
Works here. No longer looks like you pay extra for strong passwords :)

antanst:
Works here as well.

PR #264, opened 2016-12-10
Reduce balance font size

sammy007:
Fixes https://github.com/monero-project/monero-core/issues/215#issue-191814693

Was

Now


sammy007:
Sorry, I checked how it looks like with 4 figures and it sucks.

PR #285, opened 2016-12-15
Clarified and broadened Mnemonic Seed description

xmr-eric:
Simplified (shortened) original seed description, then added emphasis (exclamation point) that users should keep their seed secret. Original seed description didn't communicate this need for privacy.

Also added that the seed has "restore" functionality.

### DIFF
--- a/pages/Settings.qml
+++ b/pages/Settings.qml
@@ -143,7 +143,7 @@ Rectangle {
                 color: "#4A4646"
                 Layout.fillWidth: true
                 wrapMode: Text.WordWrap
-                text: qsTr("It is very important to write it down as this is the only backup you will need for your wallet.")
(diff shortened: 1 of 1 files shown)


dEBRUYNE-1:
Not sure if we should remove "very important" out of it. I'd like to hear other opinions on this PR.

expez:
Good changes, but I agree with @dEBRUYNE-1 about keeping "very important".

xmr-eric:
@dEBRUYNE-1 I removed that phrase but added the exclamation point - which still emphasizes how important it is to write the seed down while using less characters overall.

If the phrase 'very important' is added then the text becomes two lines, and the goal is to keep the description short and to the point. The goal is to keep it on one line. By adding the exclamation point, the extra room allows for a whole bunch of extra terms (that the seed is 'secret' and that it has 'restore' functionality).

PR #291, opened 2016-12-15
fix borked wallet

Jaqueeee:


medusadigital:
works, fixes lots of things, inluding the crash when swithing wallets

PR #297, opened 2016-12-16
Don't waste too much vertical space on the left menu

moneromooo-monero:
Gives space for one more item


xmr-eric:
Lots of new space, but the overall window is still sized the same. Can we place the GUI in a smaller window now that there's more space?

PR #299, opened 2016-12-16
Rename the binary to monero-wallet-gui

moneromooo-monero:
Because I like painting my bike shed every morning


Gingeropolous:
i don't know how trivial this is though.. i mean, to [follow the allegory](https://en.wikipedia.org/wiki/Law_of_triviality), this particular "triviality" has us confusing the nuclear power plant with the bike shed.

fluffypony:
Please rebase

PR #303, opened 2016-12-17
Enable translations

Jaqueeee:
Just for testing. Do not merge until translations are approved


NanoAkron:
Would you add an Indonesian option ;)

PR #306, opened 2016-12-17
Simplify text on wizard Option page

xmr-eric:
Two fixes: 

(1) Removed first-person text. "I want to create a wallet" becomes "Create a wallet". The first person perspective wasn't helping as much as it was getting in the way of the user. It takes a long time to read the phrase "This is my first time, I want to create a new wallet" vs. the phrase "Create a new wallet". I think the original intent was to make the wizard more friendly for newbies but it really wasn't achieving that effect. New people are just going to click on "Create a new wallet" anyways, and there are plenty of people who will want to create a new wallet, but for whom it won't be their "first time".

(2) Normalized the terminology. The first two options (create / restore from seed) were calling wallets "accounts" while the third option (open file) was calling them wallets. So now they all say "wallets".

Here's the original page

screen shot 2016-12-17 at 5 33 32 pm

Here's the new page
option-wizard

### DIFF
--- a/wizard/WizardOptions.qml
+++ b/wizard/WizardOptions.qml
@@ -137,7 +137,7 @@ Item {
                 horizontalAlignment: Text.AlignHCenter
                 width:page.buttonSize
                 wrapMode: Text.WordWrap
-                text: qsTr("This is my first time, I want to create a new account") + translationManager.emptyString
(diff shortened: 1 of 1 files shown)


dternyak:
I like and agree with these changes :)

PR #311, opened 2016-12-18
Translation of Monero GUI into Bahasa Indonesia

NanoAkron:
This will no doubt be pulled apart by a native speaker, but it's at least a first version which allows us to try to connect to some 300 million potential new users (Bahasa Indonesia + Bahasa Melayu + Bahasa Brunei are almost mutually intelligible)

### DIFF
--- a/translations/monero-core_id.ts
+++ b/translations/monero-core_id.ts
@@ -1,0 +1,1285 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="id_ID">
+<context>
+    <name>AddressBook</name>
+    <message>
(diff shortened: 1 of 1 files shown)


translations/monero-core_id.ts
```diff
@@ -0,0 +1,1285 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="fr_FR">
```
Jaqueeee: This is probably not the correct language code for Bahasa. id_ID maybe?

NanoAkron: oops. Will change

Jaqueeee:
@NanoAkron More strings have been added to GUI since you added this. Please update translation when you have time.

PR #316, opened 2016-12-21
Tab text: change Verify Payment to Check Payment

xmr-eric:
Now that we have an Advanced section of the left menu, Verify Payment and Sign/Verify are now next to each other. They sound the same. This PR changes the Verify Payment tab to Check Payment to better differentiate the two, which works well since the button inside Verify Payment is labeled "Check" anyways.


dternyak:
Nice - these simple but well thought out changes definitely provide for a better UX.

PR #329, opened 2016-12-21
Address book: fix tooltip label and visibility

Jaqueeee:
Fixes https://github.com/monero-project/monero-core/issues/327


taushet:
Works.

PR #351, opened 2016-12-23
Reduce the number of "balls" in the wizards

ClementJnc:
Reduce the number of "balls" in the wizards (creation of wallet and recovery) to be coherent with the hidden steps (3 steps instead of 4). Color green for all.


xmr-eric:
Works

PR #357, opened 2016-12-24
History: print address book description if available

moneromooo-monero:


components/HistoryTable.qml
```diff
@@ -207,6 +225,21 @@ ListView {
...
+                selectByMouse: true
+                id: addressBookLookupValue
+                width: 136
+                anchors.bottom: parent.bottom
+                //elide: Text.ElideRight
+                font.family: "Arial"
+                font.pixelSize:13
+                font.letterSpacing: -1
+                color: "#545454"
+                text: "(" + lookupPaymentID(paymentId) + ")"
```
Jaqueeee: Is this for incoming only? For outgoing i think we should match against both adress and payment ID since they arent uniqe. Or address only when paymentId isn't set.

components/HistoryTable.qml
```diff
@@ -52,6 +54,22 @@ ListView {
...
+        if (paymentId.length == 16)
+            paymentId += "000000000000000000000000000000000000000000000000"
+        var count = addressBookModel.rowCount()
+        for (var i = 0; i < count; i++) {
+            var idx = addressBookModel.index(i, 0)
+            var ati = addressBookModel.data(idx, AddressBookModel.AddressBookPaymentIdRole)
+            if (ati.length == 16)
+                ati += "000000000000000000000000000000000000000000000000"
+            if (ati == paymentId)
+                return addressBookModel.data(idx, AddressBookModel.AddressBookDescriptionRole)
```
Jaqueeee: Would prefer having the lookup functions in AddressBook instead. For later reuse.

components/HistoryTable.qml
```diff
@@ -52,6 +54,22 @@ ListView {
             + translationManager.emptyString;
     }
 
+    function lookupPaymentID(paymentId) {
```
Jaqueeee: Would prefer having the lookup functions in AddressBook instead. For easier reuse later. What do you think?

moneromooo-monero: Looking at the AddressBook code, it's indexed by... index, but there's a getRowID. What is that ?

components/HistoryTable.qml
moneromooo-monero: It's incoming only. It might trigger on outgoing if there's a payment id collision, though.

components/HistoryTable.qml
Jaqueeee: index and rowID are the same. First implementation used std::map and "rowId" was stored in object. After switching to std::vector storing of rowId isn't necessary and ought to be removed.

moneromooo-monero:
The lookup is now in the address book code.
Needs https://github.com/monero-project/monero/pull/1521

PR #362, opened 2016-12-27
Create monero-core_br.ts

moneroj:
add pt_BR translation


Jaqueeee:
@moneroj  More strings have been added to GUI since you added this. Please update translation when you have time.

PR #365, opened 2016-12-27
Transfer: print mixin when moving privacy level widget around

moneromooo-monero:


xmr-eric:
Looks great. Works great.

PR #368, opened 2016-12-28
Add pool mining screen

moneromooo-monero:
Wa want to make it easy for people to mine

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -316,29 +316,6 @@ Rectangle {
                 height: 1
             }
 
-          /*  // ------------- Mining tab ---------------
-            MenuButton {
-                id: miningButton
-                anchors.left: parent.left
-                anchors.right: parent.right
-                text: qsTr("Mining") + translationManager.emptyString
-                symbol: qsTr("M") + translationManager.emptyString
-                dotColor: "#FFD781"
-                onClicked: {
-                    parent.previousButton.checked = false
-                    parent.previousButton = miningButton
-                    panel.miningClicked()
-                }
-            }
-
-            Rectangle {
-                anchors.left: parent.left
-                anchors.right: parent.right
-                anchors.leftMargin: 16
-                color: miningButton.checked || settingsButton.checked ? "#1C1C1C" : "#505050"
-                height: 1
-            }
-            */
             // ------------- Advanced tab ---------------
             MenuButton {
                 id: advancedButton
@@ -360,6 +337,29 @@ Rectangle {
                 height: 1
             }
 
+            // ------------- Mining tab ---------------
+            MenuButton {
+                id: miningButton
+                anchors.left: parent.left
+                anchors.right: parent.right
+                text: qsTr("Mining") + translationManager.emptyString
+                symbol: qsTr("M") + translationManager.emptyString
+                dotColor: "#FFD781"
+                under: advancedButton
+                onClicked: {
+                    parent.previousButton.checked = false
+                    parent.previousButton = miningButton
+                    panel.miningClicked()
+                }
+            }
+
+            Rectangle {
+                anchors.left: parent.left
+                anchors.right: parent.right
+                anchors.leftMargin: 16
+                color: miningButton.checked || settingsButton.checked ? "#1C1C1C" : "#505050"
+                height: 1
+            }
             // ------------- TxKey tab ---------------
             MenuButton {
                 id: txkeyButton
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -51,6 +51,7 @@ Rectangle {
     property History historyView: History { }
     property Sign signView: Sign { }
     property Settings settingsView: Settings { }
+    property Mining miningView: Mining { }
     property AddressBook addressBookView: AddressBook { }
 
 
@@ -143,7 +144,7 @@ Rectangle {
                PropertyChanges { target: root; currentView: settingsView }
             }, State {
                 name: "Mining"
-                PropertyChanges { /*TODO*/ }
+                PropertyChanges { target: root; currentView: miningView }
             }
         ]
 
--- a/components/StandardDropdown.qml
+++ b/components/StandardDropdown.qml
@@ -107,7 +107,7 @@ Item {
             font.bold: true
             font.pixelSize: 12
             color: "#FFFFFF"
-            text: repeater.model.get(column.currentIndex).column1
+            text: column.currentIndex < repeater.model.rowCount() ? repeater.model.get(column.currentIndex).column1 : ""
         }
 
         Text {
@@ -119,7 +119,7 @@ Item {
             font.family: "Arial"
             font.pixelSize: 12
             color: "#FFFFFF"
-            text: repeater.model.get(column.currentIndex).column2
+            text: column.currentIndex < repeater.model.rowCount() ? repeater.model.get(column.currentIndex).column2 : ""
 
             property int w: 0
             Component.onCompleted: w = implicitWidth
--- a/main.qml
+++ b/main.qml
@@ -362,7 +362,6 @@ ApplicationWindow {
         daemonRunning = false;
     }
 
-
     function onWalletNewBlock(blockHeight) {
 
             // Update progress bar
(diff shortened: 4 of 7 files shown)


xmr-eric:
Tried on OSX 10.12.2. If I click "Start poolMiner" it tries to start but fails with no log output. Also it could use a right side margin.

screen shot 2016-12-28 at 5 12 48 pm

xmr-eric:
Personally, I don't like this feature, and feel we shouldn't include it. Pool mining is the one page in the wallet where we're suddenly relying on a third party. What qualifies a pool to be added to this list? Certainly cryptmonero.com isn't included since this pool is tied to Reddit-reported malware mining. So there is already a distinction between good and bad pools, and suddenly we're tasked with making that distinction. Suddenly our work begins to grow exponentially, especially if XMR really takes off.

Suddenly pool operators have an incentive to get in good graces with XMR devs? Already we can see MoneroWorld (where I personally mine, so I love it there) is at the top of the list?

What if one of the listed pools gets hacked, malware, etc. We have to wait for a whole release before we can change the code away from an adversarial pool? 

This is not decentralized, and thus sort of goes against the spirit of The Monero Project.

I'm a big fan of adding a solo smart miner to the wallet. Not a fan of this.

moneromooo-monero:
Yes, I don't really like it either, unless it's an optional plugin. It just seems to be a lesser evil than dwarfpool at > 50% hash rate. Other opinions welcome.

moneromooo-monero:
In fact, I'll add solo mining too. Tomorrow :)

moneromooo-monero:
The latest patch removes the default pool list, and adds a "load from file" button.
We may still have to provide a default though, for ease of use. We'll see.

expez:
'load form file' seems very cumbersome.  Why not just let people copy/paste the pool address into the GUI directly, instead of editing a file (and then re/loading the file to pick up changes?)

moneromooo-monero:
Because you're not really supposed to enter URIs directly. There will *probably* be a default pool file, loaded at startup. Most people will likely then just select a pool (or keep on automatic). Third parties can make/maintain their own pool list that people can download/import. The intent is that the control of the pool list is not solely on the devs', er, control.

NanoAkron:
Will this be amended if/when a future universal p2pool miner is implemented?

moneromooo-monero:
Solo mining is now in (not smart mining yet obviously)

moneromooo-monero:
The last commit needs https://github.com/monero-project/monero/pull/1513

moneromooo-monero:
Also, for the time being, the pools list is empty.
To test or use, place the following in a file, and load that with the load button:

moneroworld: stratum+tcp://drill.moneroworld.com:7777
monerohash.com: stratum+tcp://monerohash.com:7777
poolto.be: stratum+tcp://xmr.poolto.be:3000
supportxmr.com: stratum+tcp://pool.supportxmr.com:5555
xmrpool.eu: stratum+tcp://xmrpool.eu:5555
prohash.net: stratum+tcp://xmr.prohash.net:2222

xmr-eric:
OSX 10.12.2

I built it (included #1513), built MinerD from [**here**](https://github.com/correcthorse/cpuminer-multi) and added it to the directory, and also added pools.txt containing the list above. Loaded the wallet, then the daemon, and navigated to Advanced > Mining. 

Noticed the Monero Solo Mining tool had already started? Watching it, I noticed the solo miner was starting and stopping. The "Start Mining" button would depress and it would say "Status: mining at 0 H/s" at the bottom. Then the "Stop Mining" button would depress and it would say "Status: not mining."

I also loaded the pools.txt file and clicked "Start poolMiner" but it again didn't do anything and there was no output in the log.

Here's the daemon log-level 1: https://paste.fedoraproject.org/514809/48303598/

screen shot 2016-12-29 at 1 24 32 pm

xmr-eric:
Log level 2: https://paste.fedoraproject.org/514818/14830378/

moneromooo-monero:
Have you built the node with monero-project/monero#1513 ? This is needed for solo mining,
As for pool mining, any error in the GUI output ?

moneromooo-monero:
And did the pool list fill up after you loaded it ?

Hueristic:
What about separate releases? The one with the integrated pool release add the stratum and have the default local, therefore to change from mining on a local pool the user would have to enter one them self or just as easily add their pool to the list (maybe setup and auto msg to irc for the new pool to announce and scrape it?).

antanst:
Imho the pool list should not be populated by default. Doing so may be considered as an official endorsement; it also encourages centralization to specific pools.

iamsmooth:
> Yes, I don't really like it either, unless it's an optional plugin. It just seems to be a lesser evil than dwarfpool at > 50% hash rate. Other opinions welcome.

I agree with the comment not liking providing a specific pool list in the standard software as it puts the project in the bad position of either acting as gatekeeper on the list or adding bad/scam pools to the list (knowingly or otherwise).

I also don't think this is a useful way to address a particularly pool (dwarf currently, but that's not really the point) having a high hash rate. For one thing, the number of GUI wallet users will likely be much too low to make any real difference in the near term and for another it could easily lead to even more pool concentration in the future as wallet users (when there are a lot) may still choose the largest and best-known pools.

If the project-supported software is going to support mining it should be focused solo mining, which robustly helps the problem of pool concentration and avoids potentially making the problem worse. Serious miners with high hash rate who want to pool mine will find their way to the appropriate software one way or another. I would revise the text to present it as primarily helping the network, and any rewards being secondary. But again given the likely near-term usage of the GUI wallet this likely isn't signifiant either.

Also, as I understand it, there are a few feature-related reasons that dwarfpool is popular. I've seen more than one person mention rig-specific stats and miner downtime notification. It would probably be more useful to pool concentration to add those features to the open source mining software.

moneromooo-monero:
This latest update has only solo mining.

moneromooo-monero:
Now with wording/layout/error changes when the daemon isn't local, as suggested by smooth

xmr-eric:
Similar problem to last time.

OSX 10.12.2

Built with monero-project/monero#1513. Added MinerD from **[here](https://github.com/correcthorse/cpuminer-multi)**. Upon loading the wallet, starting the daemon, and letting the sync complete, I navigated to Advanced > Mining. Here the "Start mining" button was already depressed, but "Status: mining at 0 H/s". Then the button would automatically click to "Stop mining" and the new readout would be "Status: not mining". And it would click between the two, never adding additional load to my Mac's resource monitor. If I tried to click "Stop mining" myself, the program wouldn't respond.

Here's the log-level2 output: https://paste.fedoraproject.org/515725/83141887/

See line 2703:

2016-Dec-30 18:46:29.543597 [RPC0]HTTP [127.0.0.1] GET /stop_mining
2016-Dec-30 18:46:29.543722 [RPC0]Miner has received stop signal
2016-Dec-30 18:46:29.543752 [RPC0]Not mining - nothing to stop
2016-Dec-30 18:46:29.543800 [RPC0]/stop_mining processed with 0/0/0m

moneromooo-monero:
Oops, I'd been using fixes in the GUI VM which I hadn't merged, and forgot. They're now merged in 1513.

xmr-eric:
Works great!

xmr-eric:
Perhaps the gray outline box around the miner can be removed since pool mining was taken away.

xmr-eric:
I attempted to start a miner using 521,230,093 threads and the GUI locked up.

screen shot 2016-12-31 at 4 02 33 pm

iamsmooth:
I think its okay to leave this under Advanced for now as it is new and not necessarily user friendly (doesn't do any smart mining type thinks to avoid excessive system load, etc.) but the plan should be to move this out of advanced and encourage, or a least make easily available, some "help the network" mining for everyone.

fluffypony:
I don't know whether or not I should merge this or there's more stuff to be done...can I get some input, plz?

moneromooo-monero:
(though the change for this is not in this patch, but in the daemon)

Jaqueeee:
just make sure https://github.com/monero-project/monero/pull/1513 is merged also.

PR #375, opened 2016-12-31
Add "(syncing)" after balance when syncing

moneromooo-monero:
Should cut down on people wondering why their balance is wrong


NanoAkron:
I wonder what is it syncing about...

moneromooo-monero:
The blockchain. From the node.

fluffypony:
@moneromooo-monero https://www.youtube.com/watch?v=gmOTpIVxji8

PR #383, opened 2017-01-03
wizard: Ensure temporary wallet files are deleted

Jaqueeee:


oshelper.cpp
```diff
@@ -18,6 +18,14 @@ QString OSHelper::temporaryFilename() const
     return tempFileName;
 }
 
+bool OSHelper::removeTemporaryWallet(const QString &fileName) const
+{
+    // Temporary files should be deleted automatically by default, in case they should'nt we delete them manually as well
+    if (!QFile::remove(fileName) || !QFile::remove(fileName + ".address.txt") || !QFile::remove(fileName + ".keys"))
```
moneromooo-monero: This short-circuits, is that intended ?

Jaqueeee: no. fixed. thanks

PR #384, opened 2017-01-04
Finnish translation for GUI

moimoij:
Should be at least somewhat accurate.


Jaqueeee:
@vincenj  More strings have been added to GUI since you added this. Please update translation when you have time.

PR #403, opened 2017-01-14
Change slider mixin=4 term from "low" to "normal"

xmr-eric:
A mixin of 4 isn't really "low" privacy. If people want a higher mixin, that's great, and the HIGH term does convey that a mixin of 25 is higher privacy than the default setting. But the default setting should not convey that privacy is LOW. In fact, the default mixin is among the most private tx a person can send in all of cryptoland. A better term for default XMR privacy (mixin=4) is something like "normal" / "standard" / "typical" / "average" / "default". Of these "normal" seems the best for the newbie Monero user.

### DIFF
--- a/components/TickDelegate.qml
+++ b/components/TickDelegate.qml
@@ -52,7 +52,7 @@ Item {
             font.pixelSize: 12
             color: "#4A4949"
             text: {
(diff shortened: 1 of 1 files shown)


xmr-eric:
See this discussion. I lean towards keeping it "normal", but others have suggested "default": https://paste.fedoraproject.org/527266/14844271/

PR #411, opened 2017-01-16
exit script with appropriate status if these commands fail

danrmiller:
Right now if these commands fail, the script still exits with status 0, so automation such as buildbot won't know there is a problem. For example:

https://build.getmonero.org/builders/monero-core-ubuntu-amd64-libc2.19/builds/28/steps/deploy/logs/stdio


danrmiller:
This PR caused the desired result:

https://build.getmonero.org/builders/monero-core-ubuntu-amd64-libc2.19/builds/29/steps/deploy/logs/stdio

danrmiller:
I've come across a few more changes to make, and then I'll squash.

PR #420, opened 2017-01-18
Allow some QT lib paths to be set in environment variables

danrmiller:
Depending on the architecture, distribution, and where I installed QT from, these paths are different. Unless someone has a better idea, if we just allow them to be set like this I can make adjustments locally.

### DIFF
--- a/linuxdeploy_helper.sh
+++ b/linuxdeploy_helper.sh
@@ -11,12 +11,12 @@ platform=$(get_platform)
 
 # Copy dependencies
 EXCLUDE='libstdc++|libgcc_s.so|libc.so|libpthread'
-cp -rv /usr/lib/x86_64-linux-gnu/qt5/qml $TARGET
-cp -rv /usr/lib/x86_64-linux-gnu/qt5/plugins $TARGET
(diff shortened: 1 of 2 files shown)


linuxdeploy_helper.sh
```diff
@@ -1,5 +1,13 @@
...
+if [ -z "$QT_DIR" ]; then
+    QT_DIR="/usr/lib/x86_64-linux-gnu/qt5"
+fi
+
+if [ -z "$QTXML_DIR" ]; then
+    QTXML_DIR=$QT_DIR/../
+fi
+
 ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source $ROOT_DIR/utils.sh
```
Jaqueeee: env. variables sounds like the best solution. You could use `platform=$(get_platform)` from utils.sh if you want to specify default paths for other architectures (linux|linux32|linux64).

danrmiller: Thanks @Jaqueeee I tried that now.

PR #429, opened 2017-01-23
Minor spelling corrections

davidlatapie:
The biggest one was a missing word.

### DIFF
--- a/translations/monero-core_fr.ts
+++ b/translations/monero-core_fr.ts
@@ -22,17 +22,17 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="77"/>
         <source>Payment ID &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</source>
-        <translation>Identifiant de paiement &lt;font size=&apos;2&apos;&gt;(Optionnel)&lt;/font&gt;</translation>
+        <translation>Identifiant de paiement &lt;font size=&apos;2&apos;&gt;(optionnel)&lt;/font&gt;</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="79"/>
         <source>&lt;b&gt;Payment ID&lt;/b&gt;&lt;br/&gt;&lt;br/&gt;A unique user name used in&lt;br/&gt;the address book. It is not a&lt;br/&gt;transfer of information sent&lt;br/&gt;during the transfer</source>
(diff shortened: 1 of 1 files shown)


translations/monero-core_fr.ts
```diff
@@ -150,7 +150,7 @@
     <message>
         <location filename="../pages/History.qml" line="70"/>
         <source> selected: </source>
-        <translation> séléctionné : </translation>
+        <translation> sélectionné&nbsp;: </translation>
```
Jaqueeee: Any special reason for changing all spaces to &nbsp;? Also wonder if the space between the word and the : needed anyway? Look at the source string.

translations/monero-core_fr.ts
```diff
@@ -505,7 +505,7 @@
     <message>
         <location filename="../components/SearchInput.qml" line="69"/>
         <source>Search by...</source>
-        <translation>Chercher par...</translation>
+        <translation>Chercher par…</translation>
```
Jaqueeee: Was this change intentional?

translations/monero-core_fr.ts
```diff
@@ -1286,7 +1286,7 @@ Description : </translation>
     <message>
         <location filename="../main.qml" line="731"/>
         <source>Please wait...</source>
-        <translation>Veuillez patienter...</translation>
+        <translation>Veuillez patienter…</translation>
```
Jaqueeee: and this

translations/monero-core_fr.ts
davidlatapie: French typography requires non-breaking spaces (and in some instance, thin non-break spaces)

translations/monero-core_fr.ts
davidlatapie: Yes

translations/monero-core_fr.ts
davidlatapie: Ditto. According to Unicode, ... and … convey the same exact meaning, so you can change it back if you're not comfortable with this, though. Still, I consider ellipsis to carry a semantic value dot dot dot doesn't have.

Jaqueeee: OK. Was just curious if it was changed by you or automatically by your text editor. :)

translations/monero-core_fr.ts
Jaqueeee: OK. Cool. We haven't really tested the translations yet, so i'm not sure how the urlencoded strings will be treated.

PR #430, opened 2017-01-23
Terminology change: switch "mixin" with "ring size"

xmr-eric:
As per the discussion here:

https://www.reddit.com/r/Monero/comments/5pmz3f/maybe_a_good_idea_to_rename_mixin_in_the_code_to/

Good point by u/knaccc: "For mass adoption, people should not need to understand what a ring signature is, and then understand the mathematics of using different mixin levels, in order to feel comfortable sending a payment."

This change fits this desire well. The privacy slider is still titled "Privacy level", which even newbies can understand has something to do with increasing the privacy of the tx. But in parentheses the term will say "ring size", which is a more accurate term than the current term "mixin". Best of both worlds.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -464,7 +464,7 @@ ApplicationWindow {
                         + (paymentId === "" ? "" : (qsTr("\nPayment ID: ") + paymentId))
                         + qsTr("\n\nAmount: ") + walletManager.displayAmount(transaction.amount)
                         + qsTr("\nFee: ") + walletManager.displayAmount(transaction.fee)
-                        + qsTr("\n\nMixin: ") + mixinCount
+                        + qsTr("\n\nRing size: ") + mixinCount
                         + qsTr("\n\Number of transactions: ") + transaction.txCount
                         + (transactionDescription === "" ? "" : (qsTr("\n\nDescription: ") + transactionDescription))
                         + translationManager.emptyString
--- a/main.qml
+++ b/main.qml
@@ -464,7 +464,7 @@ ApplicationWindow {
                         + (paymentId === "" ? "" : (qsTr("\nPayment ID: ") + paymentId))
                         + qsTr("\n\nAmount: ") + walletManager.displayAmount(transaction.amount)
                         + qsTr("\nFee: ") + walletManager.displayAmount(transaction.fee)
-                        + qsTr("\n\nRing size: ") + mixinCount
+                        + qsTr("\n\nRing size: ") + mixinCount + 1
                         + qsTr("\n\Number of transactions: ") + transaction.txCount
                         + (transactionDescription === "" ? "" : (qsTr("\n\nDescription: ") + transactionDescription))
                         + translationManager.emptyString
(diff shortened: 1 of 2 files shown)


expez:
How is 'ring size' any better than 'mixin' for the uninitiated?  Wouldn't it be better to talk about 'decoys' or something to that effect?  To a layman that would give some indication that the transaction is hiding among multiple others.

xmr-eric:
@expez "Decoys" is a decent option, but it also isn't really going to help newbies, who won't ever need to understand the privacy mechanism at all. The goal is to both (1) have a generalized term while also (2) have a specific term. Our generalized term is "Privacy level". Almost everybody can grasp something about what this means. For a specific term, previously giving a generic term (Privacy level) now affords us more leeway to add terminology specific to our actual brand. Ring size is uniquely Cryptonote/Monero.

xmr-eric:
ringsize

pages/Transfer.qml
```diff
@@ -74,7 +74,7 @@ Rectangle {
         var mixin = scaleValueToMixinCount(fillLevel)
         print ("PrivacyLevel changed:"  + fillLevel)
         print ("mixin count: "  + mixin)
-        privacyLabel.text = qsTr("Privacy level (mixin %1)").arg(mixin) + translationManager.emptyString
+        privacyLabel.text = qsTr("Privacy level (ring size %1)").arg(mixin) + translationManager.emptyString
```
Jaqueeee: Same here. mixin +1

pages/Transfer.qml
```diff
@@ -560,7 +560,7 @@ Rectangle {
                     + (transaction.paymentId[i] == "" ? "" : qsTr("\n\payment ID: ") + transaction.paymentId[i])
                     + qsTr("\nAmount: ") + walletManager.displayAmount(transaction.amount(i))
                     + qsTr("\nFee: ") + walletManager.displayAmount(transaction.fee(i))
-                    + qsTr("\nMixin: ") + transaction.mixin(i)
+                    + qsTr("\nRing size: ") + transaction.mixin(i)
```
Jaqueeee: and here

main.qml
```diff
@@ -464,7 +464,7 @@ ApplicationWindow {
                         + (paymentId === "" ? "" : (qsTr("\nPayment ID: ") + paymentId))
                         + qsTr("\n\nAmount: ") + walletManager.displayAmount(transaction.amount)
                         + qsTr("\nFee: ") + walletManager.displayAmount(transaction.fee)
-                        + qsTr("\n\nMixin: ") + mixinCount
+                        + qsTr("\n\nRing size: ") + mixinCount
```
Jaqueeee: If I'm not mistaken @dEBRUYNE-1 told me that ring size = mixin +1.

xmr-eric: @Jaqueeee I only know how to change words. If we need to change the actual calculation to ringsize=mixin+1 then I'll close this PR and you can do a better one?

pages/Transfer.qml
Jaqueeee: `arg(mixin+1)`

pages/Transfer.qml
Jaqueeee: `transaction.mixin(i+1)`

main.qml
Jaqueeee: `qsTr("\n\nRing size: ") + mixinCount + 1`

Jaqueeee: ^^ should work. If you dare to try it ;)

PR #431, opened 2017-01-23
Simplified "OR ALL" button to "ALL"

xmr-eric:
The purpose of keeping "OR" in the button was to avoid people accidentally sending their entire account balance, ie. keeping "OR" helps the user realize the button is completely optional and not required for an XMR transaction.

However, "OR ALL" looks awkward as is, and the functionality is still intuitive if simplified to "ALL", so that is the purpose of this PR.


xmr-eric:
all-button

PR #432, opened 2017-01-23
Payment Description into placeholder

xmr-eric:
Mimics the same design as Payment ID. Keep description of field in placeholder rather than above the field as how Description originally was.


xmr-eric:
placeholder-change

PR #433, opened 2017-01-23
AddressBook.qml placeholders

xmr-eric:


pages/AddressBook.qml
```diff
@@ -114,6 +116,7 @@ Rectangle {
         anchors.leftMargin: 17
         anchors.rightMargin: 17
         anchors.topMargin: 5
+        placeholderText: qsTr("Saved to local wallet") + translationManager.emptyString
```
Jaqueeee: All fields in address book are saved to local wallet. So maybe this placeholder isn't needed?

PR #437, opened 2017-01-26
Restore wallet from keys

Jaqueeee:
requires #1618 and #1635

Also includes a small redesign of wizard.

### DIFF
--- a/src/libwalletqt/WalletManager.cpp
+++ b/src/libwalletqt/WalletManager.cpp
@@ -86,6 +86,22 @@ Wallet *WalletManager::recoveryWallet(const QString &path, const QString &memo, 
     return m_currentWallet;
 }
 
+Wallet *WalletManager::createWalletFromKeys(const QString &path, const QString &language, bool testnet,
+                                            const QString &address, const QString &viewkey, const QString &spendkey,
+                                            quint64 restoreHeight)
+{
+    QMutexLocker locker(&m_mutex);
+    if (m_currentWallet) {
+        qDebug() << "Closing open m_currentWallet" << m_currentWallet;
+        delete m_currentWallet;
+        m_currentWallet = NULL;
+    }
+    Monero::Wallet * w = m_pimpl->createWalletFromKeys(path.toStdString(), language.toStdString(), testnet, restoreHeight,
+                                                       address.toStdString(), viewkey.toStdString(), spendkey.toStdString());
+    m_currentWallet = new Wallet(w);
+    return m_currentWallet;
+}
+
 
 QString WalletManager::closeWallet()
 {
@@ -181,6 +197,16 @@ bool WalletManager::paymentIdValid(const QString &payment_id) const
 bool WalletManager::addressValid(const QString &address, bool testnet) const
 {
     return Monero::Wallet::addressValid(address.toStdString(), testnet);
+}
+
+bool WalletManager::keyValid(const QString &key, const QString &address, bool isViewKey,  bool testnet) const
+{
+    std::string error;
+    if(!Monero::Wallet::keyValid(key.toStdString(), address.toStdString(), isViewKey, testnet, error)){
+        qDebug() << QString::fromStdString(error);
+        return false;
+    }
+    return true;
 }
 
 QString WalletManager::paymentIdFromAddress(const QString &address, bool testnet) const
--- a/wizard/WizardManageWalletUI.qml
+++ b/wizard/WizardManageWalletUI.qml
@@ -29,12 +29,16 @@
 import QtQuick 2.2
 import moneroComponents.TranslationManager 1.0
 import QtQuick.Dialogs 1.2
-import QtQuick.Layouts 1.1
+import QtQuick.Layouts 1.2
 import "../components"
+import 'utils.js' as Utils
 
 // Reusable component for managing wallet (account name, path, private key)
 
-Item {
+ColumnLayout {
+    anchors.left: parent.left
+    anchors.right: parent.right
+    spacing: 5
 
     property alias titleText: titleText.text
     property alias accountNameText: accountName.text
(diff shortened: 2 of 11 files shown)


src/libwalletqt/WalletManager.cpp
```diff
@@ -86,6 +86,21 @@ Wallet *WalletManager::recoveryWallet(const QString &path, const QString &memo,
...
 }
 
+Wallet *WalletManager::createWalletFromKeys(const QString &path, const QString &language, bool testnet,
+                                            const QString &address, const QString &viewkey, const QString &spendkey,
+                                            quint64 restoreHeight)
+{
+    QMutexLocker locker(&m_mutex);
+    if (m_currentWallet) {
+        qDebug() << "Closing open m_currentWallet" << m_currentWallet;
+        delete m_currentWallet;
```
moneromooo-monero: This can leave m_currentWallet dangling if the call below throws. Can it ? If not, all is fine.

src/libwalletqt/WalletManager.cpp
```diff
@@ -183,6 +198,16 @@ bool WalletManager::addressValid(const QString &address, bool testnet) const
     return Monero::Wallet::addressValid(address.toStdString(), testnet);
 }
 
+bool WalletManager::keyValid(const QString &key, const QString &address, bool isViewKey,  bool testnet) const
+{
+    std::string error;
+    if(!Monero::Wallet::keyValid(key.toStdString(), address.toStdString(), isViewKey, testnet, error)){
+        qDebug() << QString::fromStdString(error);
```
moneromooo-monero: That just prints on stdout I guess, not in the log.

wizard/WizardManageWalletUI.qml
```diff
@@ -29,12 +29,18 @@
 import QtQuick 2.2
 import moneroComponents.TranslationManager 1.0
 import QtQuick.Dialogs 1.2
-import QtQuick.Layouts 1.1
+import QtQuick.Layouts 1.2
 import "../components"
+import 'utils.js' as Utils
 
-// Reusable component for managing wallet (account name, path, private key)
+// Reusable component for mnaging wallet (account name, path, private key)
```
moneromooo-monero: wat

wizard/WizardManageWalletUI.qml
```diff
@@ -45,15 +51,45 @@ Item {
...
+        addressLine.error = !addressOK && addressLine.text.length != 0
+        viewKeyLine.error = !viewKeyOK && viewKeyLine.text.length != 0
+        spendKeyLine.error = !spendKeyOK && spendKeyLine.text.length != 0
+
+        return addressOK && viewKeyOK && spendKeyOK
+    }
+
+    function checkNextButton(){
+        wizard.nextButton.enabled = false
+        console.log("check next", recoverFromSeed.visible)
```
moneromooo-monero: Third logging system ?

wizard/WizardManageWalletUI.qml
```diff
@@ -45,15 +51,45 @@ Item {
...
+        if(recoverMode && !recoverFromSeedMode) {
+            console.log("checking key fields")
+            wizard.nextButton.enabled = checkFields();
+        } else if (recoverMode && recoverFromSeedMode) {
+            console.log("Checking seed")
+            var wordsArray = Utils.lineBreaksToSpaces(uiItem.wordsTextItem.memoText).split(" ");
+            wizard.nextButton.enabled = wordsArray.length === 25;
```
moneromooo-monero: I'd put this check in a separate function (ie, is_seed_ready or something) so it's obvious, as it seems likely to change next time we add things to the seed (eg, creation date, etc).

wizard/WizardMemoTextInput.qml
```diff
@@ -24,15 +24,27 @@ Column {
...
             property int minimumHeight: 100
             height: contentHeight > minimumHeight ? contentHeight : minimumHeight
+
+            Text {
+                id: memoTextPlaceholder
+                anchors.fill:parent
+                font.pixelSize: 16
+                anchors.margins: 8
+                font.bold:true
+                text: "Enter your 25 word mnemonic seed"
```
moneromooo-monero: qsTr

src/libwalletqt/WalletManager.cpp
Jaqueeee: Not sure if i understand what you mean. If wallet is created with error it will be deleted by caller. https://github.com/monero-project/monero-core/pull/437/files/83f617e4861dbb6abdbeb678db8d14f9ef963dfa#diff-15b9367dcbcc645d6ebb7fd91e46e6e1R98

src/libwalletqt/WalletManager.cpp
Jaqueeee: That's right. Need to catch those msgs and direct them to the logger. Haven't researched how yet.

wizard/WizardManageWalletUI.qml
Jaqueeee: fixed

wizard/WizardManageWalletUI.qml
Jaqueeee: done

wizard/WizardMemoTextInput.qml
Jaqueeee: done

src/libwalletqt/WalletManager.cpp
moneromooo-monero: I meant the pointer is not function local, but class local, and will be pointing to a deleted object, so I'd set it to NULL if an exception can be thrown before it is reset in this function.

src/libwalletqt/WalletManager.cpp
moneromooo-monero: Doesn't MERROR(x) work ?

src/libwalletqt/WalletManager.cpp
Jaqueeee: ah. It's actually a guarded pointer, so I don't think it's needed?
http://doc.qt.io/qt-5/qpointer.html

src/libwalletqt/WalletManager.cpp
Jaqueeee: Not out of the box. I tried to include contrib/epee/include/misc_log_ex.h but got build errors from boost. I'll investigate it further outside this PR.

src/libwalletqt/WalletManager.cpp
moneromooo-monero: I doubt that's OK. I'd just set this to NULL after delete, so any dereference will be easy to catch.

moneromooo-monero:
ok

PR #440, opened 2017-01-29
IOS build settings

Jaqueeee:


main.cpp
```diff
@@ -46,13 +46,17 @@
 #include "TransactionHistory.h"
 #include "model/TransactionHistoryModel.h"
 #include "model/TransactionHistorySortFilterModel.h"
-#include "daemon/DaemonManager.h"
 #include "AddressBook.h"
 #include "model/AddressBookModel.h"
 
+// IOS exclusions
+#ifndef Q_OS_IOS
```
MoroccanMalinois: IMHO, I think it would be easier to maintain with a "feature" define instead of "os" define. (_wants to use these modifs for android but super lazy :)_ )

fluffypony:
Please rebase

Jaqueeee:
rebased

main.cpp
Jaqueeee: Good idea. Didn't get it to work with the "DEFINES+=" method you used for the scanner for some reason. I'll try again some other day and create a new PR if i get it to work.

PR #456, opened 2017-02-04
main: mixin 4 is ring size 5, not 41...

moneromooo-monero:


medusadigital:
https://github.com/monero-project/monero-core/issues/457

PR #463, opened 2017-02-05
Send all debug output to easylogger

Jaqueeee:
requires #1685

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -49,15 +49,23 @@
 #include "model/TransactionHistorySortFilterModel.h"
 #include "AddressBook.h"
 #include "model/AddressBookModel.h"
+#include "wallet/wallet2_api.h"
 
 // IOS exclusions
 #ifndef Q_OS_IOS
 #include "daemon/DaemonManager.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


fluffypony:
Please rebase

main.cpp
```diff
@@ -50,10 +50,18 @@
 #include "daemon/DaemonManager.h"
 #include "AddressBook.h"
 #include "model/AddressBookModel.h"
+#include "wallet/wallet2_api.h"
 
+void messageHandler(QtMsgType type, const QMessageLogContext &context, const QString &msg)
+{
+    // Send all message types to logger
+    Monero::Wallet::debug(msg.toStdString());
+}
```
moneromooo-monero: I would add another parameter here, the category. Otherwise all of these will end up with the generic category defined in src/wallet/api/wallet.cpp (maybe wallet or default). But maybe that's enough ? If you think it is, I'm fine with it.

Jaqueeee: OK for now with same category as wallet API (WalletAPI). Could update that later on monero side.

PR #472, opened 2017-02-14
Create monero-core_eo.ts

dnaleor:
Esperanto translation


Jaqueeee:
@dnaleor  More strings have been added to GUI since you added this. Please update translation when you have time.

PR #479, opened 2017-02-18
translations: use &amp; instead of raw &

moneromooo-monero:


moneromooo-monero:
Already fixed in a pending PR.

PR #487, opened 2017-02-24
fix daemon startup with --check-updates disabled

Jaqueeee:


Jaqueeee:
closing in favour of #491

PR #494, opened 2017-02-25
Daemon rpc login support

Jaqueeee:
requires https://github.com/monero-project/monero/pull/1803


Jaqueeee:
@fluffypony don't merge until https://github.com/monero-project/monero/pull/1803 is merged or not.

Jaqueeee:
Note that plain text user/password is stored in settings file currently. @vtnerd

PR #495, opened 2017-02-26
Activate Bahasa Indonesia translation with flag

NanoAkron:


fluffypony:
This looks like we're merging support for Pokemon Go :-P

Issue #379, opened 2016-12-31
Stylistic color bar is covered up and titles are uneven


ghost:
**Old build** (Color bar exists and "Monero" & "Amount" are vertically centered):
 
![goodd](https://cloud.githubusercontent.com/assets/21302237/21579347/15fb0820-cf78-11e6-91fa-1279d53f4fcc.png)

**Current build** (Corners of the color bar poking out the sides, "Monero" and "Amount" uneven):

![badd](https://cloud.githubusercontent.com/assets/21302237/21579346/07a1d2e0-cf78-11e6-8bb6-ea80490d2940.png)

skftn:
Old user interface. 0.12.2 features the black theme.

+resolved

PR #498, opened 2017-02-27
Transfer: hide advanced options

Jaqueeee:


xmr-eric:
Looks great. Only quibble is that the Transfer tool feels a bit squished at the top with all that open space beneath it. Could be fixed via #379, but no biggie for now.

Jaqueeee:
@xmr-eric +1.  top margin added

PR #502, opened 2017-02-27
Remove reference to Taiwan from the codebase, replace with 'Traditional'

NanoAkron:
May seem petty, but it's a real political sticking-point in China

We're using the same 'Chinese' (mainland) flag for both traditional and simplified writing systems, so once someone writes a simplified translation file, we need to rename the translation files appropriately, and list them in the menu in their native Chinese names.


NanoAkron:
Don't merge yet. I'm going to rename 'Mandarin' to 'Chinese (Traditional)' which is more correct as multiple Chinese languages can be read with the same character system. 

However, may I suggest that with non-Roman alphabets we use the native names (so it would say 'Chinese' but in traditional characters so as to be obvious for a native reader)

PR #504, opened 2017-02-27
Keep text mixed case

xmr-eric:


xmr-eric:
Tried reverting PrivacyLevelSmall since mixed case is already in #506, but I accidentally added another commit...

Jaqueeee:
@xmr-eric if you do 
`git rebase -i origin/master`
you can replace 'pick' with 'drop' or 'd' to remove the commit you don't want. In this case you could also use 's' to squash the commit into previous commits.

fluffypony:
Merging it anyway - please squash for next time :)

PR #512, opened 2017-03-01
Many fixes and additions to the dutch translation

rpcjacobs:
These additions and fixes should complete the dutch translation. Made sure to have uniform translations and well created sentences (which was lacking).

Would be great if the dutch flag could be added as well!

### DIFF
--- a/translations/monero-core_nl.ts
+++ b/translations/monero-core_nl.ts
@@ -1165,7 +1165,7 @@ Graag updaten of verbind met een andere node</translation>
     <message>
         <location filename="../wizard/WizardFinish.qml" line="128"/>
         <source>You’re all set up!</source>
-        <translation>Je bent klaar met instelling!</translation>
+        <translation>U bent klaar met instellen!</translation>
     </message>
     <message>
         <source>You’re all setup!</source>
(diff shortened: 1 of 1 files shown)


rpcjacobs:
You're totally right, goofed that one up. I've added a fix.
Anyhow, despite Dutch not being your native language, well done on the translations!

When a working build with Dutch is available, i'll give it another good look to see if everything is correct in the context of the GUI-options. Unfortunately couldn't get Msys2, QT and Mingw32 to play along and work together for easy development (win10). Any tips/guide to get that working are welcome!

Jaqueeee:
@rpcjacobs what issues did you run into more specifically on windows?

rpcjacobs:
A lot, mainly building the GUI itself, but also tried to hook everything together in Qt Creator. But that didn't work either. My msys2 environment should be good (fully updated), have built many things with it. Qt recently updated. But i'm having this problem of building the gui on windows for quite a while now. Works fine on other platforms (e.g. MacOSX) . I have wondered many times if the (win-)guide needs updating. But it might also be the combination of tools i have installed, e.g. %PATH% setup.

Although i did notice a couple of new things, which previously didn't appear, even though the build process didn't work back then either!

- Running the ./get_libwallet_api.sh command via msys2 shell (either msys/mingw32/mingw64), reports 'dashless' errors (related https://github.com/Alexpux/MSYS2-packages/issues/735)

- Building the gui fails, reporting 'dict-src.h' couldn't be found. But also after generating it manually, the build still fails.

Unfortunately don't have the build log available atm.

Jaqueeee:
@rpcjacobs I also have the dashless errors. Doesnt seem to affect the build though. 

Do you use the build.sh script?  if not, make sure you also build src/zxcvbn-c. After you have run build.sh once you can switch to building with Qt Creator if you prefer.

Issue #510, opened 2017-02-28
OSX: Daemon does not shut down if GUI is closed


xmr-eric:
The first time the daemon is started, it starts fine. But if I "X" out of the GUI in OSX 10.12.3 and open it a few hours later, the wallet doesn't seem to start the daemon. I then clicked "Show Status" in Settings and saw that the daemon had been running for over two hours, even though the GUI was closed.

Jaqueeee:
This is intentional. You need to close the daemon manually if you don't want it running.

xmr-eric:
Really? I don't know if I like this. This means anybody who runs the GUI will (without their knowledge) be turning their computers into full time nodes?

NanoAkron:
This is a non obvious behaviour for non experts. I think it should change to shut down automatically but present an option somewhere to 'leave daemon running in background after gui shuts down'

Jaqueeee:
@xmr-eric @NanoAkron 
Added a confirmation popup on exit in #515 to let the user choose to shutdown daemon or keep it running in bg.

xmr-eric:
@jaqueeee You are the man!!

xmr-eric:
@Jaqueeee Side question, why was the functionality added to keep the daemon running in the background after the GUI is closed? Multiple GUI instances or something?

Jaqueeee:
There was a discussion in #monero-dev and also this request https://github.com/monero-project/monero-core/issues/484

IMO it's more convenient having it running in the background. Minimal sync time when starting up wallet.

PR #515, opened 2017-03-01
Confirm daemon running in background on exit

Jaqueeee:
works on mac. needs to be tested on windows and linux @medusadigital


Jaqueeee:
fixes #510

xmr-eric:
Works

PR #533, opened 2017-03-03
Update Ubuntu 16.04 x64 dependencies

amiuhle:
I needed this on Ubuntu 16.10 x64, so I'm guessing it'll be necessary on 16.04 also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Packaging for your favorite distribution would be a welcome contribution!
 
   - For Ubuntu 16.04 x64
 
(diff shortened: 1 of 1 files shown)


amiuhle:
Is there something like `[skip-ci]` for changes like this?

Jaqueeee:
Don't think there is a command to skip builds. @danrmiller might know?

danrmiller:
> Is there something like [skip-ci] for changes like this?

@amiuhle I'm going to use a change_filter in buildbot for ignoring changes to files not related to the build. Do you use irc, and/or python, maybe I can run some things by you? I tried here for kovri but still have some testing to do: https://github.com/monero-project/meta/commit/c2c16bc890a7ea4737d8bb348d77a6d288508503

amiuhle:
Yep, I'm on IRC, just let me know what I can do.

PR #534, opened 2017-03-03
Adjust top margin when custom decorations are used

### LINKED ISSUE #354: Better top margin for left panel with custom decorations
Currently the custom decorations setting is overlaying on top of the left panel, rather than pushing it down. This gives the word "MONERO" an uneven margin all around it, which looks bad. Either the title bar should stack better on top of the wallet (rather than rendering within it), or the left panel should be given a wider top-margin if the custom decorations option is checked.

monero-wallet-90bdc7

Jaqueeee:
Fixes #354


xmr-eric:
Works

PR #540, opened 2017-03-04
Windows-build and guide fixes

rpcjacobs:
Added some fixes for building the GUI on windows (through msys2-mingw32), since currently that doesn't work as the guide prescribes. I've added the fixes in a way, which shouldn't mess with build environments already customized and the build-bot.  Also changed the windows build guide in respect to the changes.

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -147,23 +147,29 @@ win32 {
     # Win64 Host settings
     contains(QMAKE_HOST.arch, x86_64) {
         message("Host is 64bit")
-        MSYS_PATH=c:/msys64/mingw32
+        MSYS_PATH=c:/msys64/mingw64
+	MSYS_MINGW_PATH=/mingw64
 
         # boost root path
         BOOST_PATH=c:/msys64/mingw64/boost
+	BOOST_MINGW_PATH=/mingw64/boost
 
     # WIN32 Host settings
     } else {
         message("Host is 32bit")
         MSYS_PATH=c:/msys32/mingw32
+	MSYS_MING_PATH=/mingw32
 
         # boost root path
         BOOST_PATH=c:/msys32/mingw32/boost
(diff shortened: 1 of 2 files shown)


rpcjacobs:
@Jaqueeee ,the above fixes have been tested a lot, and work splendid!

monero-wallet-gui.pro
```diff
@@ -144,26 +144,36 @@ CONFIG(WITH_SCANNER) {
 
 win32 {
 
+    # QMAKE_HOST.arch is unreliable, will allways report 32bit if mingw32 shell is run.
+    # Obtaining arch through uname should be reliable. This also fixes building the project in Qt creator without changes.
+    MSYS_HOST_ARCH = $$system(uname -a | grep -o "x86_64")
+
+    # Even if host is 64bit, we are building win32 spec, so mingw path should allways be "/mingw32"
```
danrmiller: @rpcjacobs What does this mean, building "win32 spec"? Previously my static boost libs were in /mingw64/boost/lib and aren't found with this change. Its no big deal, I'll adjust but I'm curious.

Jaqueeee:
@rpcjacobs @danrmiller @fluffypony 
I reverted the parts that messed up the win64 build in #544. We could reapply them when we have updated the buildbot and fluffys build machines. But i'd prefer doing this after beta 2 release.

monero-wallet-gui.pro
rpcjacobs: Because https://github.com/monero-project/monero-core/blob/master/monero-wallet-gui.pro#L145 species a win32 build.

The code you are quoting checks if the host in win32 or win64, as to make sure the msys path is correct (either c:/msys64 or c:/msys32)

Previously `contains(QMAKE_HOST.arch, x86_64) {` would always detect the host machines as win32, even when it is not.

PR #544, opened 2017-03-08
Revert "Fixes reliable building in both mingw32 shell as Qt creator."

Jaqueeee:
This reverts commit 630a852669118b8dab86ee9a311caf770e878124.

We can't use the official mingw binaries from Qt because it causes random crashes when combined with custom built boost. Therefore we use Qt built from source instead.  

630a852669118b8dab86ee9a311caf770e878124 can be reapplied after beta2 with appropriate exceptions for buildbot/official builds.

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -144,29 +144,25 @@ CONFIG(WITH_SCANNER) {
 
 win32 {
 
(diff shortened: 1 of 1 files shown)


Jaqueeee:
closed in favor of #545

PR #549, opened 2017-03-09
README - Updating Mac OSX Build notes and adding heading.

jonathancross:
I had no problems building on OSX using homebrew using the latest version of Qt (v`5.8`).
Just wanted to update these build notes as brew is much easier than downloading from qt.io

Also added a heading so that it is clear this this the GUI project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# Monero GUI
+
 Copyright (c) 2014-2017, The Monero Project
 
 ## Development Resources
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
+# Monero Core GUI
```
amiuhle: How about just **Monero GUI**?

I've gotten used to it, but it is kind of confusing that the GUI repo is called `monero-core` and the monero core project is called `monero`. Might as well be clear in the README about what's what :)

Jaqueeee: +1

README.md
```diff
@@ -1,3 +1,5 @@
+# Monero GUI
```
jonathancross: Now is just "Monero GUI"

README.md
jonathancross: Soudns good, now is just **Monero GUI**

PR #555, opened 2017-03-13
Correct the lan flag for zh_tw and translation update

Lafudoci:
The china flag should be used for simplified Chinese translation. But we don't have that translation yet.
So I made new flag for traditional Chinese translation (zh_tw).


Jaqueeee:
nice. But please squash those commits
```
rebase -i origin/master
```
replace pick with squash (or 's'), except the top commit. 
and the force push: `git push -f`

Lafudoci:
Sorry, I really not good at git. I updated these file by webpage instead of CLI. So actually I don't know how to squash these commits with that command.

NanoAkron:
I hate to say this, but can we please not use the Taiwanese flag to indicate traditional chinese?

Hong Kong and Macau are far more politically neutral issues.

danrmiller:
Well that shows how using the flag of a state to represent a language is flawed (but I don't have a better idea). But even so, who decides which side is OK to "offend" and which isn't? And what potential consequences could there possibly be for having a controversial flag?

NanoAkron:
Chinese market big, Taiwanese market small

danrmiller:
Size of a "market" isn't really enough information to know how the subset of those people that might use monero might feel and whether they would feel strongly enough to notice at all, and if so then to not use the software for this reason. Is there any precedent for projects losing users or contributers because of a flag in a setup configuration wizard representing a language that they didn't even use?

NanoAkron:
http://www.ft.com/cms/s/2/d07011b8-d9d6-11d9-b071-00000e2511c8.html?ft_site=falcon&desktop=true

http://www.philstar.com/world/2017/02/07/1669934/cambodian-pm-citing-one-china-policy-bans-taiwan-flag

https://www.theguardian.com/us-news/2015/jan/05/china-denounces-raising-taiwanese-flag-washington

Taiwan and its flag is a pretty touchy subject in China and its allies

danrmiller:
Well I know I'm bikeshedding, but I want to register my ridicule.

Lafudoci:
Taiwan (ROC) is now the only one offilcially use traditional Chinese. Hong Kong, Macau and even Malaysia already start using simplified Chinese as China. Actually the traditional Chinese also has some difference between Taiwan and those place. So I'll happy to see somebody will also provide Hong Kong translation. The zh_tw translation I provided in GUI now is commonly used in Taiwan. I don't think it's a monopolization. Just because we don't have other lang likes zh_hk yet.

Lafudoci:
I think we can change the title of flag, now that is "Chinese (traditional Chinese)".
But "Traditional Chinese (Taiwan)"="繁體中文 (台灣)" is better.
Flag present local translation has less controversy?
So then we'll have these whatever lan-area combination they want like:
"Traditional Chinese (Hong Kong)"="繁體中文 (香港)"
"Simplified Chinese (China)" "簡體中文 (中國)"

Lafudoci:
Finally I understand how to use rebase now, haha. 
I squashed all commits into 2 commits, one for flag, one for translation update.

PR #556, opened 2017-03-14
Add missing, fix and improve translations

rpcjacobs:
Adds some missing translations and fixes/improves some display bugs.


xmr-eric:
Would it be possible to split this up into two different PRs, one for translations and one for display bugs?

rpcjacobs:
Maybe i should have noted, but the display bugs i'm referring to have to do with the translations. Either not working or displayed without adjusting the width.

With this PR every single string should be translateble and showed properly.

PR #562, opened 2017-03-17
Improve text wrap on mining page

### LINKED ISSUE #536: Dynamic text formatting for Mining tab
Right now the description on the Mining page reads very haltingly. It has five short sentences that are broken up with line breaks because in compact view the words would run off the side.

This page needs dynamic resizing like the Check Payment page, which has a long sentence that automatically breaks down to the next line when it hits the edge of the page.

mining

The description on Mining could then be two paragraphs instead of five (this is new copy):

**Mining with your computer helps strengthen the Monero network. The more mining is done, the harder it is for the network to be attacked, and every little bit helps.**

**Mining also gives you a small chance to earn some Monero. Your computer will create hashes looking for block solutions. If you find a block, you will get the associated reward. Good luck!**

sir-pinecone:
Fixes #536 

Some screenshots:

default
compact

### DIFF
--- a/pages/Mining.qml
+++ b/pages/Mining.qml
@@ -50,7 +50,9 @@ Rectangle {
     /* main layout */
     ColumnLayout {
         id: mainLayout
-        anchors.margins: 10
+        anchors.margins: 40
+        anchors.bottomMargin: 10
+
         anchors.left: parent.left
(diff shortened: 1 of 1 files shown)


xmr-eric:
This is great

PR #569, opened 2017-03-18
Better copy for mining page

xmr-eric:
screen shot 2017-03-18 at 7 32 01 pm


xmr-eric:
Typo in the screengrab "The that people mine" is correct in the actual commit "The more that people mine".

PR #577, opened 2017-03-19
Update monero submodule to new tag

Jaqueeee:


xmr-eric:
Built from scratch with just #577. Test tx sent successfully.

PR #582, opened 2017-03-20
Updated translation files and improved dutch

rpcjacobs:
Updated translation files for recent code changes.
Also improved dutch translations.

On  a side note, i think it would be wise to add some notes, on updating translations. Because each code change, changes the line numbers inside the files and each string change (might) invalidate previous completed work.

As to make sure translators don't miss the opportunity to add missing translations, a regular update of the files would be a good thing (as opposed to have them PR yet again for minor changes.

@fluffypony and @Jaqueeee  Is it possible to add a hook/integration for his purpose? The command which should run is 'lupdate -verbose monero-wallet-gui.pro' and it includes some logging. Or am i overlooking a issue with this idea?

### DIFF
--- a/translations/monero-core_ar.ts
+++ b/translations/monero-core_ar.ts
@@ -87,7 +87,7 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../components/AddressBookTable.qml" line="106"/>
+        <location filename="../components/AddressBookTable.qml" line="104"/>
         <source>Payment ID:</source>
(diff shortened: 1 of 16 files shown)


Jaqueeee:
@rpcjacobs I think that would be a good idea. Maybe we could set up a bot that updates translations and creates a PR automatically on a regular basis. 

However, I don't think it matters if only the line numbers change. The translations would still be valid. Or have you noticed they're not?

PR #585, opened 2017-03-20
Hide processing splash when transfer fails

### LINKED ISSUE #576: Spinning "Creating transaction" box
Let's say I load a wallet with an insufficient amount of money: 0.007 XMR and then I try to send 0.01 from it to another address. It will give me an insufficient balance message, then spins a Creating transaction box.

Video: https://youtu.be/EJO9QdJ7YY0?t=10s

Jaqueeee:
fixes #576


xmr-eric:
You rock

PR #586, opened 2017-03-21
Main.qml: Capitalize Insufficient funds message

xmr-eric:


Jaqueeee:
Not sure if it's a good idea to merge these kind of small typo fixes before the release, since they could invalidate the translations (needs to be tested). It could be worked around by updating the source string in the translations files at the same time. @xmr-eric @rpcjacobs

fluffypony:
Plz squash commits, don't need to have them all in separate commits:)

xmr-eric:
@fluffypony How do I do that? I've only used Github.com via my browser to edit the repository. Do I have to download software?

xmr-eric:
Figured it out :)

PR #592, opened 2017-03-22
fix language codes

Jaqueeee:


Jaqueeee:
closing. Patch included in #593

Issue #602, opened 2017-03-24
Automatically Connect to Open Remote Nodes in GUI


SamsungGalaxyPlayer:
Many users have asked for an easier way to use a remote node in the GUI. Other users have expressed concerns for having and maintaining a "master list" of nodes somewhere, and while a plugin (for example: a MoneroWorld Nodes plugin) can be helpful, it still takes the user an extra step and relies on an extra party.

Gingeropolous wrote [moneriote](https://github.com/Gingeropolous/moneriote) for scanning the network for open nodes. Can this be added to the GUI, in the initial configuration settings and settings panel, to have the service provide the option to scan the network for open nodes and automatically connect to a random one?

In the initial GUI config, show three options: 1) I would like to use a local node on this computer (safest), 2) I would like to connect to a specific remote node by URL, 3) I would like to connect to a random open node on the network

Some advantages of this method:

1. No need to keep or maintain a list of open nodes. Users can find them for themselves with minimal effort

2. May make it easier for some users to connect to a remote node, since they can do so directly from the GUI instead of searching for other sources (notably MoneroWorld)

3. By selecting one at random, it can reduce the impact of a single node registered to a centralized directory

Some disadvantages of this method:

1. I have not tested the performance of moneriote. It might not be reasonable for a normal user to use in its current form. It may take too long, use too much bandwidth, etc.

2. An attacker can control a large number of these open nodes, but they can already do the same for centralized lists. For instance, suppose an attacker launched FastMoneroNodez that many people used. It's probably more risky to have only a handful of centralized node lists imo.

3. People may not want to have their node process the requests of these users. To address this, the moneriote currently searches for the nonstandard RPC port 18089. We can keep using this port.

I'd love to hear some of your thoughts on this.

Jaqueeee:
related to #572

SamsungGalaxyPlayer:
[Here is a crude markup](http://i.imgur.com/KeAcWtUl.png).

loopyman:
I personally do not trust the idea of selecting a "random" node even if they are vetted prior or "approved". However if I could create my own list of my trusted nodes like a favorites/bookmark list then I would be just as happy.

xmr-eric:
I would prefer a dev meeting discussion about this (people smarter than myself)

SamsungGalaxyPlayer:
I know that connecting to a random node is not ideal. Running your own node is ideal. 

[...]

Issue #572, opened 2017-03-19
Feature Request - Smart Nodes/Wallet


DaveyJonesXMR:
While surfing our reddit i stumbled over some comment that brought me to this request. Can`t find it sadly.

Request:

Implement smart-nodeing - former means that at first and future sync up the wallet automatically will use a remote node ( probably hosted by core or sth like node.moneroworld.com ) to connect at start so especially new users can have a wallet user experience right from the start, but also starts syncing a local node in the background and once fully synced will switch to the local node, 

This way especially people with slower HDD will still have a easy first-time  or once-in-a-month user-experience, on the other hand it will enhance our node network when people leave those wallets/nodes open because of the ease of use and may generally attract people that would not have the time/nerves to go through a full-sync at start.

As usual their should be disclaimers in the software about disk-usage, ram-usage etc.

Jaqueeee:
I like this idea, but I'd like to have some input from @fluffypony and other core devs before implementing this. 

My suggestion is that we add a page in the wizard after creating a new wallet where we explain in 2-3 sentences why the user should run a local node. And after that 2 checkboxes:

1. (checked) Run local daemon. + info about the drawbacks. i.e diskspace, network usage, time until wallet can be used. 

2. (unchecked) Connect to remote daemon while local daemon in syncing. + info about the lesser privacy. 

It would be great if @xmr-eric or any other native speaker could provide the copy for this page.

hyc:
We definitely need to get away from the notion of a "wallet" being separate from a "daemon." I.e., the part of the wallet that interacts with a daemon today, should be running 24/7 and should always have a local daemon (also running 24/7) to talk to. Everything else is just UI, talking via RPC to the wallet server. That solves our mobile use case as well as every other PC use case.

In the meantime - yes, we can explain why the user should run a local node. But for usability, we're going to need to separate the UI from the wallet sync and cache files. Nobody wants to wait for their wallet to sync over 3G.

Jaqueeee:
@DaveyJonesXMR this was discussed today at [dev meeting](https://monerobase.com/wiki/DevMeeting_2017-03-26) where @fluffypony made clear #605 is not going to be merged. 
Also #605

dEBRUYNE-1:
+in progress

iBobik:
[...]

PR #605, opened 2017-03-24
Remote nodes

Jaqueeee:
See discussion in #572 and #602 

Not ready to be merged yet.  

- Added page in wizard for daemon settings
- Added option to automatically connect to remote node while syncing local node
- added node.xmrbackb.one and node.moneroworld.com to dropwdown in wizard and settings

Ping @Gingeropolous, Snipa (github user?)

### DIFF
--- a/components/NetworkStatusItem.qml
+++ b/components/NetworkStatusItem.qml
@@ -51,6 +51,8 @@ Row {
         if (status == Wallet.ConnectionStatus_Connected) {
             if(!appWindow.daemonSynced)
                 return qsTr("Synchronizing")
+            if(appWindow.remoteNodeConnected)
+                return qsTr("Remote node")
             return qsTr("Connected")
         }
         if (status == Wallet.ConnectionStatus_WrongVersion)
--- a/components/ProgressBar.qml
+++ b/components/ProgressBar.qml
@@ -37,7 +37,7 @@ Item {
     visible: false
     //clip: true
 
-    function updateProgress(currentBlock,targetBlock, blocksToSync){
+    function updateProgress(currentBlock,targetBlock, blocksToSync, statusTxt){
         if(targetBlock == 1) {
             fillLevel = 0
             progressText.text = qsTr("Establishing connection...");
@@ -54,7 +54,14 @@ Item {
             else
                 var progressLevel = (100*(currentBlock/targetBlock)).toFixed(0);
             fillLevel = progressLevel
-            progressText.text = qsTr("Blocks remaining: %1").arg(remaining.toFixed(0));
+            if(typeof statusTxt != "undefined" && statusTxt != "") {
+                progressText.text = statusTxt + (" %1").arg(remaining.toFixed(0));
+            } else {
+                progressText.text = qsTr("Blocks remaining: %1").arg(remaining.toFixed(0));
+            }
+
+
+
             progressBar.visible = currentBlock < targetBlock
         }
     }
--- a/components/RemoteNodeDropdown.qml
+++ b/components/RemoteNodeDropdown.qml
@@ -34,23 +34,23 @@ import QtQuick.Layouts 1.1
 RowLayout {
     Layout.preferredWidth: 200
     function getSelected() {
-        return remoteNodeList.get(remoteNodeDropdown.currentIndex).url;
+        return daemonAddr.text + ":" + daemonPort.text
     }
 
-    ListModel {
-         id: remoteNodeList
-         ListElement { column1: "node.xmrbackb.one" ; column2: ""; url: "node.xmrbackb.one:18081" }
-         ListElement { column1: "node.moneroworld.com" ; column2: ""; url: "node.moneroworld.com:18081" }
-     }
-
-    StandardDropdown {
+    LineEdit {
+        id: daemonAddr
+        Layout.preferredWidth:  200
         Layout.fillWidth: true
-        id: remoteNodeDropdown
-        shadowReleasedColor: (parent.enabled)? "#FF4304" : "#c1c1bf"
-        shadowPressedColor: "#B32D00"
-        releasedColor: (parent.enabled)? "#FF6C3C" : "#c1c1bf"
-        pressedColor: "#FF4304"
-        dataModel: remoteNodeList
-        z: 1
+        text: persistentSettings.remoteNodeAddress.split(":")[0]
+        placeholderText: qsTr("Hostname / IP") + translationManager.emptyString
+    }
+
+
+    LineEdit {
+        id: daemonPort
+        Layout.preferredWidth: 100
+        Layout.fillWidth: true
+        text: persistentSettings.remoteNodeAddress.split(":")[1]
+        placeholderText: qsTr("Port") + translationManager.emptyString
     }
 }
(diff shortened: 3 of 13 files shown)


xmr-eric:
I am strongly against the idea of adding a dropdown list of remote nodes to the GUI. Here's my thoughts as posted recently on Reddit ([source](https://www.reddit.com/r/Monero/comments/614k9b/preview_of_some_changes_in_the_gui_beta_2_release/dfbsiyu/)):

>> how does having a drop down list make it centralized?

>Great question. It's the exact situation we struggled in the GUI with during the Dwarfpool hashrate centralization problem that happened in December. To combat the miner centralization, we rushed to include pool mining (not just solo mining) in the official GUI. In the end, we decided against it, because having a list of pools in the GUI would amount to a certain degree of centralization, along with all the problems inherent in that. Here's what I wrote at that time ([source](https://github.com/monero-project/monero-core/pull/368)):

>> Pool mining is the one page in the wallet where we're suddenly relying on a third party. What qualifies a pool to be added to this list? Certainly cryptmonero.com isn't included since this pool is tied to Reddit-reported malware mining. So there is already a distinction between good and bad pools, and suddenly our devs are tasked with making that distinction. Suddenly our work begins to grow exponentially, especially if XMR really takes off and a hundred new pools appear and have to be evaluated. Who does the evaluating and what is the criteria?
>> Suddenly pool operators have an incentive to get in good graces with XMR devs? Already we can see MoneroWorld (where I personally mine, so I love it there) is at the top of the list?
>> What if one of the listed pools gets hacked, malware, or leaks private data? Are we partly responsible / could be sued? We have to wait for a whole release before we can change the code away from an adversarial pool?
>>This is not decentralized, and thus sort of goes against the spirit of The Monero Project.

> Including a list of default remote nodes in the GUI leaves us with a similar (albeit milder) situation. It tasks us with maintaining a list of good and bad remote nodes, adding to the official dev workload. It creates vulnerabilities within the GUI, where if one node gets compromised then a large portion of Monero users are at risk simply by using the official software. And removing a compromised remote requires an entire software release, and not all users update their software religiously, so the problem could be with us for years.
> Decentralization is our credo. I'm personally open to compromising on this value, but it needs to be non-contentious and very thought out.

Snipa22:
I could see the complaint for this.  Perhaps there's something to be said here for the possibility of moving to SRV records?  I know that I would happily register my nodes in a central database that could maintain a status check on all of the nodes involved.

It would be a small set of changes to move such support into something like: https://github.com/monero-project/xmr-seeder (Which I already redid to CF support), and using SRV records, we could use non-standard ports that would be registered at the DNS servers.  Then an automation tool could manage these records, keeping an eye on the state (Height anyone...) of the nodes and knock them in and out automatically.

This would allow the DNS records to be controlled centrally by the Monero project, and would allow control without excessive centralization, then allow a manual punch-in override box.  Yes, it's still centralization, but it's no better/worse than the current DNS seeding system in my eyes.

xmr-eric:
As mentioned in #602 I think we should wait with integrating non-manual node support until we have a discussion with the wider dev community, perhaps include this as an item during the Sunday dev meeting. If people like it, the concept of scanning the p2p network for open remote nodes (or whatever mechanism we decide on) could be integrated at the protocol/CLI level, and the wallet could simply add pure GUI functionality (which, I think is all we've been doing anyways).

danrmiller:
Snipa22 wrote:

> but it's no better/worse than the current DNS seeding system in my eyes.

Suggesting a node to connect to is less severe than suggesting a node to leak all the outputs you are interested in to.

fluffypony:
I'm against this as I don't think we should be encouraging remote nodes at all - they place an incredible burden on the network by chewing up the same amount of downstream bandwidth as a full node, but with no upstream assistance to the network, and with no validation. People are better served either running a full node, or waiting for the light wallet / remote viewkey scanning stuff that we're building out. If people want to manually connect to a full node, that's fine, but we shouldn't make it easy:)

SamsungGalaxyPlayer:
@danrmiller see [here](http://monero.stackexchange.com/questions/38/what-privacy-or-security-trade-offs-are-associated-with-not-running-your-own-ful/86#86).

You no longer leak the outputs you are interested in, but you connect your IP address with your outgoing transactions, and you leak some other metadata.

Jaqueeee:
@fluffypony Don't really understand what you are saying here. There's no difference between a remote node and and a local node in p2p bandwidth usage. If you are referring to the wallet bandwidth usage it's ~98% less than a node after the recent wallet optimization work iirc. Please correct me if i'm wrong moneromooo-monero and @luigi1111. 

I do however fully understand the points raised by @xmr-eric about centralization.

The point of this PR is to improve the user experience by speeding up the time between download and first usage from several hours to seconds. As soon as the local node is synced the wallet automatically switch to the local node.

fluffypony:
@Jaqueeee a wallet needs all the block headers, and then it needs a copy of every new tx output as blocks come in. That's ~the same bandwidth as a full node, from an inbound perspective, just that it doesn't have to relay blocks are txs so it doesn't use upstream bandwidth.

maitscha:
Monero's principle is Security by Default. So the GUI should not encourage users to use untrusted (random) remote nodes.

dternyak:
This is ultimately a debate over optimizing for usability or privacy.

There is not a lack of a security by using untrusted remote nodes, but a
lack of privacy (and only some loss of privacy). Additionally, it wouldn't
need to be the default option, only an option.

Personally I believe the slight hit to privacy to be okay in this instance,
as *most* users will opt to use a remote node anyway, whether its through a
light-wallet or this. This just makes it easier for them to do it.

Riccardo's point about the node getting some extra usage is a good one, so
perhaps we don't want to encourage usage of remote nodes at such a level
that those nodes begin to suffer.

fluffypony:
Ok it's added to the meeting agenda for tomorrow's dev meeting, so we can discuss it then.

https://github.com/monero-project/meta/issues/58

MBuffenoir:
What about having a **normal** mode where everything is taken care of for you (automatically connecting to a load-balanced pool of remotes nodes) and an **advanced** mode (where more options granularity is available ?)

Jaqueeee:
@dternyak after recent wallet optimizations the bandwidth used by each wallet is not a big issue. However, the trust issue seems to be the biggest concern. I'll remove the dropdown by request from @fluffypony. People who want to use remote nodes will need search for them.

Jaqueeee:
Dropdown replaced with an empty input field.
+ added advanced daemon settings checkbox
+ renamed "start/stop daemon" to "start/stop local node"

Please test @xmr-eric and @medusadigital 

Question is if this PR adds any value to the user when they have to input the daemon address manually? I think it does because it adds a page in wizard where new users gets information about daemon/node setup. It also makes it easier to follow the status of the local node sync when connected to a remote node.

iamsmooth:
@fluffypony I don't understand how headers+outputs would be similar bandwidth to a full node. That would not include signatures nor range proofs, which make up most of the current bandwidth usage.

Eventually I think lightweight wallets (by which I mean lightweight in terms of bandwidth) are very possible with ~90% less bandwidth usage than full nodes (not even factoring in upstream). They can use the p2p and check PoW just like Bitcoin light wallets do. (This might require some protocol changes, but it is do-able). There was a previous objection to this in terms of a wallet with private keys in process memory making untrusted p2p connections, but that could be addressed with a separate "light deamon".

Whether that is desirable in terms of privacy and security is a different matter.

Jaqueeee:
Closing for now. Could bring up the discussion again after next beta.

PR #606, opened 2017-03-24
Update submodule to v0.10.3

Jaqueeee:
pulling the v0.10.3 didn't work on windows for some reason. Reverted to pinning submodule to commit hash instead.

### DIFF
--- a/get_libwallet_api.sh
+++ b/get_libwallet_api.sh
@@ -15,9 +15,9 @@ BUILD_LIBWALLET=false
 if [ ! -d $MONERO_DIR/src ]; then
     git submodule init monero
 fi
-#git submodule update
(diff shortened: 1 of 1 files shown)


xmr-eric:
Perhaps because the 0.10.3 tag was pulled?

Jaqueeee:
@xmr-eric Windows build has failed all day: https://build.getmonero.org/builders/monero-core-win64
The other targets works fine.

danrmiller:
The other targets could not pull the non-existent tag either, so the previously downloaded files were used. The 64-bit windows machine ran out of disk space and I removed the build directories, so this job had no previously downloaded files to fall back on.

Jaqueeee:
@danrmiller @xmr-eric you were right. closing

PR #638, opened 2017-03-29
Update monero-core_de.ts

louvetic:
Ich habe Zeilen 1-1042 Korrektur gelesen. // I've checked lines 1-1042.

### DIFF
--- a/translations/monero-core_de.ts
+++ b/translations/monero-core_de.ts
@@ -11,7 +11,7 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="61"/>
         <source>Address</source>
-        <translation>Addresse</translation>
+        <translation>Adresse</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="63"/>
@@ -21,7 +21,7 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="72"/>
         <source>QRCODE</source>
-        <translation type="unfinished"></translation>
+        <translation>QR-Code</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="94"/>
@@ -31,22 +31,22 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="103"/>
         <source>Payment ID &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</source>
-        <translation>Zahlungs-ID &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</translation>
+        <translation>Zahlungs-ID &lt;font size=&apos;2&apos;&gt;(optional)&lt;/font&gt;</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="105"/>
         <source>&lt;b&gt;Payment ID&lt;/b&gt;&lt;br/&gt;&lt;br/&gt;A unique user name used in&lt;br/&gt;the address book. It is not a&lt;br/&gt;transfer of information sent&lt;br/&gt;during the transfer</source>
-        <translation>&lt;b&gt;Zahlungs-ID&lt;/b&gt;&lt;br/&gt;&lt;br/&gt;Ein eindeutiger Benutzername aus&lt;br/&gt;dem Addressbuch. Wird nicht zum&lt;br/&gt;Transfer von Informationen verwendet</translation>
+        <translation>&lt;b&gt;Zahlungs-ID&lt;/b&gt;&lt;br/&gt;&lt;br/&gt;Ein eindeutiger Benutzername aus&lt;br/&gt;dem Adressbuch. Wird nicht zum&lt;br/&gt;Transfer von Informationen verwendet</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="117"/>
         <source>Paste 64 hexadecimal characters</source>
-        <translation type="unfinished"></translation>
+        <translation>64 hexadezimale Zeichen einfügen</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="126"/>
         <source>Description &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</source>
-        <translation type="unfinished"></translation>
+        <translation>Beschreibung &lt;font size=&apos;2&apos;&gt;(optional)&lt;/font&gt;</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="128"/>
@@ -56,27 +56,27 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="139"/>
         <source>Give this entry a name or description</source>
-        <translation type="unfinished"></translation>
+        <translation>Einen Namen oder eine Beschreibung festlegen</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="155"/>
         <source>Add</source>
-        <translation type="unfinished"></translation>
+        <translation>Hinzufügen</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="160"/>
         <source>Error</source>
-        <translation type="unfinished">Fehler</translation>
+        <translation>Fehler</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="163"/>
         <source>Invalid address</source>
-        <translation type="unfinished"></translation>
+        <translation>Ungültige Adresse</translation>
     </message>
     <message>
         <location filename="../pages/AddressBook.qml" line="167"/>
         <source>Can&apos;t create entry</source>
-        <translation type="unfinished"></translation>
+        <translation>Eintrag kann nicht erstellt werden</translation>
     </message>
     <message>
         <source>&lt;b&gt;Payment ID&lt;/b&gt;&lt;br/&gt;&lt;br/&gt;A unique user name used in&lt;br/&gt;the address book. It is not a&lt;br/&gt;transfer of information sent&lt;br/&gt;during thevtransfer</source>
@@ -140,19 +140,19 @@
     <message>
         <location filename="../components/DaemonConsole.qml" line="108"/>
         <source>Close</source>
-        <translation type="unfinished"></translation>
+        <translation>Schließen</translation>
     </message>
     <message>
         <location filename="../components/DaemonConsole.qml" line="119"/>
         <source>command + enter (e.g help)</source>
-        <translation type="unfinished"></translation>
+        <translation>Befehl + Enter (z.B. help)</translation>
     </message>
 </context>
 <context>
     <name>DaemonManagerDialog</name>
     <message>
         <source>Daemon doesn&apos;t appear to be running</source>
-        <translation type="vanished">Daemon scheint nicht aktiv zu sein</translation>
+        <translation type="vanished">Dämon scheint nicht aktiv zu sein</translation>
     </message>
     <message>
         <source>Start daemon</source>
@@ -165,17 +165,17 @@
     <message>
         <location filename="../components/DaemonManagerDialog.qml" line="84"/>
         <source>Starting Monero daemon in %1 seconds</source>
-        <translation type="unfinished"></translation>
+        <translation>Starte Monero-Dämon in %1 Sekunden</translation>
     </message>
     <message>
         <location filename="../components/DaemonManagerDialog.qml" line="106"/>
         <source>Start daemon (%1)</source>
-        <translation type="unfinished"></translation>
+        <translation>Starte Dämon (%1)</translation>
     </message>
     <message>
         <location filename="../components/DaemonManagerDialog.qml" line="123"/>
         <source>Use custom settings</source>
-        <translation type="unfinished"></translation>
+        <translation>Eigene Einstellungen verwenden</translation>
     </message>
 </context>
 <context>
@@ -259,7 +259,7 @@
     <message>
         <location filename="../pages/History.qml" line="273"/>
         <source>Filter</source>
-        <translation type="unfinished"></translation>
+        <translation>Filter</translation>
     </message>
     <message>
         <source>Incremental search</source>
@@ -305,7 +305,7 @@
     <message>
         <location filename="../pages/History.qml" line="329"/>
         <source>Type of transaction</source>
-        <translation>Art</translation>
+        <translation>Transaktionsart</translation>
     </message>
     <message>
         <location filename="../pages/History.qml" line="364"/>
@@ -359,17 +359,17 @@
     <message>
         <location filename="../components/HistoryTable.qml" line="266"/>
         <source>(%1/10 confirmations)</source>
-        <translation type="unfinished"></translation>
+        <translation>(%1/10 Bestätigungen)</translation>
     </message>
     <message>
         <location filename="../components/HistoryTable.qml" line="270"/>
         <source>UNCONFIRMED</source>
-        <translation type="unfinished"></translation>
+        <translation>UNBESTÄTIGT</translation>
     </message>
     <message>
         <location filename="../components/HistoryTable.qml" line="271"/>
         <source>PENDING</source>
-        <translation type="unfinished"></translation>
+        <translation>OFFEN</translation>
     </message>
     <message>
         <location filename="../components/HistoryTable.qml" line="300"/>
@@ -446,7 +446,7 @@
     <message>
         <location filename="../LeftPanel.qml" line="289"/>
         <source>Address book</source>
-        <translation type="unfinished"></translation>
+        <translation>Adressbuch</translation>
     </message>
     <message>
         <location filename="../LeftPanel.qml" line="290"/>
@@ -481,7 +481,7 @@
     <message>
         <location filename="../LeftPanel.qml" line="407"/>
         <source>Check payment</source>
-        <translation type="unfinished"></translation>
+        <translation>Zahlung prüfen</translation>
     </message>
     <message>
         <location filename="../LeftPanel.qml" line="431"/>
@@ -518,12 +518,12 @@
     <message>
         <location filename="../MiddlePanel.qml" line="45"/>
         <source>Balance</source>
-        <translation type="unfinished">Guthaben</translation>
+        <translation>Guthaben</translation>
     </message>
     <message>
         <location filename="../MiddlePanel.qml" line="47"/>
         <source>Unlocked Balance</source>
-        <translation type="unfinished"></translation>
+        <translation>Verfügbares Guthaben</translation>
     </message>
     <message>
         <source>Unlocked Balance:</source>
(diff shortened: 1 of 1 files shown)


translations/monero-core_de.ts
```diff
@@ -815,17 +815,17 @@
...
         <location filename="../pages/Receive.qml" line="385"/>
         <source>Save QrCode</source>
-        <translation type="unfinished"></translation>
+        <translation>QR-Code speichern</translation>
     </message>
     <message>
         <location filename="../pages/Receive.qml" line="386"/>
         <source>Failed to save QrCode to </source>
-        <translation type="unfinished"></translation>
+        <translation>Konnte QR-Code nicht speichern als </translation>
```
amiuhle: What's the context? Is this message about a path to which the QR Code couldn't be saved? In that case, **als** would be wrong here.

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <source>(only available for local daemons)</source>
-        <translation type="unfinished"></translation>
+        <translation>(nur für verfügbar bei lokalem Daemon)</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="86"/>
         <source>Mining with your computer helps strengthen the Monero network. The more that people mine, the harder it is for the network to be attacked, and every little bit helps.&lt;br&gt; &lt;br&gt;Mining also gives you a small chance to earn some Monero. Your computer will create hashes looking for block solutions. If you find a block, you will get the associated reward. Good luck!</source>
-        <translation type="unfinished"></translation>
+        <translation>Mit deinem Computer zu schürfen hilft dabei, das Monero-Netzwerk zu stärken. Je mehr Leute schürfen, desto schwieriger kann das Netzwerk angegriffen werden. Jeder kann helfen! &lt;br&gt; &lt;br&gt;Schürfen bringt dir auch eine kleine Chance, Monero zu verdienen. Dein Computer errechnet Hash-Werte für neue Blöcke. Wenn du einen neuen Block findest, bekommst du die Belohnung dafür. Viel Erfolg!</translation>
```
amiuhle: Should we translate Mining? *Wallet* isn't translated either, which I think is fine. Maybe change this to *Mit deinem Computer zu schürfen (Mining) hilft dabei* ... and then just go with Mining in the smaller translations?

How do other languages handle this?

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="96"/>
         <source>CPU threads</source>
-        <translation type="unfinished"></translation>
+        <translation>CPU-Kerne</translation>
```
amiuhle: CPU **Threads** -- *Kerne* would be *cores*

Not sure about the dash.

translations/monero-core_de.ts
```diff
@@ -998,17 +998,17 @@
...
         <location filename="../pages/Settings.qml" line="302"/>
         <source>Layout settings</source>
-        <translation type="unfinished"></translation>
+        <translation>Layout-Einstellungen</translation>
     </message>
     <message>
         <location filename="../pages/Settings.qml" line="319"/>
         <source>Custom decorations</source>
-        <translation type="unfinished"></translation>
+        <translation>Eigenes Design</translation>
```
amiuhle: *Design* isn't exactly right, but I don't have a better idea right now

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="73"/>
         <source>Solo mining</source>
-        <translation type="unfinished"></translation>
+        <translation>Solo schürfen</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="80"/>
         <source>(only available for local daemons)</source>
-        <translation type="unfinished"></translation>
+        <translation>(nur für verfügbar bei lokalem Daemon)</translation>
```
louvetic: You're right!

translations/monero-core_de.ts
louvetic: That's a good idea and really an elegant way to express it.

translations/monero-core_de.ts
louvetic: I've checked again some sources, _CPU-Threads_ would work in German as well, we've adopted "thread" and don't have an own German word.

translations/monero-core_de.ts
louvetic: Could be. Let's just wait and see? It's just a beta, isn't it? Or simply dismiss the change, would be perfectly okay for me, too!

translations/monero-core_de.ts
louvetic: Something like _Benutzerdefinierte Oberfläche_ or _Angepasste Verzierungen_ might work as well?

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="118"/>
         <source>Background mining (experimental)</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen im Hintergrund (experimentell)</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="134"/>
         <source>Enable mining when running on battery</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen erlauben bei Batterie-Betrieb</translation>
```
JamesCullum: Schürfen im Batterie-Betrieb erlauben

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="104"/>
         <source>(optional)</source>
-        <translation type="unfinished"></translation>
+        <translation>(optional)</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="118"/>
         <source>Background mining (experimental)</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen im Hintergrund (experimentell)</translation>
```
JamesCullum: Im Hintergrund schürfen

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="134"/>
         <source>Enable mining when running on battery</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen erlauben bei Batterie-Betrieb</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="144"/>
         <source>Manage miner</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen verwalten</translation>
```
JamesCullum: Sounds very weird, the original talks about a noun ("the" miner), while we use the verb. Probably "Schürfeinstellungen"?

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="153"/>
         <source>Start mining</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen starten</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="163"/>
         <source>Error starting mining</source>
-        <translation type="unfinished"></translation>
+        <translation>Fehler beim Starten des Schürfens</translation>
```
JamesCullum: Schürfen konnte nicht begonnen werden

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="164"/>
         <source>Couldn&apos;t start mining.&lt;br&gt;</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen konnte nicht gestartet werden&lt;br&gt;</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="166"/>
         <source>Mining is only available on local daemons. Run a local daemon to be able to mine.&lt;br&gt;</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen ist nur verfügbar bei lokalem Dämon. Starte einen lokalen Dämon, um selbst schürfen zu können.</translation>
```
JamesCullum: Schürfen ist nur bei lokalem Dämon verfügbar. Starte einen um selber schürfen zu können.

translations/monero-core_de.ts
```diff
@@ -535,87 +535,87 @@
...
         <location filename="../pages/Mining.qml" line="178"/>
         <source>Stop mining</source>
-        <translation type="unfinished"></translation>
+        <translation>Schürfen stoppen</translation>
     </message>
     <message>
         <location filename="../pages/Mining.qml" line="194"/>
         <source>Status: not mining</source>
-        <translation type="unfinished"></translation>
+        <translation>Status: Schürfen ist aus</translation>
```
JamesCullum: "Nicht aktiv" oder "Nicht am schürfen"

translations/monero-core_de.ts
```diff
@@ -691,17 +691,17 @@
     <message>
         <location filename="../components/PrivacyLevelSmall.qml" line="100"/>
         <source>Low</source>
-        <translation type="unfinished"></translation>
+        <translation>niedrig</translation>
```
JamesCullum: Uppercase

JamesCullum: Same for following

translations/monero-core_de.ts
```diff
@@ -751,22 +751,22 @@
     <message>
         <location filename="../pages/Receive.qml" line="138"/>
         <source>1 confirmation: %2 (%1)</source>
-        <translation type="unfinished">1 Bestätigung %2 (%1)</translation>
+        <translation>1 Bestätigung %2 (%1)</translation>
```
JamesCullum: "Eine" evtl besser?

translations/monero-core_de.ts
JamesCullum: I think its used for saving the QR-Code using the right mouse button, so "unter" instead of "als"

translations/monero-core_de.ts
```diff
@@ -954,22 +954,22 @@
     <message>
         <location filename="../pages/Settings.qml" line="102"/>
         <source>Create view only wallet</source>
-        <translation type="unfinished"></translation>
+        <translation>Schreibgeschätze Wallet erstellen</translation>
```
JamesCullum: *Schreibgeschützte (Tippfehler)

translations/monero-core_de.ts
JamesCullum: What about "Anpassungen"?

translations/monero-core_de.ts
```diff
@@ -140,19 +140,19 @@
...
-        <translation type="unfinished"></translation>
+        <translation>Befehl + Enter (z.B. help)</translation>
     </message>
 </context>
 <context>
     <name>DaemonManagerDialog</name>
     <message>
         <source>Daemon doesn&apos;t appear to be running</source>
-        <translation type="vanished">Daemon scheint nicht aktiv zu sein</translation>
+        <translation type="vanished">Dämon scheint nicht aktiv zu sein</translation>
```
louvetic: Wikipedia starts its respective entry with "Als Daemon oder Dämon (auch häufig in der Schreibweise Demon)". So any way of spelling is okay, we just have to think about which is more beautiful. In my opinion, when there is a German word, we shouldn't be shy using it. Most people not familiar with IT processes would guess that the "ae" means "ä" anyway - my mother might even believe it's a bad translation and wonder why she should let an evil spirit in her computer! ;) But in the end, I don't care. I just want to make progress in the project.

translations/monero-core_de.ts
```diff
@@ -31,22 +31,22 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="103"/>
         <source>Payment ID &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</source>
-        <translation>Zahlungs-ID &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</translation>
+        <translation>Zahlungs-ID &lt;font size=&apos;2&apos;&gt;(optional)&lt;/font&gt;</translation>
```
louvetic: Because it's not the beginning of the title or standalone, but follows other words bildschirmfoto 2017-03-30 um 08 27 51

translations/monero-core_de.ts
JamesCullum: I find the english word to be more aesthetic, but I understand that this basically comes down to taste. If we then agree on "ä", wouldn't we need to HTML-encode all those occurences? Should be easy by replacing all "ä" with "&auml;". And thanks for your work, it definitly is helping :+1:

translations/monero-core_de.ts
```diff
@@ -998,17 +998,17 @@
...
         <location filename="../pages/Settings.qml" line="302"/>
         <source>Layout settings</source>
-        <translation type="unfinished"></translation>
+        <translation>Layout-Einstellungen</translation>
     </message>
     <message>
         <location filename="../pages/Settings.qml" line="319"/>
         <source>Custom decorations</source>
-        <translation type="unfinished"></translation>
+        <translation>Angepasste Oberfläche</translation>
```
JamesCullum: Angepasste = already done, while I think this setting is supposed to be for allowing the user to change it. What about "Oberfläche anpassen"?

translations/monero-core_de.ts
```diff
@@ -954,22 +954,22 @@
     <message>
         <location filename="../pages/Settings.qml" line="102"/>
         <source>Create view only wallet</source>
-        <translation type="unfinished"></translation>
+        <translation>Schreibgeschütze Wallet erstellen</translation>
```
JamesCullum: "Schreibgeschützt" is really confusing in this context. View-only means that you can only see incoming transactions but can't spend anything, while "schreibgeschützt" implies that you later cant change the file or that it protects any write process on OS level. I understand the difficulty for a translation and maybe we are stuck with a weird translation, but what about something like "Bankauszug - Wallet" (heavily pointing out that you only see and cant do anything), "Beobachtende Wallet" or "Beschränkte Wallet"?

translations/monero-core_de.ts
```diff
@@ -518,12 +518,12 @@
...
         <location filename="../MiddlePanel.qml" line="45"/>
         <source>Balance</source>
-        <translation type="unfinished">Guthaben</translation>
+        <translation>Guthaben</translation>
     </message>
     <message>
         <location filename="../MiddlePanel.qml" line="47"/>
         <source>Unlocked Balance</source>
-        <translation type="unfinished"></translation>
+        <translation>Verfügbares Guthaben</translation>
```
JamesCullum: encoding necessary (here and at other places)?

translations/monero-core_de.ts
louvetic: On my Mac the checkbox just switches between the "regular" native window title bar and replaces it with a "custom" black title bar. Nothing else happens as far as I can see, you cannot even chose or specify - just accept what's already done / built-in or use the regular window title bar:

bildschirmfoto 2017-03-30 um 10 51 38

bildschirmfoto 2017-03-30 um 10 51 50

Maybe in other OS more magic happens with that checkbox?

translations/monero-core_de.ts
louvetic: I get your point, read-only is not writing-forbidden, true. "Beobachten" sounds too much like spying, "beschränkt" alone doesn't make clear, what the limitation is. Often you see the terms "Lesezugriff", "Leseberechtigung" or "nur lesen" in such cases, I guess? Maybe something like "Wallet mit beschränktem Lesezugriff erstellen" or "Wallet nur mit Leseberechtigung erstellen" is clearer?

JamesCullum: "beschränkter Lesezugriff" sounds like view-only is even more limited. "Wallet nur mit Leseberechtigung erstellen" is the best one I think.

translations/monero-core_de.ts
JamesCullum: I think its supposed to be an option for later, customizable settings.

translations/monero-core_de.ts
Jaqueeee: No. not needed.

translations/monero-core_de.ts
Jaqueeee: There's a chance Daemon is going to be renamed to Node in english version. So maybe you could use the german word for Node if it suits better.

louvetic: That is "Knoten" and would avoid the umlaut-entities, nice. But we cannot jump in boots yet and use it already?

translations/monero-core_de.ts
JamesCullum: Alright, then this should be good to go.

JamesCullum:
Looks like we need to do the opposite - remove all html entitites. Build fails with `Parse error at ../translations/monero-core_de.ts:143:34: Entity 'szlig' not declared.`

translations/monero-core_de.ts
louvetic: Ist angepasst.

translations/monero-core_de.ts
```diff
@@ -56,27 +56,27 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="139"/>
         <source>Give this entry a name or description</source>
-        <translation type="unfinished"></translation>
+        <translation>Einen Namen oder eine Beschreibung hinzufügen</translation>
```
louvetic: Ist angepasst.

Jaqueeee:
That's right @JamesCullum. HtmlEntities are not allowed in the translation files.

translations/monero-core_de.ts
louvetic: So we leave it at which translation?

translations/monero-core_de.ts
```diff
@@ -140,19 +140,19 @@
     <message>
         <location filename="../components/DaemonConsole.qml" line="108"/>
         <source>Close</source>
-        <translation type="unfinished"></translation>
+        <translation>Schließen</translation>
```
louvetic: @Jaqueeee said no. I cannot tell.

translations/monero-core_de.ts
```diff
@@ -140,19 +140,19 @@
     <message>
         <location filename="../components/DaemonConsole.qml" line="108"/>
         <source>Close</source>
-        <translation type="unfinished"></translation>
+        <translation>Schlie&szlig;en</translation>
```
Jaqueeee: as @JamesCullum pointed out, build fails on this line because of the htmlencoding.

louvetic: Changed back to normal ß.

NanoAkron:
Please squash commits for sake of future reviewers

louvetic:
@NanoAkron Squashing commits like mentioned on https://github.com/blog/2141-squash-your-commits is enabled, but I just can't find a button somewhere here to actually do it. If it has to be done via command line, please do it for me, as it would exceed my modest skills.

NanoAkron:
Ah ok. Unfortunately I don't know the desktop version very well, and I can't squash your commits for you :(

Try 'git rebase -i' on the command line

Jaqueeee:
Maybe @fluffypony can help you squashing all these commits before merging?

translations/monero-core_de.ts
M5M400: I find Dämon and Knoten both very unaestethic, to be honest. How about going with something more familiar to Windows users (after all, they will probably the majority of users?) like "Der Monero Dienst scheint nicht aktiv zu sein" or "Das Monero Dienstprogramm scheint nicht aktiv zu sein"?

amiuhle:
I squashed the commits here: https://github.com/amiuhle/monero-core/tree/i18n-de

@louvetic If you give me write access to your repository I can force push this to your branch and then I think we're good to go!

louvetic:
@amihle Were you successful? 

Should we first focus on the **Dämon & Knoten** discussion and related aesthetics before we continue translating more?

amiuhle:
@louvetic Yes, I squashed everything down to one commit, which should contain all the changes you've previously made in several commits.

translations/monero-core_de.ts
amiuhle: https://de.wikipedia.org/wiki/Daemon

I don't have any strong feelings on Daemon vs Dämon. I don't like the idea to use node in the English translation, but that's another issue. However, I think this shouldn't be part of this PR. If Daemon does get renamed to Node, then the renaming of all the translations should be part of the Pull Request that handles this (or a follow up to that PR).

JamesCullum: I agree that the Node renaming should be handled seperately. I am not a big fan of the rather literal translation "Dämon", so I would prefer "Daemon". Do we have any opposing strong opinion? Otherwise, if nobody but me really cares we use Daemon, do we? ;)

amiuhle: Fine with me.

If there are no objections, I'll replace "Dämon" with "Daemon" tonight, then squash again so this can get merged.

Cc @louvetic

louvetic: Neither "Daemon" nor "Dämon" is speaking for itself, if you don't have a tech background. But Monero is supposed to be widely acceptable and usable. Therefore we should focus on an understandable language. So I prefer a more neutral term like "Service" or "Dienst".

translations/monero-core_de.ts
JamesCullum: At the moment it doesn't really matter, so we gotta see how they plan to use it before we can really say anything. Guess we leave it like this then.

translations/monero-core_de.ts
JamesCullum: If you don't have a tech background at all, the english version or quite a lot of things don't make a lot of sense (for example mining). I don't think using a more simplified term helps them a lot, while the more common amount of people would be confused of such a big change. At the moment there is not a lot of information available in German, so people trying to find out about parts of the software will be unable to find anything when googling for "service", as the translation is not obvious. Even if they stumble over someone talking about nodes or daemons in english, he probably won't even know thats the service he is looking for.

So basically:

- Non-Techie people make up a very low amount of actual users and such a slight change won't do much

- Techie people don't care that much but will have a way easier time finding more information about it when we use "Daemon".

louvetic: The fact, that maybe only few non-techies use Monero _(what is not proven - but if it was, maybe it's because using Monero is too technically challenging?)_, is not a valid argument to leave it at the techie-style language. Actually, we could leave everything in English as a consequence, why bother with a translation at all? Tell me, why do they build barrier-free entries to buildings, although no wheelchair driver ever used the steps before? True, by that I exaggerate pointedly... ;)

I personally know of a flight attendant and she bought some Monero last week after she got it all explained from her boyfriend. Of course, she's clever _(come on, she owns some Monero, that alone makes you a lot smarter!)_ and wouldn't suspect an evil spirit in her laptop, when working with the GUI. But with something like "Service" or "Dienst" she would at least have expectations what the program/function does, without having to google it first. It's more respectful not to expect every user to be not less than an advanced amateur.

The French use "démon" and not an English word, what's maybe not too surprising for them. Whereas the Spanish use "daemon" and not their native-language "demonio". 

**So whatever we use, we're among good friends! :)**

JamesCullum: So all languages translate it literally to the local version of "daemon"? Isn't that another argument for using that one? ;) 

However, lets be progressive in this possible deadlock - do we want to cast a vote, or do you have another good idea for settling this point? :)

louvetic:
Thanks a lot @amiuhle @JamesCullum @M5M400 @Jaqueeee - should we now check on lines 1043 and following? Who starts a branch, so I can contribute somewhere?

PR #654, opened 2017-03-31
Rename title of Readme.md to Monero Core (GUI) - No merge yet

xmr-eric:
Right now, we have a disconnection that is happening. The repo is called monero-core while the title on the readme calls it Monero GUI. I think originally it was also named Monero Core, but was changed to Monero GUI because it was confusing people looking for the GUI.

This PR proposes to rename the Readme back to Monero Core, but to include (GUI) in parentheses. It's still a good idea to keep calling the repo Core, since this will be the core software that the majority of users will use. And it's good to bring the Readme title back in line with the repo name, ie Monero Core. But including GUI in parenthesis will help people looking for the GUI who were previously confused, which solves the reason for the original change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Monero is a private, secure, untraceable, decentralised digital currency. You ar
 
 ## About this Project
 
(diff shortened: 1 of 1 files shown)


xmr-eric:
Superceded by #674

PR #670, opened 2017-04-11
translate to chinese zh-CN

zhizhongzhiwai:
I am a developer from China

This submission fixes some of the wrong translations, while making the translation more in line with the Chinese mainland's reading habits

PR #671, opened 2017-04-11
Fix words.

p1gd0g:
"回复" should be "恢复".


NanoAkron:
Could you please recommend this change to #670 and discuss it there, rather than adding it at this stage before that large change has been merged?

zhizhongzhiwai:
#670  has been fix the wrong translation.

PR #676, opened 2017-04-14
Swedish translation

snaggen:
Here is a full Swedish translation, I have also added a swedish flag and added it to the Setup guide. I don't know if there were some particular order, so I just added it in the middle after France since that is a near by country. I have also run "lupdate", so it should be complete with the latest strings.


Jaqueeee:
Looks good, but please squash all those commits into one.

snaggen:
Ok, I have squashed my commits to one single commit.

PR #678, opened 2017-04-14
Simplify fee structure (divide all by 4)

xmr-eric:
Right now the fees are structured Low (x1 fee), Default (x4 fee), Medium (x20 fee), and High (x166 fee). But this makes it look like the default fee is paying four times the regular fee (or, default is paying 4x the default), which is confusing.

This PR suggest lowering all fee multipliers by a factor of four. So Default is now just x1 fee. Low is x0.25 fee, Medium is x5 fee, and High is x41.5.

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -199,10 +199,10 @@ Rectangle {
       ListModel {
            id: priorityModelV5
 
-           ListElement { column1: qsTr("Low (x0.25 fee)") ; column2: ""; priority: 1}
+           ListElement { column1: qsTr("Slow (x0.25 fee)") ; column2: ""; priority: 1}
            ListElement { column1: qsTr("Default (x1 fee)") ; column2: ""; priority: 2 }
-           ListElement { column1: qsTr("Medium (x5 fee)") ; column2: ""; priority: 3 }
-           ListElement { column1: qsTr("High (x41.5 fee)")  ; column2: "";  priority: 4 }
+           ListElement { column1: qsTr("Fast (x5 fee)") ; column2: ""; priority: 3 }
+           ListElement { column1: qsTr("Fastest (x41.5 fee)")  ; column2: "";  priority: 4 }
 
        }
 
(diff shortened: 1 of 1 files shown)


NanoAkron:
Please squash then we can discuss :)

NanoAkron:
So I have to ask, does the maths still work i.e. for paying the appropriate penalties to encourage the miners to push the block size up for that one extra high priority transaction? 

ping @JollyMort

JollyMort:
That part stays the same. This is just changing the labels of each setting. Min. fee will still be "neutral" to a 1.2% block size increase and the default will be neutral to a 4.8% increase, which is just right to squeeze 1 TX of max. 14.4kB above the median. The "sweet spot" for the miner is at 7.2kB where he'd maximize his profit. Anything between 14.4kB and 0kB increase will earn him more compared to exactly the median-sized block.

bigreddmachine:
@NanoAkron et al. - This seems like a pretty straightforward change and one that makes sense from a UX standpoint. I don't think it actually changes anything, just changes how the user views what they are doing. That being said, I still don't think this is the ideal way to do fees at all, but I haven't had a chance to fully build out a model to evaluate that due to work obligations. I hope to get back to that in May and write up something end of May/early June.

iamsmooth:
I agree with the premise here but maybe "Low Medium High" are poor labels given the expected use case where Low incurs a risk of long delays and possibly being dropped altogether. L/M/H suggests the three are all much the same other than amount, but Low is really intended as a different sort of request.

dternyak:
Maybe change them to Slow, Average, Fast while we're here?

bigreddmachine:
I've seen Bitcoin wallets use terms like "Priority", "Normal", "Economic", etc. Maybe something like that works here?

xmr-eric:
Left default as default, since it means people don't have to think, which is the goal of good design. But switched "Low" to "Slow, "Medium" to "Fast", and "High" to Fastest.

fluffypony:
Reviewed

PR #685, opened 2017-04-18
Address Book context menu : Send to this address

jonathancross:
This is a small improvement to the context menu which is shown for each entry in the `AddressBookTable`.  Currently, there is an option "**Send to same destination**" which is unclear as it suggests some vague notion that the user has already sent XMR to this destination and would like to do so again.

This PR replaces the above message with "**Send to this address**", as the currently selected address will be inserted into the Send view as the recipient address.

Screenshot:
monero-send-to-this-address

I found several other instances of this string and updated all for consistency.

Changes / suggestions welcome.

### DIFF
--- a/components/HistoryTable.qml
+++ b/components/HistoryTable.qml
@@ -443,7 +443,7 @@ ListView {
         id: dropModel
         ListElement { name: "<b>Copy address to clipboard</b>"; icon: "../images/dropdownCopy.png" }
         ListElement { name: "<b>Add to address book</b>"; icon: "../images/dropdownAdd.png" }
-        ListElement { name: "<b>Send to same destination</b>"; icon: "../images/dropdownSend.png" }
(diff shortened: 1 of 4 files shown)


components/HistoryTable.qml
```diff
@@ -443,7 +443,7 @@ ListView {
         id: dropModel
         ListElement { name: "<b>Copy address to clipboard</b>"; icon: "../images/dropdownCopy.png" }
         ListElement { name: "<b>Add to address book</b>"; icon: "../images/dropdownAdd.png" }
-        ListElement { name: "<b>Send to same destination</b>"; icon: "../images/dropdownSend.png" }
+        ListElement { name: "<b>Send to this address</b>"; icon: "../images/dropdownSend.png" }
```
jonathancross: FYI: This is currently unused as the code above for `TableDropdown` is commented out.

PR #688, opened 2017-04-23
Add tip text to labels

### LINKED ISSUE #684: [FEATURE REQUEST] User Help via Tiptext
Every application needs some help for new users. The idea isnt really to explain to the user how cryptocurrencies work in general, but to shed some additional light in some of the most confusing parts of Monero. This is probabaly the PaymentId, Integrated Address, Unlock Time, Smart Mining and so forth. 

personally i am in favour of using classical Tiptext to display this additional Information. not only its convenient but the implementation of tiptext could help also solve other issues without overloading the layout (for example https://github.com/monero-project/monero-core/issues/142)

This issue should be the placeholder for all this work and related discussion. 

some points worth talking about (and there are sure more) :

- should it be possible to activate/deavtivate the help ?
- is tiptext the right choice to do this ?

+feature

Jaqueeee:
Fixes #684 

Adds tip text to all labels (se screenshot).
 
Known issues: 
- The [?] button is only visible in custom title bar. When native bar is used we need to move the button somewhere else. 
- I've added placeholder text to all labels now to make testing easier. They have to be replaced or removed before next release. 
- When final help texts are added we probably need to adjust the positioning for some of the labels. 

screenshot

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -128,7 +128,7 @@ Rectangle {
             text: qsTr("Balance") + translationManager.emptyString
             anchors.left: parent.left
             anchors.leftMargin: 50
-            tipText: qsTr("Test tip 1<br/><br/>line 2") + translationManager.emptyString
+            tipText: qsTr("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.") + translationManager.emptyString
         }
 
         Row {
@@ -172,7 +172,7 @@ Rectangle {
             text: qsTr("Unlocked balance") + translationManager.emptyString
             anchors.left: parent.left
             anchors.leftMargin: 50
-            tipText: qsTr("Test tip 2<br/><br/>line 2") + translationManager.emptyString
+            tipText: qsTr("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. ") + translationManager.emptyString
         }
 
         Text {
--- a/components/Label.qml
+++ b/components/Label.qml
@@ -56,26 +56,27 @@ Item {
         anchors.left: label.right
         anchors.leftMargin: 5
         source: "../images/whatIsIcon.png"
-        visible: appWindow.whatIsEnable
+        visible: appWindow.whatIsEnable && item.tipText != ""
     }
 
     MouseArea {
         anchors.fill: icon
-        enabled: appWindow.whatIsEnable
+        enabled: appWindow.whatIsEnable && icon.visible
         hoverEnabled: true
         onEntered: {
-            icon.visible = false
-            var pos = appWindow.mapFromItem(icon, 0, -15)
-            tipItem.text = item.tipText
-            tipItem.x = pos.x
-            if(tipItem.height > 30)
-                pos.y -= tipItem.height - 28
-            tipItem.y = pos.y
-            tipItem.visible = true
+            var pos = rootItem.mapFromItem(icon, 20, -35)
+            appWindow.toolTip.text = item.tipText
+            appWindow.toolTip.x = pos.x - appWindow.toolTip.width/2
+            appWindow.toolTip.tipX = appWindow.toolTip.width/2 - 10;
+            if(appWindow.toolTip.height > 30)
+                pos.y -= appWindow.toolTip.height - 30
+            appWindow.toolTip.y = pos.y
+            appWindow.toolTip.visible = true
+            appWindow.toolTip.z = 3
+
         }
         onExited: {
-            icon.visible = Qt.binding(function(){ return appWindow.whatIsEnable; })
-            tipItem.visible = false
+            appWindow.toolTip.visible = false
         }
     }
 }
--- a/components/TableDropdown.qml
+++ b/components/TableDropdown.qml
@@ -220,8 +220,7 @@ Item {
                                 var pos = rootItem.mapFromItem(delegate, 30, -25)
                                 appWindow.toolTip.text = qsTr(name) + translationManager.emptyString
                                 appWindow.toolTip.x = pos.x - appWindow.toolTip.width
-//                                if(appWindow.toolTip.height > 30)
-//                                    pos.y -= appWindow.toolTip.height - 30
+                                appWindow.toolTip.tipX = appWindow.toolTip.width - 10
                                 appWindow.toolTip.y = pos.y
                                 appWindow.toolTip.visible = true
                                 appWindow.toolTip.z = 3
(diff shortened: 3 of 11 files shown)


xmr-eric:
The white on orange text is not ideal. Little contrast results in poor legibility. Is it possible to get simple white on black?

jonathancross:
Adding help text in the UI like this is very important, thanks!
However, I find the specific implementation using icons to be a bit noisy (ie distracting in the UI, the orange color makes it even stronger).

I am wondering if we can instead adding a tooltip indicator for each text label (normally a dashed underline).  Then we can have the tooltip bubble appear over the label text on hover?

This would also allow us to retire the jarring modal "help" text already in use on the **Received** tab, etc.

(I'll mock up the idea and add here)

Jaqueeee:
@xmr-eric yes, i'll try with a different bg color. 

@jonathancross Good idea! The icons are a bit noisy indeed, note that they're only visible when the question mark button in title bar is clicked. If we implement your suggestion we can skip that button which also solves point 1 above.

jonathancross:
@Jaqueeee Great!  I'd be happy to help test, add tooltip text (or anything else) next week.

fluffypony:
Please rebase

jonathancross:
Hi @Jaqueeee, have you done any more work on this?
Anything I can help out with?

Jaqueeee:
@jonathancross no i haven't. This PR was discussed on a dev meeting a couple of weeks back. @fluffypony suggested we use an overlay instead of the tooltips. Something like this: http://www.inspireux.com/wp-content/uploads/IMG_0087.jpg

I've been busy with other stuff lately, so i haven't started working on it yet. Do you have any thoughts on this? I'm thinking it would be good if someone with UX experience could have a look at it before we move on.

jonathancross:
> @fluffypony suggested we use an overlay instead of the tooltips.

Although I think an overlay would be okay as an introduction to the UI, I think it is not suitable for explaining individual items, expected values, impact of choices, etc.  The details are best left as in-context tooltips IMO.  From my reading of the [dev meeting notes](https://getmonero.org/2017/05/07/overview-and-logs-for-the-dev-meeting-held-on-2017-05-07.html), it seems like he was suggesting the "overlay" as part of a seperate "unified" help system:

> **\<fluffypony\>** tooltips are fine, but if we're going to do some sort of unified help then I would veer towards an overlay that shows once the first time you enter a screen, and can be re-called by clicking the [?] button on the taskbar

Seems there is two different help systems being discussed here:

1. Individual, in-context "tooltip" bubbles explaining one particular field or control (linked text which would show the bubble on mouseover and onclick).  We need the click / tap for mobile.
2. An "[Instructional Overlay](https://www.nngroup.com/articles/mobile-instructional-overlay/)" / intro (fluffypony suggests a full-screen modal with notes on how to use the controls, general sections of the UI, etc). This should only be displayed for new users to introduce each screen.  For those familiar with the GUI, this will get annoying fast, so I'd suggest this be easy to dismiss.  The user would have a "help" button that could be used to recall the overlay later.

I'd suggest we separate these into two PRs / issues as they address different needs.

Thoughts?

fluffypony:
@jonathancross I was suggesting only the overlay, and that the overlay can be recalled / shown at any point by clicking the help. Tooltips aren't terrible, but they're a little dated and can get in the way.

jonathancross:
Thanks for the clarification @fluffypony.

I'm coming from the perspective that the primary problem being solved is:  The Monero wallet requires specialized knowledge and can be relatively daunting for new users.  We want to assist new users in getting started, but also in helping more experienced users confidently work with the wallet.

Do others agree with this definition of the challenge and the goals?

If so, read on...

I thought it might be helpful to explore some UI patterns to help us to determine the best approach.

For me, the "tooltips" represent an [Inline Help Box](http://ui-patterns.com/patterns/InlineHelpBox). Usage:
* Use when an interaction with your website is not necessarily intuitive and self-explanatory in itself.
* Use inline tips to document your interface
* Use when you want to motivate the user to start using the system.
* Use when you want to aid the user in getting started with your web application.
* Use when you want to guide the user to get a good start with your web application.
* Use to gently introduce functionality to the new and untrained user.
* Use the ‘hide’ functionality to avoid frustration of the experienced and trained user.

The "overlay" as suggested seems to be closer to an [Onboarding UI pattern](http://ui-patterns.com/patterns/onboarding/list), something like an non-interactive [Guided Tour](http://ui-patterns.com/patterns/Guided-tour) and a visual Help system combined.

Personally I find there is a steep learning curve with Monero and would like detailed info / explanation of each item in the UI.  The overlay is a variation of of a [Modal Window](http://ui-patterns.com/patterns/modal-windows), it blocks the user from doing anything until dismissed. Once the overlay is dismissed, the user is back where they started and has no explanation of what each item is (until they switch back to the "help" overlay).  This causes an abrupt break in the flow and forces the user to deal with an entirely new screen each time (semi-transparent background is only a minor help in maintaining context).  I'm also concerned that there is not much space for details (user will not remember details once the modal is dismissed anyway).

I have some hesitation towards the "help" overlay just because it is unusual in desktop apps (I can't find an example in any of the apps I use on the Mac).

The benefit of the tooltip is that users can seek out detailed info _as needed_ and in-context, non-linearly while in the process of filling out the required fields without distracting them from the task they are trying to accomplish. 

> Tooltips aren't terrible, but they're a little dated and can get in the way.

I don't feel they are necessarily dated (_or maybe I'm just dated?_ :-), but see this pattern as an established tool to provide inline help in an unfamiliar UI.  Tooltips are fairly ubiquitous in desktop applications.  I agree that if not implemented carefully, they can get in the way. For mobile, they would require tapping the text label.  For desktop, we could show on-click or after x number of seconds on hover over the text.

Seems like neither UI pattern will properly all needs alone (_**detailed explanation of each form element and significance**_ vs **_introduction to the application, each screen and how the UI works_**).

I'm new here, so don't want to step on any toes, but hope this helps the conversation  :-)

We can of course implement the overlay, try it out and see if it meets user's needs, but I don't see it helping much with the challenges I personally faced as a new user.

Jaqueeee:
I agree with @jonathancross on this. Tooltips and overlays fulfil different goals imo. 

I suggest we merge this PR (after i rebased and implemented the dashed underline) so contributors could start adding and translating help texts. We could add overlay later if that's what we want. 

btw, I found this article while researching overlay implementations. 
https://uxplanet.org/mobile-onboarding-interact-don-t-tell-f0c35da2b2b4

jonathancross:
That is a fantastic article @Jaqueeee, thanks for sharing!  For mobile users, we might consider a _hybrid_ approach where the item being interacted with (eg form input) is isolated visually from other elements with "tooltip" helper text being displayed.
2017-06-06_tooltipmobile

fluffypony:
FYI: this PR has merge conflicts, not sure if this is WIP or not:)

PR #704, opened 2017-05-02
fix version info

### LINKED ISSUE #701: [Feature Request] GUI needs to display its version prominently
Will seriously help debugging with noobs and general troubleshooting.

Jaqueeee:
Fixes #701 

Outputs tag + commit hash on settings page:

screenshot

### DIFF
--- a/utils.sh
+++ b/utils.sh
@@ -53,6 +53,12 @@ function get_tag()
         echo "You are ahead of or behind a tagged release"
         VERSIONTAG="$COMMIT"
       fi
+      # save tag name + commit if availible
+      TAGNAME=$(git describe | sed -e 's/[\t ]*//')
+      if test -z "$TAGNAME"
+      then
+        TAGNAME="$VERSIONTAG"
+      fi
     fi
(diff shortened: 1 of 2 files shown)


utils.sh
```diff
@@ -53,6 +53,12 @@ function get_tag()
         echo "You are ahead of or behind a tagged release"
         VERSIONTAG="$COMMIT"
       fi
+      # save tag name + commit if availible
+      TAGNAME=$(git describe | sed -e 's/[\t ]*//')
```
jonathancross: Should double quote chars (`"`) also be stripped out?
I took a look at [what is allowed in git tags](https://git-scm.com/docs/git-check-ref-format) and did not see anything preventing quotes. Even if it is a very unlikely situation, it could cause problems when the `version.js` is created above.

Example:
```
% TAGNAME=$(echo 'foo"bar')
% echo "var GUI_MONERO_VERSION = \"$TAGNAME\""
var GUI_MONERO_VERSION = "foo"bar"
```

Issue #448, opened 2017-02-02
set the blockchain dir during startup and a convenient way after as well.


madLyfe:
would be nice to have the ability at install for the gui to ask where to install the large blockchain rather than a default dir. dealing with flags is not very user friendly. this would also need to be thought of for when someone wants to move it at a later date as well w/o having to use flags. maybe a blockchain dir field in the main settings with a button to 'change dir'(it was already set at install, in the case of the feature above) and it would move it for you wherever you like with a progress bar since its 10gb.

i am new to all of this, and i take it the gui is supposed to be entry level friendly. these would be good improvements.

Jaqueeee:
Added blockchain folder selection to settings page in #709. Not added to wizard yet.

dEBRUYNE-1:
+resolved

PR #709, opened 2017-05-04
Blockchain folder settings (data-dir)

Jaqueeee:
Fixes parts of #448

### DIFF
--- a/pages/Settings.qml
+++ b/pages/Settings.qml
@@ -224,11 +224,38 @@ Rectangle {
                 }
             }
 
-
-
         }
 
-        ColumnLayout {
+        RowLayout {
+            id: blockchainFolderRow
+            Label {
+                id: blockchainFolderLabel
+                color: "#4A4949"
+                text: qsTr("Blockchain location") + translationManager.emptyString
+                fontSize: 16
+            }
+            LineEdit {
+                id: blockchainFolder
+                Layout.preferredWidth:  200
+                Layout.fillWidth: true
+                text: persistentSettings.blockchainDataDir
+                placeholderText: qsTr("(optional)") + translationManager.emptyString
+
+                MouseArea {
+                    anchors.fill: parent
+                    onClicked: {
+                        mouse.accepted = false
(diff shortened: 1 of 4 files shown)


fluffypony:
Please rebase and check in-line comments

pages/Settings.qml
```diff
@@ -437,6 +465,58 @@ Rectangle {
 
     }
 
+    // Choose blockchain folder
+    FileDialog {
+        id: blockchainFileDialog
+        title: "Please choose a folder"
```
Jaqueeee: yes! fixed

pages/Settings.qml
```diff
@@ -437,6 +465,58 @@ Rectangle {
...
+        selectFolder: true
+        folder: "file://" + persistentSettings.blockchainDataDir
+
+        onAccepted: {
+            var dataDir = walletManager.urlToLocalPath(blockchainFileDialog.fileUrl)
+            var validator = daemonManager.validateDataDir(dataDir);
+            if(!validator.valid) {
+
+                confirmationDialog.title = qsTr("Warning") + translationManager.emptyString;
```
Jaqueeee: I try to avoid them in the qml files because that seemed to be the most common case when i started contributing. But since they are required in the c++ files it's easy to forget.

pages/Settings.qml
```diff
@@ -437,6 +465,58 @@ Rectangle {
...
+                confirmationDialog.text = "";
+                if(validator.readOnly)
+                    confirmationDialog.text  += qsTr("Error: Filesystem is read only") + "\n\n"
```
Jaqueeee: good point. Added curly brackets. Altough, this page states they shouldn't be used for one-liners. https://wiki.qt.io/Qt_Coding_Style

Jaqueeee:
@fluffypony: rebased and fixed commented issues. 

@jonathancross: Don't think we have a styleguide. Fluffy might know?

fluffypony:
There's no style guide in Monero, but there is one in Kovri that us reasonable.

PR #711, opened 2017-05-04
remote nodes (again) + ux + wizard daemon settings

Jaqueeee:
- Adds daemon settings in wizard
- Adds option to be connected to remote node while local node is syncing. 
- ux improvements (advanced daemon settings etc)

Related to #605, #602, #572

Screenshot

screenshot2


Gingeropolous:
You should add a disclaimer underneath the "connect to 3rd party node" field that communicates the loss of privacy and trust involved when using a remote node, and encourage the user just to be patient.

SamsungGalaxyPlayer:
What about users who want to connect to a remote node and do not want to have a full node on their computer?

I assume they should untick the first box and keep the second ticked with the remote node, but I think the new wording here can lead to potential confusion for users only using a remote node. It assumes all users who use a remote node will also have a full node running locally.

Perhaps the top box can be selected by default, with the bottom one unchecked. When the top box is deselected, maybe the wording can shorted to just "connect to a 3rd party remote node ".

Jaqueeee:
@SamsungGalaxyPlayer that's actually exactly how it's implemented currently. =)
First option is checked by default. If it unchecks the wording on the second one change to "connect to a 3rd party remote node ."

I was disussing with @fluffypony yesterday about adding lightwallet support to official GUI. I.e option connect wallet wallet to a MyMonero or OpenMonero node that performs the blockchain scanning with view key. If we go down that road, there will be 3 options on this page instead. It will be further discussed on next dev meeting.

fluffypony:
Please rebase

PR #729, opened 2017-05-15
Use "Transaction speed" instead of "Transaction priority"

xmr-eric:
Easier for laypeople to understand, and corresponds with the new Slow/Default/Fast/Fastest terms in #678


iamsmooth:
Don't agree with this as we can't and don't promise any particular speed. If there is a high volume of transactions at the same or higher priority then there will be delays. (Also variation in block time will also unpredictably effect speed.)

I think priority better describes what is actually going on.

xmr-eric:
@iamsmooth You're probably right. Unless others feel differently I'll close this one.

PR #733, opened 2017-05-16
Restore wallet: Adding space after 'Your wallet is stored in'

jonathancross:
Just adding space and a colon between the sentence "Your wallet is stored in" and the path that follows.  I added the space _outside_ of the translated string so that existing translations would not break.

Here is a screenshot demonstrating the issue:
2017-05-17_missing-space-restore-wallet

### DIFF
--- a/wizard/WizardManageWalletUI.qml
+++ b/wizard/WizardManageWalletUI.qml
@@ -242,7 +242,7 @@ ColumnLayout {
             Layout.fillWidth: true
             Layout.topMargin: 20
             fontSize: 14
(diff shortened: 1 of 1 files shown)


NanoAkron:
Strange it's pulled in all those whitespace changes in the comments

jonathancross:
My editor removes trailing whitespace by default. Happy to revert those if that is preferred?

fluffypony:
@jonathancross better to undo those and squash, otherwise the PR looks bigger than it is

jonathancross:
@fluffypony Thanks for the feedback, all cleaned up now.

PR #737, opened 2017-05-23
option to view seed on wallets recovered from keys

### LINKED ISSUE #728: [Feature request] Seed for CLI wallets generated from keys / Error for MyMonero seed
Unless I'm mistaken, the GUI shows a blank seed for wallets generated from keys. However, the CLI will display a seed for wallets generated from keys if the spend/view keys are from the CLI, but *not* from MyMonero. See this discussion: https://www.reddit.com/r/Monero/comments/6azzii/why_dont_wallets_restored_with_viewspend_keys/

So theoretically we could display a seed for wallets restored from keys if the keys are from the CLI.

Similarly, we should show an error specific to MyMonero. The seed box could show something like "Restored wallet version does not allow backup seeds. If you wish to have a backup seed then create a fresh wallet."

Jaqueeee:
Fixes #728

### DIFF
--- a/pages/Settings.qml
+++ b/pages/Settings.qml
@@ -467,6 +467,11 @@ Rectangle {
 
         onAccepted: {
             if(appWindow.password === settingsPasswordDialog.password){
+                if(currentWallet.seedLanguage == "") {
+                    console.log("No seed language set. Using English as default");
+                    currentWallet.setSeedLanguage("English");
+                }
+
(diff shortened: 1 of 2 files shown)


Jaqueeee:
Added fallback to english for wallets without seed language set. New wallets restored from keys will have their seed language set to the selected language in wizard.

PR #762, opened 2017-06-11
Monero Core + GUI Wallet Beta 2 Windows installer, initial version

rbrunner7:
This is the initial version of a Inno Setup based Windows installer for the GUI Wallet Beta 2.

Instructions of how to build the installer .exe are in README.md.  (The file ReadMe.htm is the setup's ReadMe for the people that install.)

As I am new to git, GitHub and Open Source in general, there may well be some issues with this very first version. And of course this installer is in need of some beta testers that have a good look at it.

 Eagerly awaiting your comments, corrections and feedback!


rbrunner7:
Thanks, danmiller. Will you also kind of beta-test the resulting installer first and have a look at the naming of the icons etc.? That would be reassuring. As I said I am new to this stuff.

I could only find the license of Inno Setup itself, wich is [here](http://www.jrsoftware.org/files/is/license.txt). But as it says *Permission is granted to anyone to use this software for any purpose,* I assume the resulting installers are unencumbered as well. Anyway, Inno Setup looks true Open Source.

I include a near-empty *monero-wallet-gui.log* so that I can copy it to the app directory and give it writing rights; the GUI Wallet could not create it itself in a directory like *C:\Program Files\Monero Core*. Details about this little "trick" are in comments in the .iss file.

rbrunner7:
After a discussion with some dev's on #monero-dev last night I dropped the controversial and mistakable "core" from the installer in all places: Instead of *Monero Core + GUI Wallet* it's now simply *Monero GUI Wallet*.

So the pull request is now 3 commits instead of 2. (I hope I did the right thing, i.e. nothing, just pushing the changes to GitHub, but not making a new pull request and neither deleting this one.)

medusadigital:
hei @rbrunner7 , we discussed your work again yesterday and think it would be wise if you could add:

- an uninstaller
- put everything except the application launcher and the uninstaller into a sub folder in the start menu, named "Utilities"


what do you think ?


sorry for the slow feedback

rbrunner7:
medusadigital, no problem about the feedback. Things usually get better anyway if they have some time to "ripen".

I had the impression that separate *Uninstall* icons were out of style and rare nowadays because you can always uinstall in the control panel anyway, but a quick look through the program groups on my PC showed me that this is not the case: More than half of the applications do include such an "icon". I will add one.

I first had a sub-group for all the non-essential icons (I had called it *Troubleshooting*) but gave that up when I saw that while Windows 7 displays such a sub-group just fine, Windows 10 does not display it at all. But it's no problem to get it back. Still, I think I will keep the "strange" icon titles nevertheless, exactly for Windows 10.

I also noticed in the meantime that my installer misses a page that other installers offer, a page where you can wish for a desktop icon and a "quickstart" icon. I will check how difficult it is to add that as well.

rbrunner7:
Done.

Did already somebody have a good look at "ReadMe.htm" that I wrote as part of this installer? I know, many people probably won't read it anyway, but if they do, they should not find mistakes in there, or things that are contrary to the "Monero philosophy".

PR #763, opened 2017-06-13
Checkout submodules

athanclark:
the `monero` project itself needs to be available under `./`, and I noticed that it was a submodule, so this step was just left out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ Packaging for your favorite distribution would be a welcome contribution!
       `emerge dev-qt/qtmultimedia:5 media-gfx/zbar`
 
 5. Build the GUI
+  - Checkout submodules
+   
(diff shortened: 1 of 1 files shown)


Jaqueeee:
It's being handled by the build scripts. 
https://github.com/monero-project/monero-core/blob/master/get_libwallet_api.sh#L15-L18

jonathancross:
Hi @athanclark,
Just to be clear, as @Jaqueeee mentioned, get_libwallet_api.sh handles the checkout of the submodule and that script is [called automatically from build.sh](https://github.com/monero-project/monero-core/blob/master/build.sh#L57)

Do you agree this be closed?

athanclark:
@jonathancross Yes I do, apologies!

PR #769, opened 2017-06-18
Fix font display on Tails

### LINKED ISSUE #768: Running monero gui 2 on tails does not display text
On running monero gui 2 on tails the text is not displayed. To correct this I included this line in the start-gui.sh file.

```bash
export QT_QPA_FONTDIR=/usr/share/fonts/truetype/unifont/
```

Perhaps a specific font file can be included in the distribution files.

Another issue is that the gui screens are not scrollable. On my tails os screen I was not able to access the advanced option buttons on the send transaction screen.

jonathancross:
Fixes #768


jonathancross:
Sounds like this was caused by missing `fontconfig` on the build system, so we should probably just prevent that from happening again instead of this change.

PR #787, opened 2017-07-02
Update monero-core_eo.ts

Keksoj:
Hi there,
The previous translator did a tremendous job. Yet there were a lot of small erors in style and (sometimes) vocabular. I guess there are still some long and complicated entries (long CSS-like line) that aren't ready yet. 
One thing : what's the difference between an "account" and a "wallet", because the previous translator used the same word "monujo" for both (that means "wallet"). He didn't use the esperanto word "konto", which means "account".
Still some tiny mistakes around, but I guess we'll work on that.


Keksoj:
Wait, still some bloody mistakes I let go through, and some code needs revision. Plus new parts of the translation needs doing (all the <translation type="unfinished"></translation> lines). I get to it.

PR #791, opened 2017-07-09
Add window title

### LINKED ISSUE #747: GUI Beta 2, Windows 8.1 an 10, Program name missing in taskbar
GUI software name is not shown in the Windows taskbar, when taskbar grouping is disabled. Tested it on Windows 8.1 and 10, 64 Bit.

gui_taskbar

MaxXor:
Fixes #747

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -42,7 +42,7 @@ import "wizard"
 
 ApplicationWindow {
     id: appWindow
-    title: "Monero GUI"
+    title: "Monero"
 
     property var currentItem
     property bool whatIsEnable: false
--- a/main.qml
+++ b/main.qml
@@ -42,7 +42,7 @@ import "wizard"
 
 ApplicationWindow {
     id: appWindow
-
+    title: "Monero GUI"
 
     property var currentItem
     property bool whatIsEnable: false


main.qml
```diff
@@ -42,7 +42,7 @@ import "wizard"
 
 ApplicationWindow {
     id: appWindow
-
+    title: "Monero GUI"
```
snaggen: Not saying this is a bad title, but I wonder if "Monero Wallet" isn't a better title? The GUI part feels redundant since in a desktop environment everyting that is shown is GUI:s. And it is a wallet (even if it has some daemon controlling capabilities, that is just some bonus features and not the primary function), hence my suggestion "Monero Wallet".

MaxXor: I used the title of the [README.md ](https://github.com/monero-project/monero-core/blob/master/README.md) file. Well, the title of the Monero wallet just says "Monero" when it's running. So we have basically three options: Monero, Monero Wallet or Monero GUI. Maybe @fluffypony or @Jaqueeee can help us with the decision?

xmr-eric: What it should be is the actual name of the piece of software, like Monero Prime or Monero Kerna. But the community hasn't yet come to a conclusion about what the name for the GUI should be. See discussion here: https://github.com/monero-project/monero-core/issues/674

jonathancross: Maybe best to just use "Monero" for now until we have clear direction on other naming efforts?

xmr-eric: Probably best 👍

jonathancross:
Thanks @MaxXor, have you confirmed this fixes the Windows taskbar issue in #747 ?
Any effect on other platforms?

MaxXor:
Window title is correctly shown on Ubuntu 16.04 32bit. Unfortunately building fails on Windows 8.1 32bit so I can't really test it there.

PR #836, opened 2017-08-24
Korean translation

erciccione:
Translations made by reddit user zentropicmaximillist
i've checked, updated, cleaned the monero-core_ko.ts file and prepared the PR

Also thanks to denk0815, who made the flag-button


erciccione:
I'm not a korean speaker so i can't answer your questions @mrsmonero , but thanks for your review

PR #837, opened 2017-08-24
Updated Swedish translation

snaggen:
This pull request should provide an updated Swedish translation, with obsoleted translations removed.


snaggen:
I re-did my commit, to add a missed punctuation. Hope that didn't mess things up... you were faster than I expected with the approval :)

PR #840, opened 2017-08-27
Finished translation for monero-core_ja.ts (until new strings are added)

sanecito:
Would appreciate /r/Monero 'XMR Contributor' flair for /u/xmrscott.


dEBRUYNE-1:
Added flair on Reddit.

PR #853, opened 2017-09-04
Romanian translation

mandrill-pie:
- I've created the Romanian translations file and reviewed it
- I've also added a romanian.png flag file
- also changed a few other files as instructed by @erciccione (thank you)

While I'm quite sure the translation file can successfully be used in its currrent form, I'll add a few notes that may help anyone reviewing this:

Romanian language must use one of two ways of addressing: formal or informal. I opted to always use the informal version.
I assumed the Monero voice is more like "Hi, John" and less like "Hello, Mr. Smith".

"please"
Using "please" removes any (beneficial) ambiguity about which voice (formal/informal) is being used and it accentuates the informal voice to the point it may be perceived as almost rude in Romanian. It also has the tendency to make the formal voice (if it were ever used) even more formal, to the point of becoming rigid. So "please" was removed in the few instances where it was found.

"ID"
I decided to translate "ID" to "Identifier", but if a shorter version is ever needed, "ID" or "ID-ul" (the ID) can also be used.

"backup"
Decided to use the English "backup" because it is widely accepted and used in Romanian and the only alternative was "safety copy". There is also no verb in Romanian for backup, so it would effectively become "make a safety copy" which is long and silly.

"clipboard"
I know of no elegant Romanian alternative so I used the English word, not that common in Romanian. An alternative to "copy to clipboard" in Romanian would be "copy to memory", which I don't personally like.

"ringsize"
I'm not sure if this technical term should be translated, so I left it as is.

"txpool"
I translated this to "the fund of transactions", which I thought was appropriate, but preferred to keep the technical "txpool" also, in parenthesis.

"blockchain"
I would really not translate this to the rigid and almost funny "chain of blocks" so I think a much better solution was to keep the original.

"block height"
A "block" is more like a block of flats in Romanian than anything else. This was translated directly to "height of block" even thought it sounds a bit forced/physical in Romanian.

", and"
Commas before "and" in Romanian are a big no-no (with a few very limited exceptions) so I removed them whenever I found them.

"seed"
I initially rather forcefully translated this to the Romanian "sămânță" which is more commonly used for plant/fruit seeds or kernels, because the English word "seed" is not really common (except for maybe torrent seeders and leechers). But then I thought a better approach would be to simply use "seed", because "mnemonic seed"/"sămânța mnemonică" would definitely cause a few Romanian chuckles.

"view"/"read"
Both "view" and "read" were translated to the Romanian "citi/citire" which can be used for actual reading (like a book), readings (like, eg, metrics) or reading from disk. For lack of a better alternative, "read-only" was roughly translated to either "only reading" or "only reading mode".

**And some questions:**

"spent outputs"
I'm not sure I understand this term, so it was translated only as "outputs" or rather "exits". Would "used/exhausted exits" work better?

"secret spend key"
Not sure about this one. It was translated to "secret key for spending" or rather "secret spending key". Is this still correct?

"Sweep Unmixable"
I have no idea at this time how to translate this, maybe because I don't understand the term at all.

"daemon"
Not commonly used at all in Romanian outside of technical circles. I see IT, FR and ES decided to leave it as is. I decided to use the Romanian "service", simply because "process" has strong associations with "trial" in Romanian. But "process" ("proces" in Romanian) could also be used if "service" ("serviciu" in Romanian) is not good enough.

"restore height"
Not sure "restore" is supposed to be a verb here, I assumed it wasn't. This was translated to "the height to go back to".

"success"
As part of WizzardMain, I translated "Success" to an informal "We're done" or "We've finished", because the Romanian "succes" usually means "good luck" when used by itself alone. Is this acceptable?

Also:
While I did my best to translate every single word, if anyone is ever in doubt about the correct translation to Romanian, I believe using the English word is more often than not the right route. Words like "backup, download, upload, copy, paste, upgrade, update", even possibly "log, item, embedded, seed, pool, etc", are commonly more used in normal conversations than their Romanian counterparts.


medusadigital:
+next release

PR #868, opened 2017-09-09
pro: add libreadline to link line for Linux

radfish:
Otherwise linking fails with undefined symbols `rl_*` (at least on Arch Linux x86_64).

Probably also needed for iOS case below (near -lepee), but I can't test iOS, so not adding that to this PR.

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -91,7 +91,8 @@ LIBS += -L$$WALLET_ROOT/lib \
         -lwallet_merged \
         -lepee \
         -lunbound \
-        -leasylogging
+        -leasylogging \
+        -lreadline \
 }
 
 


monero-wallet-gui.pro
```diff
@@ -91,7 +91,8 @@ LIBS += -L$$WALLET_ROOT/lib \
         -lwallet_merged \
         -lepee \
         -lunbound \
-        -leasylogging
+        -leasylogging \
+        -lreadline \
```
nssy: Thanks  for the fix. I suggest escape slash can be left out since we are at the end.

radfish: Ending slash is good because that way entries can be added to the list without changing other lines. This is good for not polluting git blame. The current diff is an example of bad diff that is due to the lack of the slash.

benma:
Please move ahead with this PR. I have the same issue on archlinux, and confirm this PR fixes it.

medusadigital:
+next release

hyc:
IMO this was the wrong fix. The GUI has absolutely no use for readline. readline should have been turned off, instead. Like so:
```cmake
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3eddfca..0802bb6 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -677,7 +677,15 @@ endif()
 
 list(APPEND EXTRA_LIBRARIES ${CMAKE_DL_LIBS})
 
+# when ON - will install libwallet_merged into "lib"
+option(BUILD_GUI_DEPS "Build GUI dependencies." OFF)
+
 option(USE_READLINE "Build with GNU readline support." ON)
+
+if(BUILD_GUI_DEPS)
+  set(USE_READLINE OFF)
+endif()
+
 if(USE_READLINE)
   find_package(Readline)
   if(READLINE_FOUND AND GNU_READLINE_FOUND)
@@ -743,9 +751,6 @@ if(BUILD_DOCUMENTATION)
   endif()
 endif()
 
-# when ON - will install libwallet_merged into "lib"
-option(BUILD_GUI_DEPS "Build GUI dependencies." OFF)
-
 # This is not nice, distribution packagers should not enable this, but depend
 # on libunbound shipped with their distribution instead
 option(INSTALL_VENDORED_LIBUNBOUND "Install libunbound binary built from source vendored with this repo." OFF)
```

jonathancross:
Thanks @hyc, can you submit a PR for this?

PR #876, opened 2017-09-14
Fix weird behaviour of privacy level bar

### LINKED ISSUE #863: Windows GUI ringsize UI glitch
GUI version: v0.10.3.1-151-ge217103 (it would fantastic if the version was copy/pastable)

Clicking slightly to the left of medium privacy level gives a ringsize of 8. Slightly to the right gives 9 and a bit further (but before the next mark) 10. I think this is not the intended behaviour.

MaxXor:
This PR fixes the weird behaviour of the privacy level bar described in #863. I fixed it by hiding the ring size of 9, so it's a bit of a hack. The actual bug is that the ring size of 9 is too close to the 8. Couldn't get it to work right now as I'm still new to QT.

Fixes #863

### DIFF
--- a/components/PrivacyLevel.qml
+++ b/components/PrivacyLevel.qml
@@ -95,7 +95,7 @@ Item {
                     }
                 }
 
-                if(index !== -1) {
+                if(index !== -1 && index !== 4) {
                     fillRect.width = Qt.binding(function(){ return row.positions[index].currentX + row.x })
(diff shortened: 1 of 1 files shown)


medusadigital:
+next release

Jaqueeee:
I'm working on a more permanent fix for this one
-next release

PR #885, opened 2017-09-19
Windows installer: Updates for release Helium Hydra

rbrunner7:
I updated the Windows installer script and the related text files for release Helium Hydra.

There are no substantial / behaviour changes, but a lot of textual changes mostly because of the new release name, and changes in the 2 pictures for the installer.

I was not sure which branch is the correct one to make the pull request against, *master* or *release-v0.11.0.0*, so I chose *master* for this first attempt. Just zap the pull request and inform me for a redo if that was wrong.


rbrunner7:
The second commit mitigates a problem with the CLI wallet that somebody reported on reddit: If you give a pure filename without a path when creating new wallets with the CLI program it tries to create them in the current directory. Started by an icon without *Start in:* set current directory is the program directory, e.g. *C:\Program Files\Monero GUI Wallet* which is not writable, which leads to an error and an abort.

Modification: The icon for the CLI wallet switches to the wallet directory as per convention of the GUI wallet, e.g. *C:\users\xyz\Documents\Monero\wallets* which is of course writeable.

QuickBASIC:
Beautiful. Sorry, I didn't comment on the PR for the bug. I don't know you had updated it for 0.11.0.0 already.

Recommend maybe mentioning of the default wallet location for CLI in the HTML readme.

Also, since the intent is to make it user friendly and unthreatening , have you considered using the `start` command in the batch to set a window title for the CLI? In the same vein `@echo off` could be used at the beginning of the batch to hide the command for being run. Lastly, you might should avoid explicitly using `C:\Program Files`. Use the environment variable `%PROGRAMFILES%`.

Awesome work!

rbrunner7:
> Awesome work!

Thanks for the positive feedback!

> Lastly, you might should avoid explicitly using C:\Program Files. Use the environment variable %PROGRAMFILES%

Don't worry, that will work out: That batchfile is not static, but it's dynamically "constructed" by the installer (see procedure *CurStepChanged* in the .iss script file), because not even using *%PROGRAMFILES%* would be flexible enough: The user is free to specify *any* folder after all, and in any case the batchfile must reflect that.

Your ideas to improve the CLI starting and that batchfile are good, but I am a little reluctant to continue to modify nevertheless: I think there will be many more things to change and improve once we will have *actual* use of  and feedback about the installer, and maybe we can bundle further modifications for a "release two". What do you think?

After waiting quite some time I hope FluffyPony et al will find the time to put this into service soon :)

danrmiller:
> 4. Start Inno Setup, load Monero.iss and compile it

Tell me the exact syntax to do this from the command line, and I will add the step to the buildbot job.

rbrunner7:
The InnoSetup command-line compiler is the program *ISCC.exe*. For the Monero setup script the command line would be:

`ISCC /Q Monero.ISS`

The `/Q` stands for *Quiet* for printing errors only; without that switch the compiler prints quite a lot of progress information that seems to me to have no use in the context of a build bot.

The program sets `ERRORLEVEL`: 0 means ok, 1 means some command line arguments were not ok, 2 means there was some error compiling and/or exeuting the script.

The program really seems to be designed for batch execution; I could not get it to display anything interactive that could disturb the build bot execution.

For reference: [Command Line Compiler Help](http://www.jrsoftware.org/ishelp/index.php?topic=compilercmdline)

danrmiller:
In addition to monero-wallet-cli.exe that I mentioned in the previous comment, this installer also packages

monero-wallet-rpc.exe  
monero-blockchain-import.exe  
monero-blockchain-export.exe  
monero-utils-deserialize.exe 

Where does monero-utils-deserialize.exe come from, is it still included with monero?

danrmiller:
I don't think start-low-graphics-mode.bat exists anymore either.

danrmiller:
And where does opengl32sw.dll come from?

rbrunner7:
Well, about the presence or absence of all these files: I always ever worked with the files in the official public download archive for the WIN64 GUI wallet. No idea who "built" these, either automated or even by hand.

I left out a few files where I was quite sure that they won't be needed (as mentioned in the comment in the script file, e.g. the QML debugger files), but otherwise I used the strategy: If it's in the official download, I package it.

Anyway, I think it's a good idea that after somebody installs this, they have the *complete* Monero software installed. No need to hunt down yet another installer e.g. for the CLI wallet, etc.

danrmiller:
Thanks, I found them, I was looking in the linux file.

medusadigital:
+next release

danrmiller:
How would you dynamically determine the version strings with ISS instead of all the hardcoded stuff?

rbrunner7:
Probably program it in Pascal, as part of the .iss file. A quick Google search seems to confirm it's possible, see [here](https://stackoverflow.com/questions/14530504/innosetup-how-to-declare-a-variable-by-reading-from-a-file). Another possibility could be to set some variables on the command line when calling ISCC and then use the variable values in the .iss.

But I am quite unsure whether it would be worth it: The .iss gets more complex, more brittle, more difficult to test / run "by hand". Maybe even two versions would result: One for just building the setup stand-alone, and another one for running under Buildbot.

It looks like you could automate the version number then, but not everything: I don't think that there is a way to "automate" the list of the files to package. So, if a new version has only a single change in that file list, somebody will have to edit the .iss anyway.

That's why I have doubts about the cost-benefit ratio of the whole version number thing - but of course I don't have to babysit a complex build environment and the corresponding experiences ...

QuickBASIC:
I'm not certain that you couldn't list the files programmatically in a way that a script builds the ISS file to begin with, then ISCC can compile from the generated ISS. 

Although, honestly think the goal should be build a solution to unzip the release from the installer (this is possible Inno) so that the person compiling releases should just have to dump the release zip into the folder and build the installer.

rbrunner7:
Updated, bumped version number from 0.11.0.0 to 0.11.1.0

PR #897, opened 2017-10-04
Run get_libwallet_api.sh using its own shebang

Nexion:
Not every system has bash as default shell. Running this script in zsh environment leads to build failure.


medusadigital:
+next release

PR #793, opened 2017-07-11
Subaddresses with multiple accounts support (preliminary)

kenshi84:
screen shot 2017-07-11 at 18 57 23
screen shot 2017-07-11 at 18 58 22

https://streamable.com/3nq9i

PR #908, opened 2017-10-20
Deal with API changes due to subaddresses, with no visible change

kenshi84:


src/libwalletqt/Subaddress.cpp
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    emit refreshFinished();
+    return m_rows;
+}
+
+Monero::SubaddressRow * Subaddress::getRow(int index) const
+{
+    return m_rows.at(index);
```
moneromooo-monero: Does this need a range test ?

src/libwalletqt/Subaddress.h
```diff
@@ -0,0 +1,33 @@
...
+    void refreshStarted() const;
+    void refreshFinished() const;
+
+public slots:
+
+private:
+    explicit Subaddress(Monero::Subaddress * subaddressImpl, QObject *parent);
+    friend class Wallet;
+    Monero::Subaddress * m_subaddressImpl;
+    mutable QList<Monero::SubaddressRow*> m_rows;
```
moneromooo-monero: Why is this mutable ? If it's because addRow/setLabel/refresh is const, then why are those const ?

src/libwalletqt/SubaddressAccount.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+    m_subaddressAccountImpl->addRow(label.toStdString());
+    getAll(true);
+}
+
+void SubaddressAccount::setLabel(quint32 accountIndex, const QString &label) const
+{
+    m_subaddressAccountImpl->setLabel(accountIndex, label.toStdString());
+    getAll(true);
+}
```
moneromooo-monero: That code looks a lot like code above. Can it be factored ? Especially the  getAll functions.

src/libwalletqt/SubaddressAccount.h
```diff
@@ -0,0 +1,33 @@
...
+    void refreshStarted() const;
+    void refreshFinished() const;
+
+public slots:
+
+private:
+    explicit SubaddressAccount(Monero::SubaddressAccount * subaddressAccountImpl, QObject *parent);
+    friend class Wallet;
+    Monero::SubaddressAccount * m_subaddressAccountImpl;
+    mutable QList<Monero::SubaddressAccountRow*> m_rows;
```
moneromooo-monero: same comment about mutable

src/libwalletqt/Wallet.cpp
```diff
@@ -457,6 +515,34 @@ AddressBookModel *Wallet::addressBookModel() const
     return m_addressBookModel;
 }
 
+Subaddress *Wallet::subaddress() const
+{
+    return m_subaddress;
+}
+
+SubaddressModel *Wallet::subaddressModel() const
+{
```
moneromooo-monero: Given this does not return a const SuaddressModel, and it const casts this below, should this const be removed ?

src/libwalletqt/Wallet.h
```diff
@@ -294,8 +333,13 @@ class Wallet : public QObject
     int     m_connectionStatusTtl;
     mutable QTime   m_connectionStatusTime;
     mutable bool    m_initialized;
+    uint32_t m_currentSubaddressAccount;
     AddressBook * m_addressBook;
     mutable AddressBookModel * m_addressBookModel;
+    Subaddress * m_subaddress;
+    mutable SubaddressModel * m_subaddressModel;
+    SubaddressAccount * m_subaddressAccount;
+    mutable SubaddressAccountModel * m_subaddressAccountModel;
```
moneromooo-monero: These mutable qualifiers would like go with the const on the getter above removed

src/model/SubaddressAccountModel.h
```diff
@@ -0,0 +1,36 @@
...
+
+class SubaddressAccount;
+
+class SubaddressAccountModel : public QAbstractListModel
+{
+    Q_OBJECT
+
+public:
+    enum SubaddressAccountRowRole {
+        SubaddressAccountRole = Qt::UserRole + 1, // for the SubaddressAccountRow object;
```
moneromooo-monero: This one is not handled in roleNames and data. Is it a placeholder only?

Jaqueeee:
Looks good to me. Can be merged once @moneromooo-monero's remarks are addressed/fixed. 
@kenshi84 @luigi1111

src/libwalletqt/Subaddress.cpp
kenshi84: Range check was already done in the caller's side, but added a range test anyway.

src/libwalletqt/Subaddress.h
kenshi84: I removed mutable, and made those member functions non-const.

src/libwalletqt/SubaddressAccount.cpp
kenshi84: I don't know how they can be refactored. Personally I also don't find them too redundant.

src/libwalletqt/SubaddressAccount.h
kenshi84: OK

src/libwalletqt/Wallet.cpp
kenshi84: I think this member function was made `const` and the `m_subaddressAccountModel` member data `mutable` because `m_subaddressAccountModel` gets assigned a newly instantiated object when it's pointed to `nullptr` (i.e., *NOT* because it returns a non-const `SubaddressModel*` -- it's perfectly OK for a const member function to return a (copy of ) member data which is a pointer to a non-const object). But anyway I made those functions (`subaddress(), subaddressModel(), subaddressAccount(), subaddressAccountModel()`) non-const and removed `mutable`, as I don't see any problem in doing so.

src/model/SubaddressAccountModel.h
kenshi84: OK, removed

src/libwalletqt/Wallet.h
kenshi84: Removed

Jaqueeee:
@kenshi84 This PR becomes obsolete when https://github.com/monero-project/monero/pull/2703 is merged right?

kenshi84:
@Jaqueeee Correct, so closing this. Comments from @moneromooo-monero will be reflected in #793 .

PR #914, opened 2017-10-23
fixed create new wallet can't  scan money

Cherisher:
The new wallet can't scan money from blockchain if you don't recovery wallet from keys or  seed.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -269,7 +269,9 @@ ApplicationWindow {
         middlePanel.paymentClicked.connect(handlePayment);
         middlePanel.sweepUnmixableClicked.connect(handleSweepUnmixable);
         middlePanel.checkPaymentClicked.connect(handleCheckPayment);
-
+        
+        persistentSettings.is_recovering = true;
+        
         console.log("initializing with daemon address: ", persistentSettings.daemon_address)
         console.log("Recovering from seed: ", persistentSettings.is_recovering)
         console.log("restore Height", persistentSettings.restore_height)


dEBRUYNE-1:
Can you reliably reproduce this issue? Also is it present in the v0.11.0.0? If so, I'd have expected more reports.

main.qml
```diff
@@ -269,7 +269,9 @@ ApplicationWindow {
         middlePanel.paymentClicked.connect(handlePayment);
         middlePanel.sweepUnmixableClicked.connect(handleSweepUnmixable);
         middlePanel.checkPaymentClicked.connect(handleCheckPayment);
-
+        
+        persistentSettings.is_recovering = true;
```
Jaqueeee: I'm afraid this is going to create other bugs. Please provide more info on how to recreate the issues you're having.

Cherisher: If you create a new wallet(wallet A) by the walllet-gui, then transform some money from other wallet(wallet B) to A, and you can find this bug,  A can't find any recored. except to recover wallet according to keys or seed words.

dEBRUYNE-1: Is B also created by the GUI?

Cherisher: Whether B is created by GUI or by command line, the problem always exists. You can easily test it in your environment. Also is it present in the v0.11.0.0

dEBRUYNE-1:
Could you try the new release?

Cherisher:
I don't find any code to fixed this problem, can you tell me where are the codes. I will try the new release as soon as possible.Thanks

Jaqueeee:
@Cherisher fix is in https://github.com/monero-project/monero/pull/2724

Cherisher:
This bug has been soluted by moerno on v0.11.1.0.

PR #917, opened 2017-10-23
add option to change wallet creation height and rescan wallet cache

Jaqueeee:


pages/Settings.qml
```diff
@@ -497,9 +497,78 @@ Rectangle {
...
+                    restoreHeightRow.visible = false
+
+                    // Show confirmation dialog
+                    confirmationDialog.title = qsTr("Rescan wallet cache") + translationManager.emptyString;
+                    confirmationDialog.text  = qsTr("Are you sure you want to rebuild the wallet cache?\n"
+                                                    + "The following information will be deleted\n"
+                                                    + "- Recipient addresses\n"
+                                                    + "- Tx keys\n"
+                                                    + "- Tx descriptions\n\n"
+                                                    + "The old wallet cache file will be renamed and can be restored later.\n"
```
moneromooo-monero: Is this something the GUI does ? wallet2 doesn't AFAIK.

Jaqueeee: Correct. Added that a while back, but never used it. 
https://github.com/monero-project/monero-core/blob/master/src/libwalletqt/WalletManager.cpp#L360-L376

Jaqueeee:
Already merged

PR #929, opened 2017-10-24
show correct height after switching wallets

Jaqueeee:


medusadigital:
tested https://github.com/monero-project/monero-core/pull/929. Feature it a lil dangerous still, but good to go for now. patch does its job.

medusadigital:
+next release

PR #938, opened 2017-10-29
Update monero-wallet-gui.pro

italocoin-project:
When you build with "release-static" with Ubuntu 17.10 on x86_64 C/++ 7 you get this error:

```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libcrypto.a(dso_dlfcn.o): undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

Because you need to link -ldl

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -246,6 +246,11 @@ linux {
             -Wl,-Bdynamic \
             -lGL
     }
+     
+  CONFIG(static) {
+	LIBS+=-ldl
+	  }
+    
     # currently monero has an issue with "static" build and linunwind-dev,
     # so we link libunwind-dev only for non-Ubuntu distros
     CONFIG(libunwind_off) {


monero-wallet-gui.pro
```diff
@@ -246,6 +246,11 @@ linux {
             -Wl,-Bdynamic \
             -lGL
     }
+
```
jonathancross: Is there a reason you added a new `CONFIG(static) {}` block here rather than adding to the existing one above on line 220?

More importantly, the change you are proposing undoes a fix from last february: https://github.com/monero-project/monero-core/pull/478 On that PR, @Jaqueeee said "IIRC this would kill the static build. So please add this on non static build only." in [this comment](https://github.com/monero-project/monero-core/pull/478#discussion_r102642782).

Would be good to find out if something changed and `ldl` now works for static and non-static or if this is only correct for some architectures.

Jaqueeee: Thanks for the ping @jonathancross. Libdl is the dynamic linker. It's only required if you're linking dynamic libraries. The reason for the static build to fail could be that you don't have static versions of all the dependent libraries. Note that a static GUI build also requires a statically built Qt (built from source). Use `./build.sh release`instead of `./buid.sh release-static` if you don't need a static build.

italocoin-project: If i added above the line 249 i would have got the error i was mentioning in the pool "DSO missing from command line" this error i'm getting only if i compile with release-static, i guess there are some codes that need -ldl for some reason, i'm pretty new in c++ i don't really understand why is working if i add the -ldl just after the line 249. What i was sure about this error was that the -ldl was missing.
I'm using Ubuntu 17.10 86_64 with gcc/c++ 7 and boost 1_64, someone more expert than me maybe can explain what and why is -ldl needed for my specs.

jonathancross: Unfortunately I'm also inexperienced with C++  :-)

@Jaqueeee Are you saying that `DSO missing from command line` is expected if Qt was not built from source?  Can we add a check to provide a more meaningful message to users?
TIL that [the order in which libs appear on the commandline is critical](https://stackoverflow.com/questions/19901934/strange-linking-error-dso-missing-from-command-line#19905704) and can cause this error message as well.

@italocoin-project Seems you will need to build Qt if you want to use `release-static`.

italocoin-project: @jonathancross yes, I've realised that after @Jaqueeee ping, i will compile Qt in the next few  days and i'll let a reply here if that was the cause.

Jaqueeee:
@italocoin-project Do you still want this to be merged? Please close otherwise.

monero-wallet-gui.pro
italocoin-project: So, i've compiled the QT from source and i still get this error! Qt compile from source won't solve the problem in my end! Any other ideas?

italocoin-project:
So, i've compiled the QT from source and i still get this error! Qt compile from source won't solve the problem in my end! Any other ideas?

bomb-on:
@pazos You built `release-static` version of GUI wallet in this way?

pazos:
@bomb-on: Sorry, I miss the point. All I built was release/debug, not static ones.

At this point building a "static" application for linux needs some steps which aren't covered in the README. I see no point for doing that, but If you want to try  [look at this example:](https://paste.fedoraproject.org/paste/OrADJLKyws4qFq21XuOTng/raw), credits to @skftn, I think :)

And yes, -ldl is needed to be able to link the app for linux

Again, sorry for the missinformation

bomb-on:
@pazos No worries, I didn't consider your reply as a "misinformation" ;)

Thanks for the link, I'll study it and try to apply things which I missed during my tryout of building static gui wallet.

Not sure why did you mention that you don't see the point in making static build, isn't current linux gui wallet (the one which you can download as official release) built in the same way? I tried it out on completely fresh Ubuntu 16.04, performed only `sudo apt update && sudo apt upgrade`, even deleted folders named `qt` and `plugins` which come with the release and my gui wallet was working without a single problem.

italocoin-project:
This can be closed, thank you

PR #966, opened 2017-11-20
Bugfix with tx proof + update

stoffu:
Fixed stupid mistype of function names.
Also `getTxProof` handles failure cases with C++ exception instead of additional string arg.


danrmiller:
Works for me. Looks good to me.

PR #967, opened 2017-11-20
Spend proof

stoffu:


main.qml
```diff
@@ -812,6 +820,12 @@ ApplicationWindow {
                 informationPopup.text = qsTr("This address received nothing");
             }
         }
+        else if (results.length == 2 && results[0] == "true") {
```
skftn: Just a small thing I've noticed in the .qml javascript code here and there: might be best to explicitly use `===` instead of `==` so it doesnt coerce and do possibly crazy stuff (alltho probably rare).

stoffu: Thanks, fixed.

Issue #887, opened 2017-09-20
[FEATURE REQUEST] (send) Warn it's a watch only wallet; can't spend


zzaappedd:
My wallet is watch only and I can't spend. Today I was ready to spend for the 1st time and had forgotten it was watch only. Maybe a flag somewhere can remind users that watch only wallets can't spend. 
1st time posting here and don't know if this is the right place.
Thanks to whomever listens!

dEBRUYNE-1:
+enhancement

dEBRUYNE-1:
+resolved

PR #976, opened 2017-11-28
Left panel: add view only wallet label

Timo614:
I took a look at the associated issue #887 where the user did not realize they were on a view only wallet and could not spend money directly.

With view only wallets you can create a transaction which you can then sign on a normal wallet and later submit via the view only wallet. It does this so that you can keep your main wallet offline if you want and just use it for signing transactions generated by the view wallet. 

I think the best call here then is just a label like the testnet one that appears if you're on the testnet under the Monero branding on the top left.

The changes were pretty straightforward to add one. I've included photos of the GUI here for mainnet, mainnet view only, testnet, testnet view only to see the changes more easily.

mainnet
mainnet-view
testnet
testnet-view

Note: this branch currently fails to build by itself due to master currently having a broken build. Pulling in the changes from this commit https://github.com/monero-project/monero-gui/pull/966/commits/d8f3a52378ad61ee0d81f156bd2fc15b8a66293c / PR #966 fixes it (which I did locally but stashed those for this PR). I'll rebase once the associated PR for that commit is merged into master.

Ref: #887

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -91,13 +91,25 @@ Rectangle {
         }
 
         Text {
+            id: viewOnlyLabel
+            visible: viewOnly
+            text: qsTr("View Only") + translationManager.emptyString
+            anchors.top: logo.bottom
+            anchors.topMargin: 5
+            anchors.left: parent.left
+            anchors.leftMargin: 50
+            font.bold: true
+            color: "blue"
+        }
+
+        Text {
             id: testnetLabel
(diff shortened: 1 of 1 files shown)


Timo614:
#887 seems related as well from the comments

skftn:
Good work. I agree view-only status should be displayed in the leftPanel.

Timo614:
Just a heads up: I rebased and retested this to pick up the fix from #966. Should be set.

PR #977, opened 2017-11-29
Subaddresses minimal version: supports default account only

kenshi84:
This is an alternative to #793 where the support for account management is dropped which makes the interface change smaller. The "Receive" page has a few UI glitches left unresolved due to my limited Qt skills:

- The "Address" text input box at the top is redundant; the user should be able to directly select and copy address texts from the list.

- ~~The QR code image gets chopped off because of the subaddresses list.~~


kenshi84:
The QR code part was solved by @skftn . Thanks!

kenshi84:
@skftn did further adjustment in https://github.com/kenshi84/monero-core/pull/2. Not knowing how to preserve his commit while rebasing to the latest master, I had to squash it; hope this was OK. Thank you!

src/libwalletqt/Wallet.cpp
```diff
@@ -458,6 +512,19 @@ AddressBookModel *Wallet::addressBookModel() const
     return m_addressBookModel;
 }
 
+Subaddress *Wallet::subaddress() const
+{
+    return m_subaddress;
+}
```
moneromooo-monero: I think I've already mentioned this in a previous PR, but I don't like this system of const functions returning non const pointers to their data. It can be kept as is though, not my repo.

src/model/SubaddressModel.cpp
```diff
@@ -0,0 +1,59 @@
+#include "SubaddressModel.h"
```
moneromooo-monero: Copyright header

src/model/SubaddressModel.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    return m_subaddress->count();
+}
+
+QVariant SubaddressModel::data(const QModelIndex &index, int role) const
+{
+    if (!index.isValid() || index.row() < 0 || (unsigned)index.row() >= m_subaddress->count())
+        return {};
+
+    Monero::SubaddressRow * sr = m_subaddress->getRow(index.row());
```
moneromooo-monero: Can that return NULL ? It's dereferenced below.

src/model/SubaddressModel.h
```diff
@@ -0,0 +1,34 @@
+#ifndef SUBADDRESSMODEL_H
```
moneromooo-monero: Header

components/HistoryTable.qml
```diff
@@ -266,6 +266,45 @@ ListView {
...
+            }
+            Item { //separator
+                width: 5
+                height: 14
+            }
+            // -- "Index" value
+            Text {
+                anchors.bottom: parent.bottom
+                font.family: "Arial"
+                font.pixelSize: 13
```
skftn: Perhaps make all, or almost all constant integer values end with `* scaleRatio` ? I might be reviewing my own code here, in which case I have forgotten it :)

pages/Receive.qml
```diff
@@ -192,27 +195,78 @@ Rectangle {
...
                 width: mainLayout.labelWidth
             }
 
-            LineEdit {
-                id: addressLine
-                fontSize: mainLayout.lineEditFontSize
-                placeholderText: qsTr("ReadOnly wallet address displayed here") + translationManager.emptyString;
-                readOnly: true
-                width: mainLayout.editWidth
+            ListModel {
```
skftn: Should this testModel end up in master?

src/libwalletqt/Subaddress.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <QDebug>
+
+Subaddress::Subaddress(Monero::Subaddress *subaddressImpl, QObject *parent)
+  : QObject(parent), m_subaddressImpl(subaddressImpl)
+{
+    qDebug(__FUNCTION__);
+    getAll();
+}
+
+QList<Monero::SubaddressRow*> Subaddress::getAll(bool update) const
```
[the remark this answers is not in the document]
kenshi84: I'm also not quite familiar with the GUI codebase, and I've designed these function interfaces by simply copying and adapting the existing practice. The same idiom

```c++
virtual std::vector<T*> getAll() const = 0;
```

is used repeatedly in wallet2_api.h, so starting to use refcount here (in order to avoid potential bugs) would imply that all other similar parts need to be updated as well. Since I haven't seen any crashes or issues with the current way, I'd rather keep it as is, and perhaps make a separate PR later that replaces all these pointers with refcount in order to be extra sure (but again, my familiarity with the GUI codebase is not enough for assessing if it actually is an improvement or not).

src/libwalletqt/Wallet.cpp
```diff
@@ -241,14 +243,63 @@ bool Wallet::viewOnly() const
...
+void Wallet::switchSubaddressAccount(quint32 accountIndex)
+{
+    if (accountIndex < numSubaddressAccounts())
+    {
+        m_currentSubaddressAccount = accountIndex;
+        m_subaddress->refresh(m_currentSubaddressAccount);
+        m_history->refresh(m_currentSubaddressAccount);
+    }
+}
```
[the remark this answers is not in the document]
kenshi84: You mean, it can throw? I don't see anywhere it can throw, both in `SubaddressImpl::refresh` and `TransactionHistoryImpl::refresh`.

src/libwalletqt/Subaddress.cpp
```diff
@@ -0,0 +1,56 @@
...
+void Subaddress::refresh(quint32 accountIndex) const
+{
+    m_subaddressImpl->refresh(accountIndex);
+    getAll(true);
+}
+
+quint64 Subaddress::count() const
+{
+    return m_rows.size();
+}
```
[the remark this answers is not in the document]
kenshi84: Inserting `getAll(false);` here somehow causes the GUI to crash, to my surprise. I'm not familiar enough with the GUI repo to know why, but following the existing practice seems just fine...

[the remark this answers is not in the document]
kenshi84: My guess: `Subaddress::getAll` invokes `SubaddressModel::startReset` which then seems to somehow trigger `SubaddressModel::rowCount` to be called which then calls `Subaddress::count`, so it goes into an infinite recursion.

src/libwalletqt/Subaddress.cpp
```diff
@@ -0,0 +1,56 @@
+#include "Subaddress.h"
```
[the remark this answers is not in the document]
kenshi84: ok

src/libwalletqt/Subaddress.h
```diff
@@ -0,0 +1,33 @@
+#ifndef SUBADDRESS_H
```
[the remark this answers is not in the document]
kenshi84: ok

src/model/SubaddressModel.cpp
kenshi84: ok

src/model/SubaddressModel.cpp
kenshi84: ok

src/model/SubaddressModel.h
kenshi84: ok

pages/Receive.qml
kenshi84: OK, removed.

components/HistoryTable.qml
kenshi84: Does this apply to every single place where constant integers are used? There are quite many, like `rowSpacing, width, height, rightMargin, fontSize, ...` in addition to `font.pixelSize`.

skftn: I remember @Jaqueeee told me to do that for all constant integers IIRC

kenshi84: > we can always change it later when I focus on the mobile part (soon).

I bet that's better since doing the mobile layout right is a big deal on its own.

luigi1111:
Merging...future changes/fixes can use a new pr (you know the drill).

PR #981, opened 2017-12-01
Cleanup html/css to simplify translation

jonathancross:
I have modified several particularly noisy chunks of UI text to remove html/css formatting which was cluttering up translation files.  This should help translators to focus on their job while removing distractions and potential sources of broken html/css UI code.

The PR is broken into 3 commits to simplify review:
1. First commit was just the result from running `lupdate -no-obsolete -pro monero-wallet-gui.pro -ts translations/monero-core.ts` (ie no code changes).
2. Second is is the only code change where I isolate the strings from the html/css code.
3. Ran `lupdate` again to show the cleaned up result.

Notes:
* There is very minor visual changes to fonts based on the simplification I made.  I can provide screen-cap comparison if needed.
* I can squash once someone (@erciccione ?) has reviewed.
* I'll also refresh other translation files then (wanted to avoid the noise for now).

### DIFF
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -331,9 +331,12 @@ Rectangle {
             Label {
                 id: trackingLabel
                 textFormat: Text.RichText
-                text: qsTr("<style type='text/css'>a {text-decoration: none; color: #FF6C3C; font-size: 14px;}</style>\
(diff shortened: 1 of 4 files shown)


jonathancross:
> Having issues building the GUI...

Yeah, I am currently only able to build using `./build.sh release-static`

PR #991, opened 2017-12-05
Switch Core with GUI in README.md, fix headings

xmr-eric:
Recently the GUI repo had its name changed from monero-core to monero-gui. This PR updates the build instructions and elsewhere with the new repo name and link.

This PR also seeks to put the heading capitalization policy inline with the main Monero readme. Capitalizing every single word in a heading begins to feel redundant and cumbersome when headings become long (and many are indeed quite long). A much better practice is capitalizing just the initial word. See: https://github.com/monero-project/monero/pull/2881

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,10 @@ Copyright (c) 2014-2017, The Monero Project
 - Web: [getmonero.org](https://getmonero.org)
 - Forum: [forum.getmonero.org](https://forum.getmonero.org)
 - Mail: [dev@getmonero.org](mailto:dev@getmonero.org)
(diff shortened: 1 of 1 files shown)


xmr-eric:
As a side note, the "Void Linux" package seems to still be named monero-core, so perhaps somebody could reach out to whomever it was in charge of that to get the name changed over.

README.md
```diff
@@ -10,7 +10,7 @@ Copyright (c) 2014-2017, The Monero Project
 - Github: [https://github.com/monero-project/monero-core](https://github.com/monero-project/monero-core)
```
MaxXor: You missed this one.

PR #889, opened 2017-09-22
Wallet password dialog refactoring + allow password to be changed

stoffu:

PR #1013, opened 2017-12-11
Left Panel Hotkeys: Fix Seed & Keys

### LINKED ISSUE #1011: Left Panel: Seed & Keys hotkey not setup
The Seeds & Keys hotkey is not setup currently so users cannot `ctrl+y` to switch to that tab.

https://github.com/monero-project/monero-gui/blob/master/main.qml#L106

The behavior should mimic the clicking of the leftPanel button where the password dialog is presented.

Timo614:
This issue fixes #1011 by adding a hotkey for the "Seed & Keys" section which behaves the same way as the leftPanel clicking by triggering the popup to confirm the user's password prior to entry.

Note: until #1005 has been merged typing a password will revert the leftPanel selection and show you an incorrect left tab. You can bypass the bug #1005 is fixing by using a password-less test wallet and clicking the enter button for the password when the dialog pops (the hotkey logic runs currently even when the leftPanel has been disabled and sets the value to match the current open tab which can cause issues during the password prompt flow to switch to a new tab).

image showing current hotkey

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -104,6 +104,11 @@ ApplicationWindow {
         else if(seq === "Ctrl+I") middlePanel.state = "Sign"
         else if(seq === "Ctrl+E") middlePanel.state = "Settings"
         else if(seq === "Ctrl+D") middlePanel.state = "Advanced"
+        else if(seq === "Ctrl+Y") {
+            leftPanel.selectItem("Keys")
+            settingsPasswordDialog.open()
+            return
+        }
         else if(seq === "Ctrl+Tab" || seq === "Alt+Tab") {
             /*
             if(middlePanel.state === "Dashboard") middlePanel.state = "Transfer"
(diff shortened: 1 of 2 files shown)


main.qml
```diff
@@ -104,6 +104,11 @@ ApplicationWindow {
         else if(seq === "Ctrl+I") middlePanel.state = "Sign"
         else if(seq === "Ctrl+E") middlePanel.state = "Settings"
         else if(seq === "Ctrl+D") middlePanel.state = "Advanced"
+        else if(seq === "Ctrl+Y") {
+            leftPanel.selectItem("Keys")
+            settingsPasswordDialog.open()
```
stoffu: - Why not doing `middlePanel.state = "Keys"` like other tabs?

- I did some refactoring on the password dialog logic in #889, so this line needs to be updated accordingly.

selsta:
I think this has been resolved by some other PR.

selsta:
ping @xmrdsc :P

xiphon:
> I think this has been resolved by some other PR.

Yep, i fixed this in https://github.com/monero-project/monero-gui/pull/1489

Timo614:
Sounds good, closing as it's fixed in #1489

PR #1014, opened 2017-12-11
Receive page: QR code bug fix for invalid payment ids

### LINKED ISSUE #904: [BUG] QR code changes when using an invalid payment ID.
Reported by charutocafe on IRC:

>When using an invalid payment ID on the Receiving tab, the QR code still changes and still exists.

QR code also changes if you type letters as payment ID.

Timo614:
Simple change to fix #904 where the QR code was incorrectly updating for invalid payment IDs. Logic just checks whether or not the payment ID is valid prior to including it as part of the QR code now. 

I've attached a few screenshots which show that the QR code only updates now when there is a valid payment id as well.

no-payment-id
invalid-payment-id
valid-payment-id

### DIFF
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -80,8 +80,8 @@ Rectangle {
           s += (nfields++ ? "&" : "?")
           s += "tx_amount=" + amount
         }
-        var pid = paymentIdLine.text.trim()
-        if (pid !== "") {
+        var pid = paymentIdLine.text.trim().toLowerCase()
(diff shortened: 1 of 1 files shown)


stoffu:
While you're at it, how about making the QR code case-insensitive by changing the line 83 to:
```javascript
         var pid = paymentIdLine.text.trim()
```

Timo614:
Sure all set (went with `toLowerCase` as the TextField `text` property returns a normal string not a QString)

PR #1017, opened 2017-12-12
Embed the translation files in the monero-wallet-gui binary

glv2:
This PR allows embedding the translation files in the binary instead of fetching them in the *translations* directory.

If a translation file exists in a *translations* directory located in the same directory as the binary, it is used in priority (this can be useful when working on translations as you don't have to recompile the whole program all the time), and if no such file is found the embedded translation file is used.

I have only tested this on GNU/Linux. If someone has a Mac or Windows machine and the time to try this, please test and see if some unexpected issues appear.

### DIFF
--- a/TranslationManager.cpp
+++ b/TranslationManager.cpp
@@ -25,18 +25,11 @@ bool TranslationManager::setLanguage(const QString &language)
         return true;
     }
 
-    // translations are compiled into app binary
-#ifdef Q_OS_MACX
-    QString dir = qApp->applicationDirPath() + "/../Resources/translations";
-#else
     QString dir = qApp->applicationDirPath() + "/translations";
-#endif
-
     QString filename = "monero-core_" + language;
 
-    qDebug("%s: loading translation file '%s' from '%s",
+    qDebug("%s: loading translation file '%s' from '%s'",
            __FUNCTION__, qPrintable(filename), qPrintable(dir));
-
 
     if (m_translator->load(filename, dir)) {
         qDebug("%s: translation for language '%s' loaded successfully",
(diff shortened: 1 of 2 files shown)


erciccione:
This would make things a lot easier i guess, will test it tomorrow. Thanks for the contribution

erciccione:
Works fine, but i don't know if the translations where left outside the binaries on purpose, for some reson. @jonathancross opinion on this?

jonathancross:
> i don't know if the translations where left outside the binaries

@erciccione Sorry, all I see is that this Mac-specific condition was added in https://github.com/monero-project/monero-gui/commit/9cd73dfbbed6e200852ca9e1c1494ea3e5eae6c3  -- maybe @mbg033 can explain more?

A guess: Maybe there is some situation where translations are used for text not available during compilation? (related to daemon, etc)

> Works fine

Thanks, does that mean you tested on a Mac?

Assuming it works correctly, one maintainability drawback I see with this approach is that anyone adding a new language will have to also update `qml.qrc`

glv2:
@jonathancross When adding a new language, the *qml.qrc* file will very probably have to be updated anyway to add a new flag, so it should not be a big problem.

erciccione:
@jonathancross Ok, thanks for that. I tested it on Linux, would be great to have it tested on Mac and Windows before merging

bjacquin:
Hey @glv2, that's a really cool feature. I am getting the following error after cleaning my local tree, it looks like we have some inter-dependency or ordering issue.

```
Project MESSAGE: Building with libunwind
RCC: Error in 'qml.qrc': Cannot find file 'build/translations/monero-core_ar.qm'
RCC: Error in 'qml.qrc': Cannot find file 'build/translations/monero-core_cs.qm'
RCC: Error in 'qml.qrc': Cannot find file 'build/translations/monero-core_da.qm'
[..]
```

Is this intended ?

glv2:
@bjacquin, I added a translation files compilation step when initializing the project the first time. It should make the ```RCC: Error in 'qml.qrc': Cannot find file ...``` error messages disappear.

bjacquin:
Indeed, this does remove the errors, thanks.

One more question, should `build/` be hardcoded here? it looks like one may be building in a different directory. Even though its been used in `build.sh`, it is not mandatory to use `build/` before this change. One can build in a separate directory or in the main monoro-gui directory by just using qmake and specify where `monero-wallet-gui.pro` is located

glv2:
@bjacquin, I found a way to generate and include a resource file for translations automatically. It should allow compilation in a directory different from ```build/``` if necessary.

bjacquin:
Thank @glv2, I confirm I can now build it outside of build/, thanks a lot!

glv2:
I rebase on the current master branch and tested that it builds and works on Linux and Windows (64 bits).

PR #1021, opened 2017-12-13
add wallet name to left panel

### LINKED ISSUE #655: [FEATURE REQUEST] Display wallet name somewhere
A request from visdude at Bitcointalk:

> It would be nice to have the current/active wallet or account name on the GUI though as most users most likely maintain more than one wallet/account. It just makes sense.

https://bitcointalk.org/index.php?topic=583449.msg17299701#msg17299701

jacobzipper:
Resolves #655  by adding the wallet name to the top of the left panel like so.
walletname

### DIFF
--- a/pages/Settings.qml
+++ b/pages/Settings.qml
@@ -567,6 +567,10 @@ Rectangle {
         }
         TextBlock {
             Layout.fillWidth: true
+            text: qsTr("Wallet Name: ") + walletName + translationManager.emptyString
+        }
+        TextBlock {
+            Layout.fillWidth: true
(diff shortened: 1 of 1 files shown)


dEBRUYNE-1:
I feel like this is going to clutter the GUI wallet. Can we add it under `Debug info`?

xmr-eric:
I agree. Design wise the upper left is one of the most highly visible parts of the software, so having the wallet balance there is great. The wallet file name is not as high a priority. 

A better place would be, IMHO, in the lower left over Network status, or in the title bar (Eg: “Monero - zipper-restored”.

PR #1025, opened 2017-12-16
hello windows

alexeyneu:
compare results ( stop node , stop on close )

   
so transfer 1xmr to 454cAu1GWpYV3HdEuCdT6RCPWeaWHDbFWATAhfRfFFjHaEExVDBYGLdJQfUVzhaiyd9j1qX4pqWj9ELsvXgYsZhuUeRfVDy

### DIFF
--- a/src/daemon/DaemonManager.cpp
+++ b/src/daemon/DaemonManager.cpp
@@ -82,8 +82,14 @@ bool DaemonManager::start(const QString &flags, bool testnet, const QString &dat
     connect (m_daemon, SIGNAL(readyReadStandardError()), this, SLOT(printError()));
 
     // Start monerod
-    bool started = m_daemon->startDetached(m_monerod, arguments);
 
+#ifndef Q_OS_WIN
+    bool started = m_daemon->startDetached(m_monerod, arguments);
+#else
+	long long bt;
+	bool started = m_daemon->startDetached(m_monerod, arguments,QString(),&bt);
+	if(started) b =  OpenProcess(SYNCHRONIZE, FALSE, bt);
+#endif
     // add state changed listener
     connect(m_daemon,SIGNAL(stateChanged(QProcess::ProcessState)),this,SLOT(stateChanged(QProcess::ProcessState)));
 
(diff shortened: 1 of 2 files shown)


src/daemon/DaemonManager.cpp
```diff
@@ -77,13 +77,21 @@ bool DaemonManager::start(const QString &flags, bool testnet, const QString &dat
     m_daemon = new QProcess();
     initialized = true;
 
+
```
leonklingele: Why?

src/daemon/DaemonManager.cpp
```diff
@@ -93,6 +101,9 @@ bool DaemonManager::start(const QString &flags, bool testnet, const QString &dat
         return false;
     }
 
+
```
leonklingele: Why?

src/daemon/DaemonManager.cpp
alexeyneu: test unit's tracks

skftn:
Could you perhaps explain what this commit actually does (for those of us that are not C++ proficient)

alexeyneu:
what you have now is:
```C++
while(1)
{
     sleep(2 seconds);    //thread is frozen
     checkforsmth();      // put to account that this is request for an answer of almost stalled monerod process 
    if ( found ) break;
}
```
where my code is:
```C++
WaitForSingleObject(smth , INFINITE);  // monerod process handle  goes as event one ,fired by windows when  process get done.  
```
So on average mine is faster by second

stoffu:
Please use whitespaces instead of tabs for indentation, since the existing code is written that way.

stoffu:
The commit message needs to be more descriptive.

pazos:
@alexeyneu: could you make this portable (your implementation for win and [pevents](https://github.com/neosmart/pevents) for posix) ?

alexeyneu:
it's not possible to do with this tool. linux wont signall  process handle anyway . I don't know much about linux mt engine for now so cant say more

selsta:
Can someone explain what this change does exactly? Should we merge this?

xiphon:
@selsta 

> Can someone explain what this change does exactly? 

The change attempts to implement Windows-specific child process (monero daemon) status handling.

> Should we merge this?

The idea is valid, but i don't like the particular patch.  
I would prefer not to merge it.

xmrdsc:
Thanks for the review @xiphon and thanks for the contribution @alexeyneu 

+resolved

PR #1030, opened 2017-12-19
windows

alexeyneu:
build of both gui apps (current and this fork's) failed on linkin step coz smth in current setup drags sjlj .


skftn:
+resolved

PR #1074, opened 2018-01-14
Add MacPorts compatibility

IPGlider:


jtgrassie:
Why hardcode library search path? What about users who may have MacPorts but wish to sue system installed libraries or use homebrew? Surely best left to the person compiling what search paths to use.

IPGlider:
Changed order and check.

Homebrew paths are already hardcoded therefore having default MacPorts paths hardcoded follows the same pattern.

Having MacPorts as a fallback to Homebrew does not interfere with the current behaviour for Homebrew users and makes things easier to MacPorts devs.

jtgrassie:
You're missing my point. The search paths shouldn't be hardcoded, they should be defined by the system environment or specified / overridden by the person compiling.

selsta:
Not sure what to do with this, how do other projects support both homebrew and MacPorts?

jtgrassie:
@selsta 
> how do other projects support both homebrew and MacPorts?

They do not hardcode library paths. If any paths are added, they are just hints like `/usr/, /usr/local, /opt/local`. This PR can probably be merged though as a stop-gap until this has been properly refactored.

PR #1075, opened 2018-01-14
Store wallet log in ~/Library/Logs/ in MacOS

IPGlider:
In MacOS it is not usual to modify anything inside an application. If the user that runs the application doesn't have writing permissions the log will not be saved.

This modifies the path to store them in MacOS' standard location for application logs.

### DIFF
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -658,7 +658,13 @@ QString Wallet::getDaemonLogPath() const
 
 QString Wallet::getWalletLogPath() const
 {
-    return QCoreApplication::applicationDirPath() + "/monero-wallet-gui.log";
+    const QString filename("monero-wallet-gui.log");
+
+#ifdef Q_OS_MACOS
+    return QStandardPaths::standardLocations(QStandardPaths::HomeLocation).at(0) + "/Library/Logs/" + filename;
+#else
+    return QCoreApplication::applicationDirPath() + "/" + filename;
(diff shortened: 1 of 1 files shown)


src/libwalletqt/Wallet.cpp
```diff
@@ -658,7 +658,11 @@ QString Wallet::getDaemonLogPath() const
 
 QString Wallet::getWalletLogPath() const
 {
-    return QCoreApplication::applicationDirPath() + "/monero-wallet-gui.log";
+    #ifdef Q_OS_MACOS
+        return QStandardPaths::standardLocations(QStandardPaths::HomeLocation).at(0) + "/Library/Logs/monero-wallet-gui.log";
```
leonklingele: Why not `#define GUI_LOG_FILE_NAME "monero-wallet-gui.log"`?

Maxithi:
Linking  https://github.com/monero-project/monero/pull/3052.
It has been chosen to handle the MacOS path as the UNIX path in the core code.

IPGlider:
@Maxithi Does that apply to the command line applications only or to the GUI as well?

Maxithi:
@IPGlider This currently only affects the CLI.

IPGlider:
Yes, "Library" is right. For reference:

https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html

PR #1076, opened 2018-01-14
Refresh translations

jonathancross:
Refresh translations since #981 has been merged.
Command: `lupdate -no-obsolete -pro monero-wallet-gui.pro`
(also picked up a number of additional strings which were added in the last few months)

I added a second commit which contains manual fixes for a few situations where `lupdate` got confused about numbers.
Example:
```
(%1/10 次確認) {1/%2 ?}
```

Translations system was confused that we now have `%1` and `%2` parameters.  So I fixed as:
```
(%1/%2 次確認)
```

@erciccione - Care to have a look?

### DIFF
--- a/translations/monero-core.ts
+++ b/translations/monero-core.ts
@@ -280,22 +280,27 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../components/HistoryTable.qml" line="263"/>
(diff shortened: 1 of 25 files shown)


jonathancross:
Thanks @erciccione !

LvMsterfild:
Is it good link for update or is it more fresh (russian translation) ?

PR #1083, opened 2018-01-17
Transfer: automatically use low priority if no backlog is present (requires #3123)

stoffu:


pages/Transfer.qml
```diff
@@ -642,7 +643,7 @@ Rectangle {
 
     function updatePriorityDropdown() {
         priorityDropdown.dataModel = priorityModelV5;
-        priorityDropdown.currentIndex = 1
+        priorityDropdown.currentIndex = 0
```
leonklingele: Arghh, so many magic numbers

PR #1091, opened 2018-01-23
Bootstrap daemon (requires #3165)

stoffu:
screen shot 2018-01-23 at 15 36 57

screen shot 2018-01-23 at 15 37 10

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -478,7 +478,7 @@ ApplicationWindow {
         currentWallet.pauseRefresh();
 
         appWindow.showProcessingSplash(qsTr("Waiting for daemon to start..."))
-        daemonManager.start(flags, persistentSettings.testnet, persistentSettings.blockchainDataDir);
+        daemonManager.start(flags, persistentSettings.testnet, persistentSettings.blockchainDataDir, persistentSettings.bootstrapNodeAddress);
         persistentSettings.daemonFlags = flags
     }
 
@@ -1006,6 +1006,7 @@ ApplicationWindow {
         property string blockchainDataDir: ""
         property bool useRemoteNode: false
         property string remoteNodeAddress: ""
+        property string bootstrapNodeAddress: ""
     }
 
     // Information dialog
--- a/main.qml
+++ b/main.qml
@@ -436,7 +436,7 @@ ApplicationWindow {
                 if(localNodeConnected) {
                     leftPanel.progressBar.updateProgress(walletManager.blockchainHeight(),walletManager.blockchainTargetHeight(), 0, qsTr("Remaining blocks (local node):"));
                     leftPanel.progressBar.visible = true
-                } else if (persistentSettings.startLocalNode && !startLocalNodeCancelled) {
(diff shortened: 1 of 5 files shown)


SamsungGalaxyPlayer:
I just want to make sure I understand the workflow correctly.

> What will happen if the bootstrap node is typed in but the "connect to a remote node until my own node has finished syncing" checkbox is unticked?

The resulting action is unclear in my opinion. Read my suggestion to make this better.

> What will happen if the bootstrap node is typed in AND the "connect to a remote node until my own node has finished syncing" checkbox is ticked?

This should clearly connect to the provided bootstrap node.

> What will happen if the bootstrap node is left blank in AND the "connect to a remote node until my own node has finished syncing" checkbox is ticked?

This should clearly connect to a random remote node using the discovery service.

---

In my opinion, the specific bootstrap node option should be invisible until the "connect to a remote node until my own node has finished syncing" checkbox is ticked. If the box is ticked, then the client will connect to the specified bootstrap node if provided or else run the discovery service if not provided.

stoffu:
@SamsungGalaxyPlayer 
You seem to be misunderstanding what https://github.com/monero-project/monero/pull/3165 proposes. It's admittedly similar to the remote node mode, but not quite the same.

> What will happen if the bootstrap node is typed in but the "connect to a remote node until my own node has finished syncing" checkbox is unticked?

This is exactly the intended behavior; i.e., the bootstrap address setting can be specified only when the "Start a local node" checkbox is ticked (= the "Connect to a remote node" checkbox is unticked). Ticking the "Connect to a remote node" checkbox automatically hides the bootstrap setting, as shown in the screenshot below:

screen shot 2018-01-24 at 14 27 28

Therefore, to answer your second question:

> What will happen if the bootstrap node is typed in AND the "connect to a remote node until my own node has finished syncing" checkbox is ticked?
> This should clearly connect to the provided bootstrap node.

the GUI logic design disallows this.

> In my opinion, the specific bootstrap node option should be invisible until the "connect to a remote node until my own node has finished syncing" checkbox is ticked. If the box is ticked, then the client will connect to the specified bootstrap node if provided or else run the discovery service if not provided.

The bootstrap node option is completely separate from the remote node option, and becomes visible when the "Start a local node" checkbox is ticked and becomes invisible when the "Connect to a remote node" checkbox is ticked.

Finally, the bootstrap mode proposed in https://github.com/monero-project/monero/pull/3165 doesn't implement the idea of automatically discovering bootstrapping nodes via P2P, as originally suggested in https://github.com/monero-project/monero/issues/2204, mainly because that would add much more complexity to the code. Actually I'm quite skeptical that we could implement a secure and robust mechanism to automatically choose a bootstrap daemon using whichever guessing/voting method suggested in #2204, because ultimately, the only way to know if a chain is valid or not is to download all the blocks and verify them by yourself. Any other way is only a "cheat".

As such, the `--bootstrap-daemon-address` flag introduced in #3165 forces the user to explicitly decide which node to trust. Although it's a tedious task for new users, I think this can't/shouldn't be automated.

PR #1016, opened 2017-12-12
Catalan Translation, v 0.2

edwardlow:
New catalan translation, v.0.2

PR #1099, opened 2018-01-28
Catalan translation (recovered #1016)

erciccione:
This is the translation made by @edwardlow in #1016 . i'm recovering it since i think he abandoned it. I fixed some mistakes, resolved all conflicts and updated the monero-core_cat.ts file using `lupdate`.

~~I tested this PR using #1098 because wasn't able to build without it~~


erciccione:
@luigi1111 rebased and good to go

PR #1101, opened 2018-01-29
UnsignedTransaction: Removing the Priority enum accordingly to libwalletapi update

Neozaru:
Since `UnsignedTransaction::Priority` enum has been removed from Monero Core Wallet API, removing it from *monero-gui* as well.
Conflicts with : https://github.com/monero-project/monero-gui/pull/1098

### DIFF
--- a/src/libwalletqt/UnsignedTransaction.h
+++ b/src/libwalletqt/UnsignedTransaction.h
@@ -26,13 +26,6 @@ public:
     };
     Q_ENUM(Status)
 
-    enum Priority {
-        Priority_Low    = Monero::UnsignedTransaction::Priority_Low,
-        Priority_Medium = Monero::UnsignedTransaction::Priority_Medium,
-        Priority_High   = Monero::UnsignedTransaction::Priority_High
-    };
-    Q_ENUM(Priority)
-
     Status status() const;
     QString errorString() const;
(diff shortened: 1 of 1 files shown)


src/libwalletqt/UnsignedTransaction.h
```diff
@@ -26,13 +26,6 @@ class UnsignedTransaction : public QObject
     };
     Q_ENUM(Status)
 
-    enum Priority {
-        Priority_Low    = Monero::UnsignedTransaction::Priority_Low,
-        Priority_Medium = Monero::UnsignedTransaction::Priority_Medium,
-        Priority_High   = Monero::UnsignedTransaction::Priority_High
-    };
-    Q_ENUM(Priority)
-
```
stoffu: As commented here (https://github.com/monero-project/monero/commit/a9cae0abe7b71c4a112089016cb16ee1ee12d69f#commitcomment-27179654), the same enum defined in https://github.com/monero-project/monero-gui/blob/master/src/libwalletqt/PendingTransaction.h seems also unused. Wouldn't it then make sense to also remove that, too?

stoffu: My overlooking again: `PendingTransaction::Priority` **is** used in `libwalletqt/Wallet.*`, so it can't be removed.

PR #1107, opened 2018-01-31
Change egyptian flag

erciccione:
Substitute the old squared egyptian flag with a rounded one, for consistency with the other flags


leonklingele:
Why does this increase image dimensions? All other flags are 100px^2, this one is 256px^2.

erciccione:
Flags get automatically resized, that's why i left it like i found it. Btw I agree there is no reason to use a flag bigger than 100px, it's a waste of memory. Resized

MaxXor:
Czech flag is pretty big too (324px × 324px).

erciccione:
@MaxXor true, opened #1108 to fix it

PR #1110, opened 2018-02-04
initial slovenian translation

jernejml:
Initial translation. Not all text was translated and style is not consistent.


erciccione:
Can you squash all commits about monero-core_sl.ts in just one please?

erciccione:
Looks like this PR got a bit messed up, i suggest you to remove the last two commits and then rebase using `git rebase -i HEAD~22` and squash. Come chat on `#monero-translations` or PM me on reddit for live support

Issue #1123, opened 2018-02-20
TypeError during prove payment, silent fail.


qubenix:
When trying to prove payment there is a TypeError that is displayed in the term log, but the gui fails silently. Here is the output of the term:

```
qml: getProof: Generate clicked: txid 74baa4cbfe47ea6656c25a0a91b6b0124f1c448f8424a950b68bf3955cc94d44, address 9vz3MBqiWwLbsF4K9WCSgpH3UsW4pknan2xmyGgsCBfgA1pD8jC3XpW3s4mVffyErZfWuC2s4HqZ1JJpBXB1FgNwPCHrDm3, message: 
qml: Getting payment proof: 
qml: 	txid:  74baa4cbfe47ea6656c25a0a91b6b0124f1c448f8424a950b68bf3955cc94d44 , address:  9vz3MBqiWwLbsF4K9WCSgpH3UsW4pknan2xmyGgsCBfgA1pD8jC3XpW3s4mVffyErZfWuC2s4HqZ1JJpBXB1FgNwPCHrDm3 , message:  
qrc:///main.qml:785: TypeError: Property 'startsWith' of object InProofV11XFBohNW5t1BgV64L3MKepTb84xMT8zxWguBUUnHtkLLAx3sXbp1U11bDuHoGJYoh7542NP1Rdo54AMMWmYrQa4qT3ZrLfbftEKbyVN5qPj2dNQuaA3AfUFdGi9wj2pKixPk is not a function
```

stoffu:
Hmm, I wrote that code but I don't observe that behavior (macOS 10.12.6). Is there anyone else with this error?

qubenix:
Sorry, I should have mentioned that I'm on Debian 9.

stoffu:
OK, I also observe this issue on Ubuntu 16.04.

stoffu:
#1124

qubenix:
Closed #1123 via 5db91854091e90c819ef06d9ba1d81998faff350

PR #1124, opened 2018-02-20
startsWith is unsupported for Qt < 5.8, use indexOf instead

stoffu:
> startsWith(searchString [, position ]) // ECMAScript 6: Added in Qt 5.8

http://doc.qt.io/qt-5/qtqml-javascript-functionlist.html

On some Linux systems like Ubuntu and Debian, the installed Qt version is below 5.8 where startsWith is unsupported.


qubenix:
This fixes #1123 for me on Debian 9.

Issue #1116, opened 2018-02-14
All translations need updates


erciccione:
After #1076 all translations of the GUI need to be updated. Some are already in progress, but many other need to be checked and updated before the next release. 

These are the languages that aren't in porgress and need to be updated. I pinged in this issue all contributors who helped in past with the relative language, hoping they can contribute this time as well.

- [x] Croatian
- [x] German: @leonklingele @Schnoffel
- [x] Hebrew: @jonahar
- [x] Danish: @cryptobench
- [x] Portoguese Brazilian: @fabionitto @moneroj
- [x] Portuguese
- [x] Dutch: @potatored @fridzema @ProkhorZ
- [ ] Esperanto: @Keksoj
- [ ] Indonesian: ?
- [x] Japanese: @kenshi84
- [ ] Korean: @isaacdigs
- [x] Romanian: @mandrill-pie
- [x] Russian: @cheebeez @LvMsterfild

If you start working on a translation, please communicate it in the comments (if you have it, with the link to the WIP repository), so i can thick the relative language and consider it 'in progress'. Our tracking platform, [Taiga](https://taiga.getmonero.org/project/erciccione-monero-localization/backlog), is costantly updated and a great place to check what languages are in progress and who is working on them.

If you need help/infos ask down here, [PM me on reddit](https://www.reddit.com/message/compose/?to=ErCiccione) or come chat on `#monero-translations`

ordtrogen:
Just mentioning I'm hard at work on the Swedish one. As soon as I have a PR I'll make it known and anyone'll be welcome to comment.

ProkhorZ:
OK, thanks for the reminder. I started updating the Dutch translation. It's ~100 new or edited strings.

@potatored Thanks for correcting some bad typos. If you have time to contribute translations, could you work on the Monero website?
@fridzema I see you're busy coding. Great!

netrik182:
The strings marked with ``<translation type="unfinished">`` are ready to be translated or should we expect any changes in the english text before the release?

erciccione:
@netrik182 they are ready to be translated, just take off the `type="unfinished">` part and put your translation between `<translation></translation>`

netrik182:
@erciccione ok, thanks!

ProkhorZ:
[...]

PR #1126, opened 2018-02-20
Update Hebrew Translation

milargos:
I've updated the Hebrew translation.
Now all fields should be translated. The difficulty in Hebrew is that it is a right-to-left language so in some cases it's difficult to make it look good. I did my best and it seems to be fine now. I compiled and tested myself. Any feedback is welcomed and i'll try to do my best to fix / improve.


milargos:
#1116

milargos:
#1116

Issue #199, opened 2016-11-24
General :: Special Characters


M5M400:
1) When creating a new wallet and set the wallet_path to an existing path containing special characters (in this case "c:\I Am From Österreich\Wallets\"), the path, though picked from the "open file" dialog, will be treated as non-existent and created from scratch - with encoding error:

![screenshot from 2016-11-24 18-33-19](https://cloud.githubusercontent.com/assets/22886679/20607174/72acf688-b276-11e6-8204-ba755d661315.png)

![screenshot from 2016-11-24 18-48-05](https://cloud.githubusercontent.com/assets/22886679/20607187/8f977b7e-b276-11e6-888c-25c3f8da5446.png)

The wallet is being stored in that faulty path. Everything works as expected.

2) When loading an existing wallet from a wallet_path containing special characters, the wallet cannot be opened:

![screenshot from 2016-11-24 18-39-55](https://cloud.githubusercontent.com/assets/22886679/20607220/bf6ccdcc-b276-11e6-8e36-c0dbca662c1b.png)

![screenshot from 2016-11-24 18-40-40](https://cloud.githubusercontent.com/assets/22886679/20607223/c5b841ac-b276-11e6-8bb2-d7d69ec14adc.png)

![screenshot from 2016-11-24 18-40-57](https://cloud.githubusercontent.com/assets/22886679/20607226/cbc73c2e-b276-11e6-8789-99fc16a145c1.png)

Note that the path in the registry is correct...

![screenshot from 2016-11-24 18-42-17](https://cloud.githubusercontent.com/assets/22886679/20607227/d0c34ed4-b276-11e6-9cb7-d32dac3598f5.png)

This is the unofficial win32-161123 build from jaquee, but I suspect this also applies to current master.

thanks for discovering this to /u/electricoomph on /r/monero

moneromooo-monero:
Works fine with monero-wallet-cli, so it's a GUI only thing.

moneromooo-monero:
Actually, is works also with the GUI. Must be a Windows thing.

moneromooo-monero:
Possibly related to https://github.com/monero-project/monero/issues/938

Jaqueeee:
related to https://github.com/monero-project/monero/issues/1390
temporary fix in https://github.com/monero-project/monero-core/pull/226

xmr-eric:
@M5M400 Can this issue be closed?

medusadigital:
situation right now:

[...]

PR #1141, opened 2018-02-25
UTF-8 support for filenames and paths under Windows

rbrunner7:
This PR is the second part of a duo of PR's, the Monero GUI one - the one for Monero is [PR #3313](https://github.com/monero-project/monero/pull/3313).

Those 2 PR's fix a long-standing problem with names for files and paths on Windows: Wallets on Windows are not processed properly when either the wallet file names or the names of the containing directories contain any non-ASCII characters. See the following issues: [Monero GUI issue #199](https://github.com/monero-project/monero-gui/issues/199), [Monero GUI issue #1390](https://github.com/monero-project/monero/issues/1390) (late 2016), [Monero issue #938](https://github.com/monero-project/monero/issues/938), and possibly others.

Almost all of the necessary changes are in Monero code. For details and more info about the technical solution approach see [Monero PR #3313](https://github.com/monero-project/monero/pull/3313).

As Qt is fully UTF-8 capable anyway and Wallet2 now supports UTF-8 filenames as well, it's possible to process arbitrary-named wallets anywhere in the file system with the Windows GUI wallet. Even something like 陽葵 is possible now.

It would be best to coordinate the merging of this duo of PR's: With only the Monero one (#3313) merged the GUI wallet won't build anymore because the ICU libraries now needed by Monero will be missing in the GUI wallet's build file, i.e. in `monero-wallet-gui.pro`.

On the other hand merging only this PR alone won't improve anything yet.

After this the GUI wallet EXE references **5 ICU DLL's**, which is more than until now: `libicuioXX.dll`, `libicuinXX.dll`, `libicuucXX.dll`, `libicudtXX.dll` and `libicutuXX.dll`, with `XX` being the ICU version number. Depending on the Boost version and MSYS2 version used to build, this will be either 57 or 58, e.g. `libicuio58.dll`.

Of course this must be considered when packaging the GUI wallet. (There is no PR yet that would adjust that for the InnoSetup Windows installer script.)

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -218,6 +218,13 @@ win32 {
         -lboost_regex-mt-s \
         -lboost_chrono-mt-s \
         -lboost_program_options-mt-s \
+        -lboost_locale-mt-s \
+        -licuio \
+        -licuin \
+        -licuuc \
+        -licudt \
+        -licutu \
+        -liconv \
         -lssl \
         -lcrypto \
         -Wl,-Bdynamic \
--- a/wizard/WizardMain.qml
+++ b/wizard/WizardMain.qml
@@ -185,14 +185,6 @@ ColumnLayout {
             return false;
         }
 
-        // Don't allow non ascii characters in path on windows platforms until supported by Wallet2
-        if (isWindows) {
-            if (!isAscii(path)) {
-                walletErrorDialog.text = qsTr("Non-ASCII characters are not allowed in wallet path or account name")  + translationManager.emptyString;
-                walletErrorDialog.open();
-                return false;
-            }
-        }
         return true;
     }
 


EvertonMelo:
@Jaqueeee  

You create the struture correct? 

You can help us?

rbrunner7:
I just amended this PR, with changes in `monero-wallet-gui.pro`, to track the recent change in filenames for ICU libraries - they dropped the letter "s" in their filenames. Details see the corresponding new Monero [PR #3375](https://github.com/monero-project/monero/pull/3375).

rbrunner7:
@EvertonMelo: Hmm, I am confused now. First, this Monero GUI PR here is not yet merged, and second, the changes in the corresponding Monero PR are only in code guarded by #ifdef macro statements to only compile on Windows. No idea how it would even be possible to break an Ubuntu build with that?

How did you connect the broken build with this PR or with the corresponding PR for Monero proper?

stoffu:
The build failure on Ubuntu AMD64 is due to the OpenSSL library needing to be rebuilt with -fPIC:

```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libssl.a(s23_meth.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
```

Look like some recent merged to the Monero repo caused this.

Regarding the MSYS2/MinGW static ICU library file renaming, as in the other PR (https://github.com/monero-project/monero/pull/3375#issuecomment-372513252), I'm quite concerned that the users are forced to upgrade their MSYS2 installation. Can't we just get away without `sicuio sicuin sicuuc sicudt sicutu`?

stoffu:
I've confirmed that the GUI builds fine on Windows without linking to ICU, see #1177

stoffu:
I'm convinced in https://github.com/monero-project/monero/pull/3375 that these libraries are indeed necessary.

PR #1130, opened 2018-02-20
Dutch translation of GUI text added in January

ProkhorZ:
Plus a few corrections of older translations.

@fridzema already submitted another Dutch translation, so you're able to choose!
Who wants to review both translations and choose the best one, or suggest edits?
@rpcjacobs @potatored @dnaleor @dEBRUYNE-1

PR #1147, opened 2018-03-03
Updating Dutch Translations

Kortom-Tom:
Translations. (check if it does a good compile)


erciccione:
Hi @Kortom-Tom Thanks for this but looks like this PR is a duplicate of #1130 . Could you close this PR and propose your changes there?

Issue #1131, opened 2018-02-20
Esperanto seed option not used in GUI


ProkhorZ:
In languages.xml, the wallet_language parameter for Esperanto is still set to English, although there is a mnemonic seed word list in Esperanto now. I assume this parameter refers to the mnemonic seed language, since it's localized for other languages like Russian and Japanese that have localized seeds but not a localized CLI and daemon.

`        <language display_name="Esperanto" locale="eo" wallet_language="English" flag="/lang/flags/esperanto.png" qs="none"/>
`
in https://github.com/monero-project/monero-gui/blob/master/lang/languages.xml

I don't feel comfortable changing this myself in an important file. I'm not sure if anything else needs to be changed in order to enable selecting Esperanto. But I would like to see Moneristoj actually use the word list I compiled.

glv2:
There is no other line to change to use the Esperanto word list when the Esperanto locale is used.

So you just have to make a pull request changing the line
```<language display_name="Esperanto" locale="eo" wallet_language="English" flag="/lang/flags/esperanto.png" qs="none"/>```
to
```<language display_name="Esperanto" locale="eo" wallet_language="Esperanto" flag="/lang/flags/esperanto.png" qs="none"/>```
and new wallet generation will use the right word list.

ProkhorZ:
Thanks, that's surprisingly easy! 

Complication: I accidentally made another commit in my forked master branch. I'll make a pull request for this issue after sorting that out.

ProkhorZ:
PR #1154 merged.

PR #1154, opened 2018-03-04
Use Esperanto mnemonic seed for Esperanto GUI

ProkhorZ:
See issue #1131.


stoffu:
Looks like a lot of changes got reverted. Can you please squash commits?

moneromooo-monero:
The commits needs squashing. Looks like something like "git rebase HEAD~3" will work (and replace "pick" with "f" on the second and third lines).

ProkhorZ:
OK, wait a minute, please. I happen to have some tech support today.

ProkhorZ:
OK, I rebased with HEAD~3 as moo said, and I dropped the translation commit that got mixed up in here, as well as the reverting commit. Bit of a detour, sorry about that, but I learned to use git in a bash shell.

PR #1161, opened 2018-03-05
Stagenet

stoffu:


LeftPanel.qml
```diff
@@ -104,8 +104,8 @@ Rectangle {
 
         Text {
             id: testnetLabel
-            visible: persistentSettings.testnet
-            text: qsTr("Testnet") + translationManager.emptyString
+            visible: persistentSettings.testnet || persistentSettings.stagenet
```
moneromooo-monero: I think it'd be better to have an enum like in monero, rather than N booleans.

stoffu: I did so mostly because I wasn't sure about how to handle enums in QML. But it turns out that using an enum is indeed a much better approach. Thanks.

PR #1175, opened 2018-03-07
Disable Mining page when connected to remote node

pazos:
Mining is only available for local daemons, so disable left panel mining menu when connected to remote daemon. This PR also removes some warnings in mining page since those aren't reached anymore.

I'm not sure if I need to touch any other thing, I tested this in Ubuntu and Windows.

Feedback is welcome.

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -409,7 +409,7 @@ Rectangle {
             // ------------- Mining tab ---------------
             MenuButton {
                 id: miningButton
-                visible: !isAndroid && !isIOS
+                visible: !isAndroid && !isIOS && !persistentSettings.useRemoteNode
                 anchors.left: parent.left
                 anchors.right: parent.right
                 text: qsTr("Mining") + translationManager.emptyString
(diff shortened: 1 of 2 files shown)


moneromooo-monero:
Isn't it better to tell the user why it's not available, rather than unexpectedly missing ?

pazos:
@moneromooo-monero: good point. I guess most users who want to run the GUI against a remote node won't mind that missing feature. From a UI perspective I rather prefer to keep it simple and don't display non available options.

As far as user documentation this doesn't require changes in the  [guide to solo mine with the GUI](https://getmonero.org/resources/user-guides/solo_mine_GUI.html).

Of course I'm speaking from my own (user) perspective. I will be happy to close this PR if isn't useful.

luigi1111:
@pazos Please rebase.

I sorta agree with @moneromooo-monero but am unsure how much extra work that is and what it would look like. I'm also ok with merging as-is, pending further discussion.

moneromooo-monero:
Well, it already is disabled when using a remote node IIRC.

pazos:
Yeah, better to close this PR and keep things as is.

Thanks for your time

Issue #387, opened 2017-01-06
Suggestion for block sync progress bar


voidzero:
Right now, the progress bar is at almost 100% every time I start up the GUI because it's technically correct - right now I am at block 1217730 and need to sync to 1218152.

My idea is to make the progress bar instance specific. So when the GUI is started, the current block (1217730) could be 0% or 1%. That makes the progress bar a heck of a lot nicer.

thoriumbr:
I think showing `Updating block 4 of 175` is more user friendly than `Updating block 1217730 of 1218152.`

voidzero:
Ooh, yes, I like that one too.

PR #1182, opened 2018-03-15
Two progress bars

Jaqueeee:
Separate progress bars for daemon and wallet

Also removed the old "sync local daemon while using a remote node" logic which has become obsolete after #1091

### DIFF
--- a/components/ProgressBar.qml
+++ b/components/ProgressBar.qml
@@ -32,7 +32,8 @@ import moneroComponents.Wallet 1.0
 Rectangle {
     id: item
     property int fillLevel: 0
-    visible: false
+    property string syncType // Wallet or Daemon
+    property string syncText: qsTr("%1 blocks remaining: ").arg(syncType)
     color: "#1C1C1C"
 
     function updateProgress(currentBlock,targetBlock, blocksToSync, statusTxt){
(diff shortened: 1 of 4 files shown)


Jaqueeee:
skarmavbild 2018-03-15 kl 16 59 47
skarmavbild 2018-03-15 kl 17 01 18

components/ProgressBar.qml
```diff
@@ -44,24 +45,19 @@ Rectangle {
...
         if(targetBlock > 0) {
-            var remaining = targetBlock - currentBlock
-            // wallet sync
-            if(blocksToSync > 0)
-                var progressLevel = (100*(blocksToSync - remaining)/blocksToSync).toFixed(0);
-            // Daemon sync
-            else
-                var progressLevel = (100*(currentBlock/targetBlock)).toFixed(0);
+            var remaining = (currentBlock < targetBlock) ? targetBlock - currentBlock : 0
+            var progressLevel = (blocksToSync > 0) ? (100*(blocksToSync - remaining)/blocksToSync).toFixed(0) : 100
```
moneromooo-monero: In these two lines, the difference between targetBlock and blocksToSync isnt't clear. For the progress level to match remaining, they shoiuld be the same, no (plus or minus one) ?

Jaqueeee: It's related to #387. If the difference between currentBlock and targetBlock is small, the progress bar would be close to 100% filled during the whole sync progress. 

blocksToSync is set at wallet startup and the progress bar indicated the sync progress from 0% to 100% in the current session. If you restart the wallet, the bar restarts from 0%.

moneromooo-monero: OK, that makes sense.

skftn: I have a bug where the daemon progress is always 100%. Possibly happens when `blocksToSync` is undefined, see how its used in `main.qml`.

PR #1172, opened 2018-03-07
Add a shared ringdb management page - rings and blackballed outputs

moneromooo-monero:

PR #1187, opened 2018-03-18
pro: add lmdb to link line for Linux

radfish:
Otherwise `undefined symbol mdb_tx_abort` and others.
Possibly same is needed for other platforms. I don't have them, so didn't touch the other lists.


stoffu:
This can be closed now that #1172 was merged.

PR #1190, opened 2018-03-19
Additional DLLs in 'windeploy_helper.sh': ICU, stack protection, regexp

rbrunner7:
The GUI exe for Windows, built with the latest sources, and with a MSYS2 build environment updated to latest versions of all packages, needs some more DLLs than until now: `libssp-0.dll` (for the recently-introduced stack protection), `libpcre2-16-0.dll` (something about regular expressions?), and 5 ICU libraries (a direct consequence of my own [PR #1141](https://github.com/monero-project/monero-gui/pull/1141).

I updated `windeploy_helper.sh` to copy all those files.

Please note: If with these changes the script does not find the ICU libraries your MSYS2 environment is not updated to the latest version of ICU. Background info about the library rename in question is [here](https://github.com/monero-project/monero/pull/3375).

### DIFF
--- a/windeploy_helper.sh
+++ b/windeploy_helper.sh
@@ -17,13 +17,13 @@ fi
 
 if [[ "$BUILD_TYPE" == "Release" ]]; then
 	echo "Release build"
-	ICU_FILES=(libicuuc57.dll libicuin57.dll libicudt57.dll)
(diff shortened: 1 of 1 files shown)


luigi1111:
iDunk approved.

PR #1192, opened 2018-03-20
recurse git submodules

danrmiller:
This is so we get the unbound submodule of the monero submodule.

I'm not real familiar with git submodules or these gui build scripts, so please take a look.


danrmiller:
going to try again

PR #1195, opened 2018-03-22
Windows Installer: Updates for Lithium Luna i.e. 0.12.0.0

rbrunner7:
I updated the Windows installer to Lithium Luna.

Other than for earlier updates I did not yet have the final, official 0.12.0.0 Monero GUI Wallet package from getmonero.org with the "authoritative" file list available, so there may still be one or the other error in the files listed in the InnoSetup script to install.

Some of the files and directories also depend on the exact version of Qt used to build and package Monero. I used whatever version the MSYS2 package `mingw-w64-x86_64-qt-creator` found on a Qt website [here](https://wiki.qt.io/MinGW-64-bit) currently installs.

But any needed corrections to this installer will be probably small.


installers/windows/Monero.iss
```diff
@@ -39,7 +39,7 @@ Name: "en"; MessagesFile: "compiler:Default.isl"
 ; .exe/.dll file possibly with version info).
 ;
 ; This is far more robust than relying on version info or on file dates (flag "comparetimestamp").
-; As of version 0.11.1.0, the Monero .exe files do not carry version info anyway in their .exe headers.
+; As of version 0.12.0.0, the Monero .exe files do not carry version info anyway in their .exe headers.
```
moneromooo-monero: Is that right ?

installers/windows/ReadMe.htm
```diff
@@ -1,13 +1,13 @@
...
+<title>Monero Lithium Luna GUI Wallet</title>
 </head>
 
 <body style="font-family: Arial, Helvetica, sans-serif">
-<h1>Monero Helium Hydra GUI Wallet</h1>
+<h1>Monero Lithium Luna GUI Wallet</h1>
 
   <p>Copyright (c) 2014-2018, The Monero Project<br>
-  Date: September 19, 2017</p>
+  Date: March 18, 2018</p>
```
moneromooo-monero: What is that for ? If it's a copyright date, then it's wrong in the first place.

installers/windows/Monero.iss
rbrunner7: Yes. To be sure I just checked the latest buildbot build: No version number in the sense of Windows in there, nor any application name or copyright that you could put into the .exe header as well.

I don't know how one would have to modify the make file to make MSYS2 `ld` put the current version number into the Windows .exe header.

installers/windows/ReadMe.htm
rbrunner7: That was my idea of putting something like a "Written on this and this date" into the file, as an information for the reader. Motivated by things like articles I sometimes find on the Internet with no such date and thus impossible to judge.

If that's no good idea I zap it, no problem.

luigi1111: Seems ok to me.

PR #1200, opened 2018-03-28
Embed 'DejaVu Sans' font for portability

stoffu:
I'm not sure if this is the right thing to do, but currently I can't build a working statically-linked GUI for Linux where Qt and Boost etc needed to be rebuilt with -fPIC, where I get this error:

> QFontDatabase: Cannot find font directory /home/Qt_Static/lib/fonts.
> Note that Qt no longer ships fonts. Deploy some (from http://dejavu-fonts.org for example) or switch to fontconfig.

This will increase the size of the executable by about 2MB, which is not very nice. Also, I didn't check the license-wise feasibility. Suggestions?


qml.qrc
```diff
@@ -162,5 +162,9 @@
         <file>pages/Keys.qml</file>
         <file>images/menu.png</file>
         <file>images/appicon.ico</file>
+        <file>fonts/DejaVuSans-Bold.ttf</file>
+        <file>fonts/DejaVuSans-BoldOblique.ttf</file>
+        <file>fonts/DejaVuSans-Oblique.ttf</file>
+        <file>fonts/DejaVuSans.ttf</file>
```
pazos: I see some fonts already embedded in qml.qrc:

```
        <file>fonts/SFUIDisplay-Medium.otf</file>
        <file>fonts/SFUIDisplay-Regular.otf</file>
        <file>fonts/SFUIDisplay-Light.otf</file>
        <file>fonts/SFUIDisplay-Bold.otf</file>
```
Could those files be repurposed?

stoffu: Oh, those were added recently in #1206. Closing

PR #1205, opened 2018-03-29
link to boost lib filenames from msys2 package

danrmiller:
Don't merge this unless you are prepared to break the old behavior. 

I'm told that people manually building boost libs on windows have the filenames ending in -s. 

This change makes the gui link if using the default boost libs from msys2. 

It might be more ideal to allow either names, if someone wants to do that or tell me how.


luigi1111:
This way seems better, however if someone wants to figure out how to support both, please go ahead. :)

PR #1231, opened 2018-03-31
Add libpcsclite to link line if libwallet_merged is linked against it

iDunk5400:
This fixes unresolved dependencies when linking monero-wallet-gui if libwallet_merged was built with HW device support.


iDunk5400:
That didn't work according to the buildbot. Closing.

PR #1234, opened 2018-03-31
Adds warning message for when a transaction with a non-default mixin …

skftn:
…value is created

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -612,15 +612,16 @@ ApplicationWindow {
             // here we show confirmation popup;
             transactionConfirmationPopup.title = qsTr("Please confirm transaction:\n") + translationManager.emptyString;
             transactionConfirmationPopup.text = "";
-            transactionConfirmationPopup.text +=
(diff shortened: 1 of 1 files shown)


main.qml
```diff
@@ -608,18 +608,23 @@ ApplicationWindow {
         } else {
             console.log("Transaction created, amount: " + walletManager.displayAmount(transaction.amount)
                     + ", fee: " + walletManager.displayAmount(transaction.fee));
-
+            
+
```
moneromooo-monero: Random spurious whitespace change

components/StandardDialog.qml
```diff
@@ -100,6 +100,7 @@ Rectangle {
         RowLayout {
             id: column
             //anchors {fill: parent; margins: 16 }
+            Layout.topMargin: 14
```
moneromooo-monero: Seems unrelated

skftn: Fixed.

main.qml
skftn: Fixed.

Issue #1229, opened 2018-03-31
 syncing Daemon with GUI most of the times GUI gives a false positive "Connected Progress Bar"


juanpc2018:
When syncing the Daemon most of the times the GUI gives a false positive "Connected Progress Bar",
The true way to know if it's truly synced is:
A) press start mining.......
or
B) task manager hard drive activity....

A way to detect HardDrive activity like task manager should combined to give a proper reading.
NOT + AND Gate Logic.
also detecting the true last block independently from another source + Comparator....

skftn:
Are you saying the progress bars do not reflect that monerod is syncing/refreshing?

qubenix:
I'm not sure if this is what OP means (if I am wrong then I'll make an issue for this specifically), but I've experienced that the progress bar is either empty or full. Here you can see my blockchain is about 50% sync'd, but progress bar shows empty:

![image](https://user-images.githubusercontent.com/15707061/38314243-fc0cf0f2-37e2-11e8-9ad2-1b08e64a7197.png)

Here, just a little while later in the sync process, my bar shows full:

![image](https://user-images.githubusercontent.com/15707061/38314286-0fe350c6-37e3-11e8-9591-3c6ff7fa242f.png)

@juanpc2018 please confirm if this is what you are experiencing as well.

juanpc2018:
in Windows8.1x64 shows Connected, but still is Syncing,
GUI Freezes, but resumes Ok.

Shows Blocks Remaining, Shows Connected, but still Syncing, GUI Freezes, but resumes OK,
until is fully connected... Hard Drive does not Move "monerod.exe",
Start Mining Works Ok. etc...

stoffu:
@qubenix This is fixed in #1243.

qubenix:
Sorry to almost hijack this issue because it seems @juanpc2018 is having some freezing issue instead. My comment is fixed with pr #1243 so I don't think there's a need to open a separate issue.

PR #1243, opened 2018-04-03
ProgressBar: fix conversion to percentage

stoffu:
Previously, `progressLevel` was always set to zero when `currentBlock / targetBlock < 0.5` because `100*(currentBlock / targetBlock).toFixed(0)` is evaluated as `100*((currentBlock / targetBlock).toFixed(0)) = 100*0 = 0`.


qubenix:
This indeed fixes the issue in my comment on #1229.

PR #1258, opened 2018-04-04
Updated README.md with new dependencies for Ubuntu 16.04+:

1337tester:
qml-module-qtquick-controls2
    qml-module-qt-labs-folderlistmodel

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Packaging for your favorite distribution would be a welcome contribution!
 
   - For Ubuntu 16.04 x86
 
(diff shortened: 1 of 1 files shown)


dEBRUYNE-1:
I just noticed only x86 instructions were upgraded. Did you forget to update the instructions for x64?

1337tester:
Actually I was only focused on Ubuntu 16.04+ x64, I don't test/validate for other platforms -> might be that I put it even in the wrong place, should be 16.04+

PR #1266, opened 2018-04-05
WalletManager::isDaemonLocal: return false if string is empty

stoffu:


stoffu:
I realized that this patch doesn't really add much, because the case of empty input to `isDaemonLocal(str)` is just the following warning message which is no harm at all:

    2018-04-05 09:31:44.977	    7f3535d66800	WARN 	net.http	src/common/util.cpp:627	Failed to determine whether address '' is local, assuming not

Closing

PR #1286, opened 2018-04-07
Fix for frozen Receive page (requires #3576)

stoffu:


keatond:
Confirming that I have tested this update and having the tracking turned off by default fixed the issue. When selecting the radio button to turn on tracking the lag re-occurs. If I leave the tracking on and go to other tabs the lag continues. If i go back to the receive tab the lag will stop and the tracking button is unchecked again.

keatond:
Will this be implemented on version 0.12.2.0?

erciccione:
@keatond yes

keatond:
thanks! you guys rock!

PR #1287, opened 2018-04-07
Settings: use checkbox instead of button for daemon mode switching

stoffu:
IMO the current way of using grayed disabled buttons to indicate the selected mode is a bit unintuitive. 

screen shot 2018-04-07 at 14 23 57


1337tester:
If we wish to go full-intuitive mode here, I would prefer radio-buttons, they are most suitable to indicate a single choice among many (check-boxes can mean also multiple choice)

Issue #1288, opened 2018-04-07
GUI graphic overlay


ElectricSlacks:
![monero bitchthing](https://user-images.githubusercontent.com/35742210/38453881-48b9546e-3a5d-11e8-9f2a-3627bf70e0f4.JPG)

Hello,
As you can see an opened ''Advanced' tab creates an overlay in the gui.
I am sure that this is easily fixed.
And its a small thing. no worries.

pazos:
Also the flickable region should be dynamic based on the size of the menu.

I'm in for this issue

jamescanningcooke:
Same prob

skftn:
Fixed via #1289 #1290

+resolved

PR #1289, opened 2018-04-07
[Release]: Leftpanel overlay fix

pazos:
Temporary fix to #1288 that works in the new black theme. Since the leftpanel menu doesn't have an explicit height and the status elements are on top of the menu I changed two things:

1. statusbar elements (networkStatus and progressBars) have black background
2. made the flickable region dynamic to let us scroll to the menu items hidden by statusbar

sin nombre

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -252,7 +252,9 @@ Rectangle {
 
         Flickable {
             id:flicker
-            contentHeight: 500 * scaleRatio
+            contentHeight: (progressBar.visible)? menuColumn.height + separator.height + 
+                networkStatus.height + progressBar.height + daemonProgressBar.height : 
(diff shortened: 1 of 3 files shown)


LeftPanel.qml
```diff
@@ -568,23 +570,34 @@ Rectangle {
...
         } // Flickable
 
+        Rectangle {
+            id: separator
+            anchors.left: parent.left
+            anchors.right: parent.right
+            anchors.leftMargin: 0
+            anchors.rightMargin: 0
+            anchors.bottom: networkStatus.top;
+            height: 10
```
skftn: `height: 10 * scaleRatio`

skftn:
Nice.

PR #1298, opened 2018-04-09
[DO NOT MERGE] Workaround for std::bad_cast issue

stoffu:


src/libwalletqt/WalletManager.cpp
```diff
@@ -272,7 +272,15 @@ bool WalletManager::localDaemonSynced() const
 
 bool WalletManager::isDaemonLocal(const QString &daemon_address) const
 {
-    return Monero::Utils::isAddressLocal(daemon_address.toStdString());
+    // TODO: workaround for std::bad_cast thrown by boost::basic_regex ctor due to std::collate<char> missing in some Linux, see https://svn.boost.org/trac10/ticket/4671
+    // reverting https://github.com/monero-project/monero-gui/pull/1015
+    if (daemon_address == "")
+        return false;
+    QString daemon_host = daemon_address.split(":")[0];
+    if (daemon_host == "127.0.0.1" || daemon_host == "localhost")
```
skftn: Ideally you'd cover `127.0.0.0/8` and maybe ipv6 `::1` too but I understand this is a quick workaround. Nice one.

stoffu: I'm not so familiar with this kind of network routing stuff, but a Wikipedia article says (https://en.wikipedia.org/wiki/Localhost)

> That means any packet sent to any of those addresses (127.0.0.1 through 127.255.255.255) is looped back. The address 127.0.0.1 is the standard address normally used for IPv4 loopback traffic; the rest are rarely used in practice. The IPv6 standard assigns only a single address for loopback: ::1.

and since we're not using IPv6 currently, I guess it's OK for now?

BTW, the corresponding change was moved from here into monero/src/wallet/api/utils.cpp in https://github.com/monero-project/monero/pull/3590

PR #1299, opened 2018-04-09
Name of Turkish language and list of languages

ProkhorZ:
1. Change 'Turkish' to 'Türkçe' in the opening screen. I don't speak Turkish, but the language name is pretty obvious.
2. Change wallet_language for Turkish to English, because there's actually not a mnemonics word list in Turkish at the moment, although it's listed here in the comment above.
3. Update that list of available seed languages. Source: https://github.com/monero-project/monero/blob/master/src/mnemonics/electrum-words.cpp
4. Change wallet_language for Catalan to 'English', because there's no Catalan word list. Personally, I think Spanish as the seed language would be more useful to speakers of Catalan, but that's a political decision which requires discussion.

### DIFF
--- a/lang/languages.xml
+++ b/lang/languages.xml
@@ -2,18 +2,20 @@
 <languages>
 
 <!--
-List of available languages for your wallet's seed:
-0 : Deutsch         (German)
-1 : English
-2 : Español         (Spanish)
-3 : Français        (French)
-4 : Italiano        (Italian)
-5 : Nederlands      (Dutch)
-6 : Português       (Portuguese)
-7 : русский язык    (Russian)
-8 : 日本語          (Japanese)
-9 : 简体中文 (中国) (Simplified Chinese (Mainland))
-11 : Turkey         (Turkish)
+Note that wallet_language is not the wallet language but the mnemonic seed language!
+List of languages for which a mnemonic seed word list is available:
+English
+Nederlands	(Dutch)
+Français	(French)
+Español		(Spanish)
+Português	(Portuguese)
+日本語          (Japanese)
+Italiano	(Italian)
+Deutsch		(German)
+русский язык	(Russian)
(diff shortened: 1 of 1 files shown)


erciccione:
>Spanish as the seed language would be more useful to speakers of Catalan

This is a discussion we should avoid like plague, especially during these times. Let's keep it in English

ProkhorZ:
> This is a discussion we should avoid like plague, especially during these times. Let's keep it in English

Oh, you don't like the can of worms? It's full of protein! :)

Maybe it would be better if we didn't choose the seed language for the user at all. We're not Apple who know better than the users what they need. There are so many scenarios possible where someone would want to use another language for the seed than for the GUI. Writing down a list of words is a different skill than using a GUI.

luigi1111:
@ProkhorZ please rebase.

ProkhorZ:
Hi @luigi1111, I updated the master branch in my fork, but git asked me to resolve the merge conflict before rebasing. Isn't that up to you? What do I need to do exactly?

erciccione:
@ProkhorZ now check the `lang/languages.xml` file, you'll find some strings marked `<<<<<<<`, `=======`, `>>>>>>>`. Follow instructions [here](https://help.github.com/articles/resolving-a-merge-conflict-on-github/) to resolve, than you will be able to rebase and push (using `-f`. so it will be `git push -f origin languages)

ProkhorZ:
OK, I resolved the conflict on Github according to your link, taking the line for Serbian from the master branch, but the problem is that I can't rebase on Github. 
I still get the same error locally in Git, which refers to files I can't find. Editing lang/languages.xml counts as a new commit.

rebase conflict after solution in github

erciccione:
That because you have fixed the conflicts on github but not on your local branch. Can you add me as contributor to your repo? i will try to fix it for you

ProkhorZ:
Thanks!

PR #1305, opened 2018-04-10
ReadMe: Updated Windows/MSYS2 compile instructions

rbrunner7:
Finally 64 bits and 64 bits only, Boost from MSYS2 package, Qt5 from MSYS2 package, as per release Lithium Luna

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,57 +194,47 @@ More info: http://stackoverflow.com/a/35098040/1683164
 
 ### On Windows:
 
-1. Install [msys2](http://msys2.github.io/), follow the instructions on that page on how to update packages to the latest versions
+The Monero GUI on Windows is 64 bits only; 32-bit Windows GUI builds are not officially supported anymore.
 
-2. Install Monero dependencies as described in [monero documentation](https://github.com/monero-project/monero) into msys2 environment
-   **As we only build application for x86, install only dependencies for x86 architecture (i686 in package name)**
-   ```
-   pacman -S mingw-w64-i686-toolchain make mingw-w64-i686-cmake mingw-w64-i686-boost mingw-w64-i686-openssl mingw-w64-i686-zeromq mingw-w64-i686-libsodium
+1. Install [MSYS2](https://www.msys2.org/), follow the instructions on that page on how to update system and packages to the latest versions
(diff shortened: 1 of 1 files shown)


stoffu:
Does this produce a portable binary that can run on any vanilla Windows machines?

rbrunner7:
Well, I don't know, I am no Qt and Boost specialist. I have read about the problems or assumed problems with the Windows binaries on older computers, on computer accessed over RDP, etc. that you try to chase down now post-release, and I can't tell how they are related or not to the use of the official MSYS2 packages for Qt and Boost.

I can just tell from my experience that the produced binaries are certainly "usable"; building is successful, and I have personally used binaries produced in the described way on two Windows 7 and two Windows 10 systems without any problems.

pazos:
This produces portable builds which run in in every desktop/laptop that supports opengl 2.0. Since we're using mingw-qt, which was built with -opengl desktop, there isn't a way to change this at runtime. This is ok for most self-builds and the method used in "low graphics mode" still works (qtquick will work without opengl capable device)

For official releases we want a version of qt built with -opengl dynamic, which lets the software try different opengl implementations in this order: opengl ES -> opengl via directx -> opengl via llvmpipe.

This is done via official releases for windows (since qt5.4) or building from source. Official releases are x86 only, so building from sources should be the right step for redistributable builds.

gene-telligent:
I'd love to at least get these changes in ASAP (even if it doesn't result in a fully portable build), as when I was setting up my Windows development environment initially I spent a good day going down the compiling-for-x86-on-64bit rabbithole before discovering that it's now obsolete. Having explicitly wrong instructions in the README isn't good for anyone!

pazos:
@stoffu: of course. I'm not sure how to do that. 

Maybe I can push a qt build script under contrib/qt and a README.md on contrib?. The qt library would target main desktop oses (win,mac,linux) . It won't target mobile devices because hell.

Qt should be enlaced statically only for linux/bsd relying only on basic system libraries, using as much as qt bundled libraries as possible.

For windows/mac qt libraries & plugins can be bundled with the app 

For windows I would need some help
And in general it would take me some time :p

PR #1313, opened 2018-04-13
[qmake]: add versioning support

pazos:
In order to make the versioning easier and keep a tidy repo I'd like to move versioning related stuff to qmake. 

This PR:

1. moves version information from build.sh/version.js to version.pri
2. adds version to the application for windows/mac
3. removes old monero-core.rc (qmake will generate one for us)
4. makes versions accesible in C++ (VERSION_GUI and VERSION_CORE) and qml
5. supports corner cases (git isn't found or repo not initializated)

version.pri obtains version from git. The first two lines must be changed every major release. (e.g: "Helium Hydra" -> "Lithium Luna" / 0.11.x -> 0.12.x).

Right now we're using the get_tag function from utils.sh to obtain git version and display useful? info at build time. I deleted the call to get_tag in build.sh, but I can revert that if you wish.

Feedback is welcome, specially about git tagging, because I'm not sure if I'm doing this right :package:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 *.user.*
 translations/*.qm
 build
-version.js
 # IOS stuff below
 moc_*
 *.o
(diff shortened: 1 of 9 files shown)


pazos:
closed in favor of #1340

PR #1316, opened 2018-04-13
[Don't merge]: send output to easylogger

pazos:
install messageHandler after QApplication constructor, so we can see certain error messages returned by qt.

the messageHandler was merged in #463, but I'm not sure why it was removed. Pinging @Jaqueeee


pazos:
closed in favor of #1340

PR #1320, opened 2018-04-15
single colour privacy level bar

cryptochangements34:
Making the privacy level bar go from orange, to light green, to dark green gives the impression (especially to less knowledgable users) that a higher ring size equals better privacy, which isn't always the case. To avoid giving that impression this PR makes the privacy level bar solid orange, regardless of selected ring size, and changes the wording of "privacy level (ring size x)" to just "ring size"

[screenshot](https://imgur.com/AvSir0i)

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -79,7 +79,7 @@ Rectangle {
         var mixin = scaleValueToMixinCount(fillLevel)
         console.log("PrivacyLevel changed:"  + fillLevel)
         console.log("mixin count: "  + mixin)
-        privacyLabel.text = qsTr("Privacy level (ringsize %1)").arg(mixin+1) + translationManager.emptyString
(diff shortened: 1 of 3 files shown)


pages/Transfer.qml
```diff
@@ -435,7 +434,7 @@ Rectangle {
             Label {
                 id: privacyLabel
                 fontSize: 15
-                text: ""
+                text: qsTr("Ring size") + translationManager.emptyString
```
leonklingele: Hiding the actual ring size being used is probably not what we want to do, and might be harmful in some cases.
I suggest to rephrase the wording to: `qsTr("Ring size: %1).arg(mixin + 1)`

cryptochangements34: You're right idk how I missed this, thanks

Issue #1317, opened 2018-04-14
Typos


skftn:
From [reddit](https://www.reddit.com/r/Monero/comments/8c5bwz/images_of_a_couple_typos_bugs_in_the_v01200_gui/)

> Hi I found a couple typo style bugs in the new GUI wizard, when opened and a new wallet is setup:
Typo 1:

![](https://www.anonfiles.cc/uploads/59701ab6df410142c9d4946fb2403c2a.jpeg)

> Bugish typo 2: Even though I filled in a remote node, the final setup review screen shows that I'm using a local node:

![](https://www.anonfiles.cc/uploads/b0664ee314f7a44e452565d3c0fa7323.jpeg)

> I can confirm that when I hit "Use Monero" and the actual wallet opens, it is configured correctly (using the remote node I filled in).
> Can someone get this info to the proper people?

skftn:
+easy

skftn:
+help wanted

skftn:
+resolved

PR #1321, opened 2018-04-15
show correct address in wizard if remote node used

cryptochangements34:
Fixes the second part of #1317


MaxXor:
Hm, why do we even use two different settings for the daemon address (`daemon_address` & `remoteNodeAddress`) when it can be merged into one with [a setting](https://github.com/monero-project/monero-gui/blob/e4d75c6680bb3ed7bccdf00e8e319b25b6a7be21/src/libwalletqt/WalletManager.cpp#L273) which tells if it's remote or local node?

skftn:
What exactly are you proposing? I'm not sure if we are on the same page.

For context: `grep -ri --include="*.qml" -E 'daemon_address|remoteNodeAddress' .`

I was thinking of migrating all connection handling to it's own javascript class, used as a singleton throughout the wizards and main application. This would not be a small refactor.

cryptochangements34:
@MaxXor I was wondering myself what the purpose of having these two was, but as @skftn pointed out this system is used pretty much all throughout the code base and wouldn't be a small change

PR #1323, opened 2018-04-15
fix OpenAlias resolve

cryptochangements34:
The way the labelbutton was being used was stopping the "resolve" button from showing up when an OpenAlias address was used. Once you got it to show up it would end up replacing the "Address Book" label so this PR changes it to it's own standard button. The regex filter also has to allow '.' and '@' characters otherwise the user can't even enter a proper OpenAlias address

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -257,40 +257,8 @@ Rectangle {
               labelText: qsTr("<style type='text/css'>a {text-decoration: none; color: #858585; font-size: 14px;}</style>\
                 Address <font size='2'>  ( </font> <a href='#'>Address book</a><font size='2'> )</font>")
                 + translationManager.emptyString
-              labelButtonText: qsTr("Resolve") + translationManager.emptyString
               placeholderText: "4.."
               onInputLabelLinkActivated: { appWindow.showPageRequest("AddressBook") }
-              onLabelButtonClicked: {
-                  var result = walletManager.resolveOpenAlias(addressLine.text)
-                  if (result) {
-                    var parts = result.split("|")
-                    if (parts.length == 2) {
-                      var address_ok = walletManager.addressValid(parts[1], appWindow.persistentSettings.nettype)
-                      if (parts[0] === "true") {
-                        if (address_ok) {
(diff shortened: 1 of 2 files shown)


pages/Transfer.qml
```diff
@@ -306,6 +274,54 @@ Rectangle {
...
+              onClicked: {
+                  var result = walletManager.resolveOpenAlias(addressLine.text)
+                  if (result) {
+                      var parts = result.split("|")
+                      if (parts.length == 2) {
+                          var address_ok = walletManager.addressValid(parts[1], appWindow.persistentSettings.testnet)
+                          if (parts[0] === "true") {
+                              if (address_ok) {
+                                  addressLine.text = parts[1]
+                                  addressLine.cursorPosition = 0
```
stoffu: Looks like the property `cursorPosition` was removed when `addressLine` switched from `LineEdit` to `LineEditMulti` in the "dark theme" PR.

pages/Transfer.qml
```diff
@@ -306,6 +274,54 @@ Rectangle {
...
+              width: 80
+              text: qsTr("Resolve") + translationManager.emptyString
+              visible: isValidOpenAliasAddress(addressLine.text)
+              enabled : isValidOpenAliasAddress(addressLine.text)
+              onClicked: {
+                  var result = walletManager.resolveOpenAlias(addressLine.text)
+                  if (result) {
+                      var parts = result.split("|")
+                      if (parts.length == 2) {
+                          var address_ok = walletManager.addressValid(parts[1], appWindow.persistentSettings.testnet)
```
stoffu: `testnet` was replaced by `nettype` when stagenet was introduced. The removed code above was already using it, so I wonder why it's changed here.

pages/Transfer.qml
cryptochangements34: Good catch, thanks

skftn:
Introduces some awkward space when there is no button:

cryptochangements34:
@skftn You're right that looks strange. Fixed now

skftn:
@cryptochangements34 awesome 🎉

PR #1331, opened 2018-04-19
[RELEASE] Settings: use checkbox instead of button for daemon mode switching

stoffu:


skftn:
Merged in #1287 

+invalid
+resolved

PR #1344, opened 2018-04-22
Redesigned shared ringdb panel

skftn:
Fixes #1254

- Rearranged components
- Made the panel responsive (for mobile)

PR #1348, opened 2018-04-23
remove ringdb select/load button anchors

cryptochangements34:
The select/load buttons are anchored to the input line way too close

ringdb-gui-before

By removing this anchoring they are given a bit of space

ringdb-gui-after


skftn:
#1344 redesigns this whole page

PR #1353, opened 2018-04-24
Settings: do not invoke status command when using remote node

stoffu:


skftn:
Do you know why it invokes the status command when you go to the settings page at all, as in, what does it accomplish? (other than having output in the console popup window)

stoffu:
@skftn 
As far as I can see, the status command is invoked via `daemonManager.running()` in order to set the property `appWindow.daemonRunning` on page load completion which is used to control the visibility of `startDaemonButton` and `stopDaemonButton`.

skftn:
Alright. Good work anyhow!

PR #1356, opened 2018-04-25
Remove old unused files

pazos:
BasicPanel.qml hasn't been used since 2016.
build_libwallet_api.sh has been superseded by get_libwallet_api.sh

I also removed monero-core.rc - the resource file for windows-, which is created by qmake if RC_ICONS is specified in monero-wallet-gui.pro


skftn:
@Jaqueeee Can `build_libwallet_api.sh` be removed?

PR #1357, opened 2018-04-25
main.cpp misc changes

pazos:
- refactor some platform dependant code. At least for me is easier to read this way.

- add isDesktop, to differenciate between desktops OSes and mobile OSes. It is useful mainly because android is recognized by qt as linux too.

- add a initial cmdline parser. Move some qDebug  and force style fusion for desktop platforms so don't pollute the output of the parser. It will be useful in the future :smiley: 

- move monero stuff after app constructor + cmdline parser. (We don't want to initialize anything if we're running, for example monero-wallet-gui --version)

- force platform xcb on desktop linux, because on newer systems qt will try to use platform wayland and this isn't supported by our building system yet.

- fixed reported height and width, because it was right only when height > width.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -124,6 +124,20 @@ int main(int argc, char *argv[])
 
     qDebug() << "app startd";
 
+    // screen settings
+    // Mobile is designed on 128dpi
+    qreal ref_dpi = 128;
+    QRect geo = QApplication::desktop()->availableGeometry();
+    QRect rect = QGuiApplication::primaryScreen()->geometry();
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -73,32 +78,70 @@ void messageHandler(QtMsgType type, const QMessageLogContext &context, const QSt
 
 int main(int argc, char *argv[])
 {
-    Monero::Utils::onStartup();
+    // platform dependant settings
+#if ! defined(Q_OS_ANDROID) && ! defined(Q_OS_IOS)
```
skftn: Previous code used `!defined(` as opposed to `! defined(`

Issue #1068, opened 2018-01-09
GUI displays unexpected path to wallet log file


ordtrogen:
Why does it point to 
/usr/bin/monero-wallet-gui.log

(see screenshot)

System:

OS: Manjaro Linux i686
Model: Compaq Mini 110c-1000 03A81000000000001000
Kernel: 4.9.74-2-MANJARO
Packages: 976
Shell: bash 4.4.12
CPU: Intel Atom N270 (2) @ 1.600GHz
GPU: Intel Integrated Graphics
GPU: Intel Integrated Graphics
Memory: 298MiB / 995MiB

Daemon installed from https://aur.archlinux.org/monero.git
I typically keep it running continuously 

Recently installed GUI to begin testing:

GUI installed from https://aur.archlinux.org/monero-wallet-qt.git

First time I ran it, decided to create a wallet on testnet
When I start GUI now, it displays this:

![skarmbild_2018-01-09_23-01-43](https://user-images.githubusercontent.com/15184875/34745818-14039f1e-f592-11e7-8ffc-524c24163cc3.png)

ordtrogen:
Maybe worth mentioning these errors displayed on console on startup:

![skarmbild_2018-01-09_23-13-07](https://user-images.githubusercontent.com/15184875/34746022-dd70f6d0-f592-11e7-918e-a6b86c211d33.png)

medusadigital:
that looks like a potential bug to me.

however, please check permissioning on the file here too: https://monero.stackexchange.com/questions/2866/where-are-the-monero-core-configuration-parameters-stored

ordtrogen:
I found the config file but its location is different from what the SE question mentions. I have it

~/.config/monero-project/monero-core.conf

Are you suggesting that the properties mentioned in the console output (daemonLogPath et al) should be mentioned in the .conf file?

mmbyday:
+resolved 
by #1800

dEBRUYNE-1:
+resolved

PR #1358, opened 2018-04-27
qt/qml message logger

### LINKED ISSUE #1181: Standard log location (Linux & Android)
For monero-wallet-gui the logfile is in the same folder as the binary. This is great when running the GUI from its folder (ie: static build) but a bad choice If I want to run the GUI from /usr/bin, /usr/local/bin, where the user running the binary doesn't have permissions to write at those folders.

Using a standard location, like $HOME/.monero/monero-wallet-gui.log as a fallback would be great.

pazos:
By default qt/qml messages are sent to stdout on X11 (and maybe on osx) and to the debugger on windows. The rest of the messages are stored in a logfile which is saved in the same path of the binary).

This PR moves all qt/qml debug messages to the same file that the rest of logging and adds a cmdline option to specify a custom logfile.

The new behaviour is:
if no custom logpath is specified use OS defined path (in Logger.cpp), which should be writable.
if a custom logfile is specified use that.

default logPath by OS:
win -> same folder of the app
mac -> $HOME/Library/Logs
linux -> $HOME/

I've removed code in WalletManager.h/cpp and Wallet.h/cpp which I think is not used elsewhere.

Fixes #1181 , #1068

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -224,7 +224,7 @@ int main(int argc, char *argv[])
 
 // Exclude daemon manager from IOS
 #ifndef Q_OS_IOS
-    const QStringList arguments = QCoreApplication::arguments();
+    const QStringList arguments = (QStringList) QCoreApplication::arguments().at(0);
     DaemonManager * daemonManager = DaemonManager::instance(&arguments);
     engine.rootContext()->setContextProperty("daemonManager", daemonManager);
 #endif
--- a/main.cpp
+++ b/main.cpp
@@ -123,7 +123,7 @@ int main(int argc, char *argv[])
 
     // Log settings
     const QString logPath = getLogPath(parser.value(logPathOption));
-    Monero::Wallet::init(argv[0], "monero-wallet-gui", logPath.toStdString().c_str());
+    Monero::Wallet::init(argv[0], "monero-wallet-gui", logPath.toStdString().c_str(), true);
     qInstallMessageHandler(messageHandler);
 
 
--- a/main.cpp
+++ b/main.cpp
@@ -294,13 +294,13 @@ int main(int argc, char *argv[])
     QObject *qmlCamera = rootObject->findChild<QObject*>("qrCameraQML");
     if (qmlCamera)
     {
-        qDebug() << "QrCodeScanner : object found";
+        qWarning() << "QrCodeScanner : object found";
         QCamera *camera_ = qvariant_cast<QCamera*>(qmlCamera->property("mediaObject"));
         QObject *qmlFinder = rootObject->findChild<QObject*>("QrFinder");
         qobject_cast<QrCodeScanner*>(qmlFinder)->setSource(camera_);
     }
     else
-        qDebug() << "QrCodeScanner : something went wrong !";
+        qCritical() << "QrCodeScanner : something went wrong !";
 #endif
 
     QObject::connect(eventFilter, SIGNAL(sequencePressed(QVariant,QVariant)), rootObject, SLOT(sequencePressed(QVariant,QVariant)));
--- a/main.cpp
+++ b/main.cpp
@@ -54,6 +54,7 @@
 #include "Subaddress.h"
 #include "model/SubaddressModel.h"
 #include "wallet/api/wallet2_api.h"
+#include "Logger.h"
 #include "MainApp.h"
 
 // IOS exclusions
@@ -69,12 +70,6 @@ bool isIOS = false;
 bool isAndroid = false;
 bool isWindows = false;
 bool isDesktop = false;
-
-void messageHandler(QtMsgType type, const QMessageLogContext &context, const QString &msg)
-{
-    // Send all message types to logger
-    Monero::Wallet::debug("qml", msg.toStdString());
-}
 
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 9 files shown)


main.cpp
```diff
@@ -117,16 +112,24 @@ int main(int argc, char *argv[])
     app.installEventFilter(eventFilter);
 
     QCommandLineParser parser;
+    QCommandLineOption logPathOption(QStringList() << "l" << "log",
```
stoffu: How about using `--log-file` like all the other CLI apps, as opposed to the current implementation where the supplied path is assumed to be a directory? I think this is less confusing.

main.cpp
```diff
@@ -117,16 +112,24 @@ int main(int argc, char *argv[])
...
     parser.addHelpOption();
     parser.process(app);
 
     Monero::Utils::onStartup();
 
     // Log settings
-    Monero::Wallet::init(argv[0], "monero-wallet-gui");
-//    qInstallMessageHandler(messageHandler);
+    const QString logPath = getLogPath(parser.value(logPathOption));
+    Monero::Wallet::init(logPath.toStdString().c_str(), "");
```
stoffu: Shouldn't this be `init(argv[0], logPath.toStdString().c_str())`?

pazos: In that case the logfile will be /path/to/bin/process_name.log and the second argument won't be used. See https://github.com/monero-project/monero/blob/master/contrib/epee/src/mlog.cpp#L68. Please correct me if I'm wrong

stoffu: Looks like the `mlog_get_default_log_path` function is implemented incorrectly. I've made a patch for it: https://github.com/monero-project/monero/pull/3736

I still think the first and second arguments to `init()` should be `argv[0]` and `logPath.toStdString().c_str()`, respectively, because that agrees with the intention expressed in `set_module_name_and_folder(const std::string& path_to_process_)`.

moneromooo-monero:
I see you removed the info/warning/etc API. Is this something that's done via existing API which your code is now hooking into ? It also seems all the logs created by the GUI will be "frontend" (QML or not). Is that intended ?

moneromooo-monero:
About the log file path, it looks like you want to override the log file path, but Wallet::init doesnot let you do that. Maybe you can create another Wallet::init which lets you do that. And rename these initLogging, and replace the current init func with an empty body (or remove it, since it looks like the init code is done in an onStartup function now).

pazos:
@moneromooo-monero: Sorry, If I don't understand the first part of your question. I removed the WalletManager::categories because they weren't called anywhere in the code. I'm using the underlying Monero::wallet:categories.  The idea is use those to log all qDebug statements in the code. 

> It also seems all the logs created by the GUI will be "frontend" (QML or not). Is that intended ?

Maybe frontend is not the best name out there, but yes it was intended: using a single category to log all qt/qml stuff. Not sure if this is a good idea.

moneromooo-monero:
They were supposed to be used from qml for future logs. That said, I'm not hacking on the GUI much, so whatever others prefer is fine.

pazos:
> About the log file path, it looks like you want to override the log file path, but Wallet::init doesnot let you do that. Maybe you can create another Wallet::init which lets you do that. And rename these initLogging, and replace the current init func with an empty body (or remove it, since it looks like the init code is done in an onStartup function now).

Again, I don't know if I got it. Can I rewrite Wallet::init as

```
void Wallet::init(const char *logfile, const bool console_output) {
#ifdef WIN32
    // Activate UTF-8 support for Boost filesystem classes on Windows
    std::locale::global(boost::locale::generator().generate(""));
    boost::filesystem::path::imbue(std::locale());
#endif
    mlog_configure(logfile, console_output);
}
``` 
???

Or leave the wallet::init as is and create Wallet::initLogging with those contents?. As far as I can read the cli isn't using the api at all, so if only the gui is using the Wallet::init I would go for rewritting that.

moneromooo-monero:
Other software might be using init. The monero code itself isn't using this apart from the GUI.

stoffu:
I'm starting to find this patch more complicated than necessary; wouldn't a simple patch like below be enough, instead of adding/deleting lots of lines of code?

```diff
diff --git a/main.cpp b/main.cpp
index 38db1c5..9bac2da 100644
--- a/main.cpp
+++ b/main.cpp
@@ -68,7 +68,16 @@
 void messageHandler(QtMsgType type, const QMessageLogContext &context, const QString &msg)
 {
     // Send all message types to logger
-    Monero::Wallet::debug("qml", msg.toStdString());
+    const std::string cat = "qml"; // category displayed in the log
+    const std::string msg2 = msg.toStdString();
+    switch(type)
+    {
+        case QtDebugMsg: Monero::Wallet::debug(cat, msg2); break;
+        case QtInfoMsg: Monero::Wallet::info(cat, msg2); break;
+        case QtWarningMsg: Monero::Wallet::warning(cat, msg2); break;
+        case QtCriticalMsg: Monero::Wallet::error(cat, msg2); break;
+        case QtFatalMsg: Monero::Wallet::error(cat, msg2); break;
+    }
 }
 
 int main(int argc, char *argv[])
@@ -82,7 +91,7 @@ int main(int argc, char *argv[])
 
     // Log settings
     Monero::Wallet::init(argv[0], "monero-wallet-gui");
-//    qInstallMessageHandler(messageHandler);
+    qInstallMessageHandler(messageHandler);
 
     MainApp app(argc, argv);
 
```

I personally don't see much value in the ability to specify the custom log path, because most users likely start the GUI just by double clicking the app icon.

Also I don't like putting unrelated changes into a single commit (in this case, the deletion of apparently unused functions).

pazos:
@stoffu: On osx the log shouldn't be inside the app (also the path reported by the GUI isn't correct). On linux the log shouldn't be on the same path as the binary specially if we like some sort of deb/rpm packaging in the future) But yes, the rest is true. There is no need to modify the logpath by command line.

I'm fine finding a good path for each OS and leave the rest as is. And yes, I'll keep unrelated changes outside the commit, sorry :p

stoffu:
On a second thought, I find your patch reasonable and I retract my above criticism. The only change I find necessary is:

```diff
diff --git a/main.cpp b/main.cpp
index 9d9242c..353cd91 100644
--- a/main.cpp
+++ b/main.cpp
@@ -123,7 +123,7 @@ int main(int argc, char *argv[])
 
     // Log settings
     const QString logPath = getLogPath(parser.value(logPathOption));
-    Monero::Wallet::init(logPath.toStdString().c_str(), "");
+    Monero::Wallet::init(argv[0], "monero-wallet-gui", logPath.toStdString().c_str());
     qInstallMessageHandler(messageHandler);
```
which requires this patch I made just now: https://github.com/monero-project/monero/pull/3745

Also, the PR #1337 uses the `walletLogPath` property, so I guess the `getLogPath()` function needs to be accessible from Settings.qml in some way.

PR #1360, opened 2018-04-29
qmake: misc changes

pazos:
move deployment commands to QMAKE_POST_LINK, to don't need to call `make deploy` from build dir. 
( This breaks builbot, because it has its own rules ). In case it is not good the README should be updated because there is a missing step after the ./build.sh for linux/osx

make qmake config modular:
- move deployment stuff to deployment.pri. (deployment.pri was generated as part of the qt project and wasn't changed in 4 years)
- move translation stuff to translations.pri

### DIFF
--- a/deployment.pri
+++ b/deployment.pri
@@ -1,27 +1,32 @@
-android-no-sdk {
-    target.path = /data/user/qt
-    export(target.path)
-    INSTALLS += target
-} else:android {
-    x86 {
-        target.path = /libs/x86
-    } else: armeabi-v7a {
-        target.path = /libs/armeabi-v7a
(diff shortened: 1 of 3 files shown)


selsta:
Would you be interested in reopening these as separate PRs? Might be easier to review then. I’m definitely interested in removing the `make deploy` step.

pazos:
@selsta: changes needed are in this commit https://github.com/monero-project/monero-gui/pull/1360/commits/da3f85421d293715d41065ab940861738ca98140.

It should work against master but will generate errors during de deploy step of buildbot c.i. (since now this step does not exist)

PR #1369, opened 2018-05-03
Dutch translation update

ProkhorZ:
This is the Dutch translation of strings added shortly before and after the upgrade in April.

Line numbers are different because the source text used is from a version refreshed by @erciccione: https://github.com/erciccione/monero-core/tree/nl-updated/translations

There was a warning (not an error) in QT Linguist because of duplicate messages. I tracked this down to what's currently line 110 in pages/SharedRingDB.qml, where there are two qsTr strings on one line, while the line number is used to identify them:

```
            text: "<style type='text/css'>a {text-decoration: none; color: #FF6C3C; font-size: 14px;}</style>" +
                  qsTr("Blackballed outputs") + " <a href='#'>" + qsTr("Help") + "</a>" + translationManager.emptyString
```

### DIFF
--- a/translations/monero-core_nl.ts
+++ b/translations/monero-core_nl.ts
@@ -4,63 +4,57 @@
 <context>
     <name>AddressBook</name>
     <message>
-        <location filename="../pages/AddressBook.qml" line="50"/>
+        <location filename="../pages/AddressBook.qml" line="64"/>
         <source>Address</source>
         <translation>Adres</translation>
     </message>
     <message>
-        <location filename="../pages/AddressBook.qml" line="56"/>
+        <location filename="../pages/AddressBook.qml" line="51"/>
         <source>Qr Code</source>
         <translation>QR-code</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


ProkhorZ:
I did what you suggested two weeks ago. The translation took a long time because I had to research the Shared RingDB page. I'd be surprised if everybody translated that stuff correctly.

When will there be a code freeze and for how long? The whole point of this update is that there wasn't a code freeze for translation during the fork.

erciccione:
> I did what you suggested two weeks ago

Yes, and during these two weeks the GUI have been heavily updated (you can see that branch is 72 commits behind monero-gui/master) and it's not even done yet. I'm not telling you to throw away your work, since most of the strings you worked on will be usable, but just to wait so we don't have to open another one in few days with the updated file.

>When will there be a code freeze and for how long?
The code freeze will happen as soon as 1358 1362 1359 will be merged, then i will refresh all translation files and translators will be able to work on a definitive document.

PR #1376, opened 2018-05-08
[0.12.1] Refresh all language files

erciccione:
In average, there are 150 new strings and 75 were removed because obsolete


erciccione:
For reviewers: 
this is just the result of running `lupdate -pro monero-wallet-gui.pro -no-obsolete`

PR #1378, opened 2018-05-08
[0.12.1] Fix typos in Receive.qml and SharedRingDB.qml

erciccione:
thanks @einsteinsfool for spotting them

### DIFF
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -433,7 +433,7 @@ Rectangle {
                     onLinkActivated: {
                         receivePageDialog.title  = qsTr("QR Code") + translationManager.emptyString;
                         receivePageDialog.text = qsTr(
-                            "<p>This QR code includes the address you selected above and" +
+                            "<p>This QR code includes the address you selected above and " +
                             "the amount you entered below. Share it with others (right-click->Save) " +
                             "so they can more easily send you exact amounts.</p>"
                         )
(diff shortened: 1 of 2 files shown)


einsteinsfool:
Since you're fixing the typos, I found another one.

`<p><font size='+2'>This is a simple sales tracker:</font></p><p>Let your customer scan that QR code to make a payment (if that customer has software which supports QR code scanning).</p><p>This page will automatically scan the blockchain and the tx pool for incoming transactions using this QR code. If you input an amount, it will also check that incoming transactions total up to that amount.</p>It's up to you whether to accept unconfirmed transactions or not. It is likely they'll be confirmed in short order, but there is still a possibility they might not, so for larger values you may want to wait for one or more confirmation(s).</p>`

`<p><font size='+2'>This is a simple sales tracker:</font></p><p>Let your customer scan that QR code to make a payment (if that customer has software which supports QR code scanning).</p><p>This page will automatically scan the blockchain and the tx pool for incoming transactions using this QR code. If you input an amount, it will also check that incoming transactions total up to that amount.</p><p>It's up to you whether to accept unconfirmed transactions or not. It is likely they'll be confirmed in short order, but there is still a possibility they might not, so for larger values you may want to wait for one or more confirmation(s).</p>`

Added `<p>` in the 5th line. The upper text is the original. The bottom one is fixed.

erciccione:
Fixed @einsteinsfool

Issue #1012, opened 2017-12-11
Misc Tabbing Related Hotkey Bugs


Timo614:
The hotkeys are a bit bugged with how they are currently setup.

1. On Mac OS X if I click the `alt` key while I'm on the "Advanced" tab I'm sent to the previous tab I was on but no other tabs have this behavior (guessing since the hotkeys logic sets the leftPanel to the current main state but the advanced tab has no content / mainstate by itself so it reverts to the last tab)
2. The `ctrl+tab` or `alt+tab` behavior seems random "Send" -> "Receive" -> "Prove/check"(under advanced) -> "History" -> "Address book"(under send) -> "Sign/verify"(under advanced) -> "Settings"
3. The "alt+shift+backspace" logic doesn't appear to do anything on a mac (perhaps since it's the delete key on macs?)

erciccione:
+resolved

PR #1379, opened 2018-05-08
[shortcuts]: ctrl+tab switches pages forward | ctrl+shift+tab switch pages backward (same order as leftpanel)

pazos:
1. Fix ctrl+tab behaviour:

ctrl+tab behaviour is switching between pages in certain order:
`transfer->receive->txKey->sharedRingDB->history->addressBook->sign->settings`

This looks random on the gui because how those pages are placed in leftpanel menu.
I changed 'ctrl+tab behaviour to switch pages in this order: 
`transfer->addressbook->receive->history->mining->txkey->sharedRingDB->sign->settings`

I added mining page to the switchable pages but leave the keys page out because it asks for a password and somewhat breaks the workflow.

2. Fix ctrl+shift+tab behaviour (ctrl+tab reverse order)

### DIFF
--- a/filter.cpp
+++ b/filter.cpp
@@ -34,13 +34,20 @@ filter::filter(QObject *parent) :
     QObject(parent)
 {
     m_tabPressed = false;
+    m_backtabPressed = false;
 }
 
 bool filter::eventFilter(QObject *obj, QEvent *ev) {
     switch(ev->type()) {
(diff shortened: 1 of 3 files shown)


pazos:
Hey, updated to fix ctrl+shift+tab (switching pages backwards).

Fixes #1012 

Please review

PR #1380, opened 2018-05-08
[wip] remove duplicated shortcut for SharedRingDB page

### LINKED ISSUE #1374: 'ctrl-a' shortcut conflict
**Problem**
This shortcut is used for selection of the whole text on the cursor location, using it also for Shared RingDB tab is leading to confusion for users trying to select the whole string in some field (they suddenly find themselves in a completely different tab -> at least my case initially)

**Solution**
Assign some other shortcut for for Shared RingDB tab (anything that does not collide with some commonly used OS shortcuts)

pazos:
This shortcut was declared twice, one as ctrl+s and the other as ctrl+a. Removed the last one.

Fixes #1374


pazos:
It happens that Ctrl+S is used for both send and sharedRingDB. So I'm not sure about how to proceed.

pazos:
done in #1488, closing this

Issue #1377, opened 2018-05-08
Update all language files for 0.12.1 - Call for translators


erciccione:
The GUI is frozen, now we need to update all existent languages. Keep in mind that the new GUI will be released once all languages will be updated, so please, if you are working on something else, **give the maximum priority to the GUI**.

 ~~I refreshed all language files (see #1376 ), you can find the uploaded version of each language on the `refresh-lang` branch of my repo. Please **use these updated files, NOT the ones on the official repository**. Updated language files can be found at this link:~~ 

~~https://github.com/erciccione/monero-core/tree/refresh-lang/translations~~

Pull the file related to your language from ~~that link~~

https://github.com/monero-project/monero-gui/tree/master/translations

translate the missing strings and, when done, open a PR against monero-project/monero-gui referencing to this issue.

**! Do not update the .ts file and don't run any command, just pull the correct file for your language and translate it. Stick to this workflow, please. !**

## Languages and tasks
Refer to the Taiga task to know the status of each language (if a task's status is 'in progress' or 'assigned to', somebody else is already working on it). I listed down here all available languages, as soon as a translation is PRd, the correspondent language will be thicked.

- [ ] **Arabic** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/255)
- [ ] **Catalan** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/256)
- [x] *Merged* **Czech** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/284)
- [x] *merged* **Danish** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/260)
- [x] *Merged* **German** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/264)
- [x] *Merged* **Esperanto** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/262)
- [x] *Merged* **Spanish** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/277)
- [x] *merged* **French** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/263)
- [x] *Merged* **Hebrew** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/265)
- [ ] **Croatian** - [Taiga task](https://taiga.getmonero.org/project/erciccione-monero-localization/task/259)
[...]

PR #1384, opened 2018-05-08
Updated Polish Monero GUI translation for v0.12.1

einsteinsfool:
Updated Polish translation for the upcoming v0.12.1 release as mentioned in #1377.

If you want to review the translations, you can start with ctrl+f "blackball". I used "Zablokowany" for "Blackballed" and "Odblokowany" for "Unblackballed". Blackballing probably refers to [this](https://lab.getmonero.org/pubs/MRL-0001.pdf) paper (as mentioned by **xmrscott**). You can ctrl+f "marble" in the paper for more info. Basically blackballing means not allowing an output to be used in ring signature. Maybe someone has a better translation. Imo, even in English blackballing is a bad phrase. I'd just use Block/Unblock or Blacklist/Unblacklist.

Also, I don't feel that my translations with "heights" are optimal. So you can ctrl+f "height" as well. E.g. "restore height", "wallet creation height", etc.

### DIFF
--- a/translations/monero-core_pl.ts
+++ b/translations/monero-core_pl.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="pl_PL">
+<TS version="2.1" language="pl_PL">
 <context>
     <name>AddressBook</name>
     <message>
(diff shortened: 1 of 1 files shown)


erciccione:
Hey @szogun1987 could you review this PR please?

PR #1388, opened 2018-05-09
use radio buttons for daemon wizard

cryptochangements34:
Addresses the second part of #1338

PR #1401, opened 2018-05-11
use radio buttons for daemon wizard

cryptochangements34:
#1388 but before I screwed it up trying to rebase


cryptochangements34:
just uhhh forget this happened

PR #1415, opened 2018-05-14
export/import key images

cryptochangements34:
This functionality is needed to get the true balance of a view-only wallet

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -539,6 +539,30 @@ Rectangle {
                     submitTxDialog.open();
                 }
             }
+            
+            StandardButton {
+                id: exportKeyImagesButton
+                text: qsTr("Export key images") + translationManager.emptyString
+                small: true
+                visible: !appWindow.viewOnly
+                enabled: pageRoot.enabled
+                onClicked: {
+                    console.log("Transfer: export key images clicked")
+                    exportKeyImagesDialog.open();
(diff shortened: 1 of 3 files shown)


pages/Transfer.qml
```diff
@@ -539,6 +539,30 @@ Rectangle {
...
+                    console.log("Transfer: export key images clicked")
+                    exportKeyImagesDialog.open();
+                }
+            }
+
+            StandardButton {
+                id: importKeyImagesButton
+                text: qsTr("Import key images") + translationManager.emptyString
+                small: true
+                visible: appWindow.viewOnly && walletManager.isDaemonLocal(appWindow.currentDaemonAddress)
```
stoffu: Why does the connected daemon need to be local?

cryptochangements34: https://github.com/monero-project/monero/blob/4b728d7dd48584987f53995a141baac4f886f017/src/wallet/api/wallet.cpp#L1020 I'm not sure on the reasoning behind it but the wallet api function wants a trusted daemon, which AIUI is only local daemons since there is no `trusted-daemon` flag/option in the GUI.

luigi1111: I believe it queries the daemon with the "is_key_image_spent" function, which reveals to the daemon  the key images that the wallet is interested in.

Issue #1436, opened 2018-05-31
Add explicit option for choosing Mainnet on Welcome to Monero! in gui wallet


hvr1:
Enhancement: Add option for choosing Mainnet above Testnet and Stagenet. At the moment double clicking either will enable mainnet. However it would be more intuitive to have this as an explicit option.

![monero gui](https://user-images.githubusercontent.com/39798404/40780732-7efbb140-64ea-11e8-8070-7a7e5fc02ba1.PNG)

skftn:
+enhancement

baryluk:
Even better remove both all options, and require to use a command line flag or entry in the config file. This is not something normal user needs to know or accidentally select.

mmbyday:
+resolved

dEBRUYNE-1:
+resolved

PR #1452, opened 2018-06-07
add explicit mainnet button and move nettypes to advanced options

cryptochangements34:
Most users don't need to worry about  testnet/stagenet so they are moved to an "Advanced Options" tab. Under this tab a "Mainnet" button is also added along side the other nettypes to address #1436

### DIFF
--- a/components/CheckBox2.qml
+++ b/components/CheckBox2.qml
@@ -28,6 +28,7 @@
 
 import QtQuick 2.0
 import QtQuick.Layouts 1.1
+import QtGraphicalEffects 1.0
 import "." 1.0
 
 RowLayout {
@@ -40,6 +41,7 @@ RowLayout {
     property int fontSize: 14 * scaleRatio
     property alias fontColor: label.color
     property int textMargin: 8 * scaleRatio
+    property bool darkDropIndicator: false
     signal clicked()
     height: 25 * scaleRatio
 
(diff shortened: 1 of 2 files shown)


stoffu:
Currently `CheckBox2` uses `whiteDropIndicator.png` for the tick icon which becomes a bit difficult to see on a light gray background. The following patch adds a boolean option `darkDropIndicator` to `CheckBox2` which allows the tick icon to become dark:

```javascript
--- a/components/CheckBox2.qml
+++ b/components/CheckBox2.qml
@@ -28,6 +28,7 @@
 
 import QtQuick 2.0
 import QtQuick.Layouts 1.1
+import QtGraphicalEffects 1.0
 import "." 1.0
 
 RowLayout {
@@ -40,6 +41,7 @@ RowLayout {
     property int fontSize: 14 * scaleRatio
     property alias fontColor: label.color
     property int textMargin: 8 * scaleRatio
+    property bool darkDropIndicator: false
     signal clicked()
     height: 25 * scaleRatio
 
@@ -74,12 +76,19 @@ RowLayout {
                 anchors.left: label.right
                 anchors.leftMargin: textMargin
                 color: "transparent"
+                rotation: checkBox.checked ? 180  * scaleRatio : 0
 
                 Image {
                     id: indicatorImage
                     anchors.centerIn: parent
                     source: "../images/whiteDropIndicator.png"
-                    rotation: checkBox.checked ? 180  * scaleRatio : 0
+                    visible: !darkDropIndicator
+                }
+                ColorOverlay {
+                    anchors.fill: indicatorImage
+                    source: indicatorImage
+                    color: "#FF000000"
+                    visible: darkDropIndicator
                 }
             }
```

Also, now that these buttons really become the standard radio buttons, `checked = true;` needs to be added to the `onClicked` event of each of these buttons, otherwise it'd be possible to have all of them unchecked.

screen shot 2018-06-07 at 16 49 44

PR #1454, opened 2018-06-09
Fix for frozen settings page

pazos:
currently we are checking daemon status each time we go to settings page (when local daemon is selected). This causes a hang of a few seconds (it depends on monerod's workload). 

This is a proposal to remove this check and rely on  "onWalletConnectionStatusChanged", as we do for leftpanel's network status items.

I did check the behaviour of this patch against a local daemon (started both inside and outside the gui) and it seems to work fine for me. More testing would be required and reviews apreciated because this might be some stupid PR.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -358,6 +358,10 @@ ApplicationWindow {
         middlePanel.updateStatus();
         leftPanel.networkStatus.connected = status
 
+        // update local daemon status.
+        if(!isMobile && walletManager.isDaemonLocal(appWindow.persistentSettings.daemon_address))
+            daemonRunning = status;
+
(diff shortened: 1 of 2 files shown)


MaxXor:
@luigi1111 Can this get included in the v0.12.2.0 release as the binaries are not built yet?

luigi1111:
@MaxXor probably, but it needs to be PR'd against the branch (too) for that to happen.

pazos:
I have no problem about submitting this PR against release branch but I would like to make sure that this change is safe across different OSes and user setups.

MaxXor:
I can tell for Debian 9 amd64 it's working fine.

PR #1457, opened 2018-06-10
[WIP] implement system tray menu with contextual actions

pazos:
Don't merge.

This PR implements a system tray icon/menu with the following features: hide to systray/ show from systray and close app from systray. It also includes a slot to show system notifications, but is unused for now.

Having a system tray  can be useful to implement the following features:

1. Open/close wallets from systray
2. Lock current wallet ^^
3. Open qml pages ^^
4. Show system notifications at will.
5. View wallet/network status on icon hover
6 ...

The main reason to have a systray for me in the long run is to start monerod as a child of the gui (instead of detached) and being able to leave monerod running with all wallets closed, but a nice icon on notifications area. Sending commands to monerod via rpc is slower than using an interactive daemon session and blocks qt event loop. 

Feedback would be awesome. Both about the implemented stuff and about future implementation, ideas, etc.

### DIFF
--- a/filter.cpp
+++ b/filter.cpp
@@ -114,6 +114,12 @@ bool filter::eventFilter(QObject *obj, QEvent *ev) {
         QMouseEvent *me = static_cast<QMouseEvent*>(ev);
         emit mouseReleased(QVariant::fromValue<QObject*>(obj), me->x(), me->y());
     } break;
+    case QEvent::WindowBlocked: {
+        emit windowBlocked(true);
+    } break;
+    case QEvent::WindowUnblocked: {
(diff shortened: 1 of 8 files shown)


pazos:
@stoffu 

> Does the monero submodule need to be changed in some ways?

Not at all. Current features are just qml/qt stuff. Other features will need modifications in DaemonManager and maybe more modifications in eventFilter.

I'm not sure about how to add new persistentSettings to make systray (and systray dependant options) settable from UI.Some help would be awesome :p

pazos:
Closing because I have no time and it seems that nobody cares about it :smiley:

PR #1458, opened 2018-06-11
win build: link against libwinscard.a for PCSC

stoffu:
I encountered the same error reported in https://github.com/monero-project/monero-gui/issues/1455#issuecomment-396085411.

Does this mean that the following in monero-wallet-gui.pro:
```
CONFIG += c++11 link_pkgconfig
packagesExist(libpcsclite) {
    PKGCONFIG += libpcsclite
}
```
didn't have the intended effect for Windows build?

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -242,6 +242,7 @@ win32 {
         -licudt \
         -licutu \
         -liconv \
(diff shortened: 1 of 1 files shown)


stoffu:
Ouch, once again :) #1444

PR #1479, opened 2018-06-29
Added Lithuanian language.

kasperaitis:


translations/monero-core_lt.ts
```diff
@@ -0,0 +1,2781 @@
...
+        <source>Amount</source>
+        <translation>Suma</translation>
+    </message>
+</context>
+<context>
+    <name>RemoteNodeEdit</name>
+    <message>
+        <location filename="../components/RemoteNodeEdit.qml" line="67"/>
+        <source>Remote Node Hostname / IP</source>
+        <translation type="unfinished">Nutolusios jungties IP</translation>
```
erciccione: Forgotten `unfinished` label or left on purpose?

translations/monero-core_lt.ts
```diff
@@ -0,0 +1,2781 @@
...
+        <location filename="../pages/Settings.qml" line="695"/>
+        <source>Are you sure you want to rebuild the wallet cache?
+The following information will be deleted
+- Recipient addresses
+- Tx keys
+- Tx descriptions
+
+The old wallet cache file will be renamed and can be restored later.
+</source>
+        <translation></translation>
```
erciccione: This must be fixed. In this state no string will be picked and it will result in an empty message

translations/monero-core_lt.ts
```diff
@@ -0,0 +1,2781 @@
...
+<context>
+    <name>StandardDropdown</name>
+    <message>
+        <location filename="../components/StandardDropdown.qml" line="163"/>
+        <source>Low (x1 fee)</source>
+        <translation>Žemas (x1 rinkliavos)</translation>
+    </message>
+    <message>
+        <location filename="../components/StandardDropdown.qml" line="164"/>
+        <source>Med
```
erciccione: What happened here?

translations/monero-core_lt.ts
kasperaitis: Thanks. Fixed.
Hopefully I squashed right. 
I hardly understand 'blackball' section so any terminology would help.

translations/monero-core_lt.ts
kasperaitis: Fixed everything.
Hopefully squashed it right.
I'm hardly understand  'blackball' section so any terminology would help.

kasperaitis:
Thanks. I will check that GUI guide and will do my best to translate that part correctly in a near future.

PR #1491, opened 2018-07-03
mobile view: draggable window

xiphon:


stoffu:
Are you working on the mobile mode? I haven't seen anyone doing that for quite a while.
If possible, can you attach some screenshot or video demonstrating how it works?

xiphon:
Take a look at https://github.com/monero-project/monero-gui/blob/master/main.qml#L72
Wallet changes windows layout based on the window's size. 

To reproduce the issue just resize the window to activate mobile layout and you will notice that you can't drag it anymore. The commit fixes it.

stoffu:
Ah, right. I forgot that the mobile mode becomes active also when the window is resized.

PR #1496, opened 2018-07-06
[release] Add support for creating hardware wallet

stoffu:


src/libwalletqt/WalletManager.cpp
```diff
@@ -104,6 +104,20 @@ Wallet *WalletManager::createWalletFromKeys(const QString &path, const QString &
...
+{
+    QMutexLocker locker(&m_mutex);
+    if (m_currentWallet) {
+        qDebug() << "Closing open m_currentWallet" << m_currentWallet;
+        delete m_currentWallet;
+        m_currentWallet = NULL;
+    }
+    Monero::Wallet * w = m_pimpl->createWalletFromDevice(path.toStdString(), password.toStdString(), static_cast<Monero::NetworkType>(nettype),
+                                                         deviceName.toStdString(), restoreHeight, subaddressLookahead.toStdString());
+    m_currentWallet = new Wallet(w);
```
xiphon: if (w) {
  m_currentWallet = new Wallet(w);
}

stoffu: I'm not sure about your suggestion; this coding style was copied from the existing codebase. If error handling is to be done right, it has to be done thoroughly and simultaneously with all other similar parts.

xiphon: Yep, could be done with separate commit fixing all the similar places. Just thought it would be nice to fix it right here, once you touched this peace of code.

stoffu: Please feel free to make a PR and propose changes about this.

wizard/WizardCreateWalletFromDevice.qml
```diff
@@ -0,0 +1,125 @@
...
+
+    function onPageOpened(settingsOblect) {
+        checkNextButton()
+    }
+
+    function onPageClosed(settingsObject) {
+        settingsObject['account_name'] = uiItem.accountNameText
+        settingsObject['wallet_path'] = uiItem.walletPath
+        var restoreHeight = parseInt(uiItem.restoreHeight);
+        settingsObject['restore_height'] = isNaN(restoreHeight)? 0 : restoreHeight;
```
moneromooo-monero: That might need double checking, it's a bit wtf

wizard/WizardCreateWalletFromDevice.qml
```diff
@@ -0,0 +1,125 @@
...
+            console.log("deleting wallet")
+        }
+
+        var tmp_wallet_filename = oshelper.temporaryFilename();
+        console.log("Creating temporary wallet", tmp_wallet_filename)
+        var nettype = appWindow.persistentSettings.nettype;
+        var restoreHeight = settingsObject.restore_height;
+        var subaddressLookahead = settingsObject.subaddress_lookahead;
+
+        var wallet = walletManager.createWalletFromDevice(tmp_wallet_filename, "", nettype, "Ledger", restoreHeight, subaddressLookahead);
```
moneromooo-monero: Maybe make the "ledger" part a config/param, since the hw support code was made to be non ledger specific.

wizard/WizardMain.qml
```diff
@@ -233,6 +245,7 @@ ColumnLayout {
         appWindow.persistentSettings.auto_donations_amount = false //settings.auto_donations_amount
         appWindow.persistentSettings.restore_height = (isNaN(settings.restore_height))? 0 : settings.restore_height
         appWindow.persistentSettings.is_recovering = (settings.is_recovering === undefined)? false : settings.is_recovering
```
moneromooo-monero: Some code compares typeof(x) to undefined, this compares the value. Can a boolean be "undefined"  in javascript ?

skftn: I'm guessing it's a fail-safe against invalid attribute access, for whenever `is_recovering` is not assigned to the object `settings` at all. [Usages](https://github.com/monero-project/monero-gui/search?q=is_recovering&unscoped_q=is_recovering) show that such situation could perhaps take place, all tough I'm not completely sure.

Kind of off-topic: The current variable scope for js is kind of a mess and I'd vouch to get everything into seperate `.js` files, with classes that encapsulate logic.

wizard/WizardCreateWalletFromDevice.qml
stoffu: It does look a bit wtf, but generally I have very little idea about how things should be done QML, and this piece of code was mostly copied from WizardRecoveryWallet.qml:

https://github.com/monero-project/monero-gui/blob/cd46edb23f0764b7bb7ee1171224881554be5d2f/wizard/WizardRecoveryWallet.qml#L70

I think someone who knows QML better should handle these issues in the future refactoring.

stoffu:
@moneromooo-monero 
> Maybe make the "ledger" part a config/param, since the hw support code was made to be non ledger specific.

OK, this is what I could do with my challenged Qt-fu:
screen shot 2018-07-11 at 2 37 57

Other widgets like radio buttons could have been used, but the dropdown UI seems like the most suitable. However, I couldn't figure out a way to make it look nice with the white background. The mismatch of the color themes between the wizard and the main panels is haunting us here...

But the choice for the device name is currently the Ledger only, so I don't see much point in trying to provide an option for the user to choose. The CLI also doesn't allow the user to choose the device name FWIW.

stoffu:
Problem fixed.

PR #1503, opened 2018-07-09
Change application name back to monero-core (revert 28fb9fa)

cryptochangements34:
Changing the application name caused wallets created prior to v0.12 to "forget" their config file

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -100,7 +100,7 @@ int main(int argc, char *argv[])
 
     MainApp app(argc, argv);
 
-    app.setApplicationName("monero-gui");
+    app.setApplicationName("monero-core");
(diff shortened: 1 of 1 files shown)


ordtrogen:
Maybe I haven't understood the issue (and feel free to ignore this), but

> wallets created prior to v0.12 to "forget" their config file

if "wallet" refers to the wallet file, does this file contain a ref to the config file? (I didn't know that)

if "wallet" refers to the program, wouldn't installing a v.0.12.x version render the prior one obsolete?

cryptochangements34:
"wallet" means the program and depending on whether this was implemented before or after the v0.12 release, yes you would get the same issue. For this reason I much prefer #1505

PR #1502, opened 2018-07-09
[Release] Change application name back to monero-core (revert 28fb9fa)

cryptochangements34:
Changing the application name caused wallets created prior to v0.12 to "forget" their config file

@dEBRUYNE-1

PR #1505, opened 2018-07-10
Support for the legacy config file/registry key 'monero-core'

skftn:
The application name was switched to `monero-gui` for 0.12.2. This led to the GUI trying to locate `~/.config/monero-gui.conf` for the settings, ignoring the old `~/.config/monero-core.conf` resulting in lost configuration. 

On Windows, the same happened in the registry `HKEY_LOCAL_USER\Software\monero-core\...`

This PR sets the application name depending on if it can find the old config file/key, allowing new users to adopt the new scheme and old users keep their config.

It also adds a line to Debug info on the settings page:

#1502 #1503

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -33,6 +33,7 @@
 #include <QIcon>
 #include <QDebug>
 #include <QObject>
+#include <QSettings>
 #include <QDesktopWidget>
 #include <QScreen>
 #include "clipboardAdapter.h"
@@ -103,6 +104,20 @@ int main(int argc, char *argv[])
     app.setApplicationName("monero-gui");
     app.setOrganizationDomain("getmonero.org");
     app.setOrganizationName("monero-project");
+
+    // support for older configuration files/registry keys (monero-core.conf)
+    // @TODO: Legacy. Remove in 2019.
+    QString configPath;
+    QSettings *settingsLegacy = new QSettings(QSettings::NativeFormat, QSettings::UserScope, "monero-project", "monero-core");
+    if(!settingsLegacy->value("customDecorations").isNull()) {
+        configPath = settingsLegacy->fileName();
+        app.setApplicationName("monero-core");
+    } else {
+        QSettings *settings = new QSettings(QSettings::NativeFormat, QSettings::UserScope, "monero-project", "monero-gui");
+        configPath = settings->fileName();
+        delete settings;
+    }
+    delete settingsLegacy;
 
 #if defined(Q_OS_LINUX)
     if (isDesktop) app.setWindowIcon(QIcon(":/images/appicon.ico"));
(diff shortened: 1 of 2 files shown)


cryptochangements34:
I much prefer this solution over over #1502 and #1503. Well done!

rbrunner7:
I must say I am less than enthusiastic. This "solution" looks nice "on paper" but cements a little tussle from middle of 2018 more or less forever into the code. Larger code, more complexity, a chance that something around settings in the future will go wrong because somebody, somewhere is not aware about this quirk and writings settings again to the wrong place ...

And if there is ever a need for some other program to look at the settings in effect of the GUI wallet, it has to apply this switching logic as well if it wants to be watertight, again more or less for all eternity.

Can't we make up our mind whether we stay with `monero-core` or make a clean cut to `monero-gui`? (Personally I am for the clean cut and taking the hit of *some* people with mysterious re-downloads of the blockchain. Shit happens, and here it more or less already happened anyway.)

ordtrogen:
Would it be possible externalize the logic by placing it in the installer, the "make install" target or something like that? 

E.g copying settings from a legacy file if it exits, change its name to <something>-obsolete.conf and then use the new name

rbrunner7:
@ordtrogen: Of course this would be possible. But: I started to investigate how much work that would be in my InnoSetup-based installer and saw that I would probably work several hours on this. I came to the conclusion that it is not worth it.

That's why I despair a little about the reluctance of the community to decide once and for all "switch back" or "go forward and take the hit".

I am also seriously considering ripping out the whole "set blockchain location" feature that I programmed with much love in the installer. In the case of an update it could collide with changes done in the GUI wallet, for example, and people may not know how to change it in the icon that I try to provide to directly start the daemon.

And of course a settings location switch like it is proposed here would make it fail anyway ...

medusadigital:
i share the same opinion as @rbrunner7 

any additional logic is simply risky and unnecessary here. Risk/reward is not given for this change.

dEBRUYNE-1:
>Can't we make up our mind whether we stay with monero-core or make a clean cut to monero-gui? 

If we have to opt for this I'd argue it would be better to stay with `monero-core`. If we stay with `monero-core`, the only users affected are those that started using the GUI from v0.12.2.0 on. Those users will have to reset their preferences (and reopen their wallet). However, for users that were using earlier version of the GUI, it will simply pick up the old config file. By contrast, if we make a clean cut to `monero-gui`, every single users will have to reset their preferences. The former option, arguably, is less prone to user errors and results in less subsequent support work.

rbrunner7:
@skftn:

> I started working on this last night, then discovered on Windows it saves the configuration in the registry. Recursively copying reg entries from C++ is a pain. Not worth it.

Exactly :)

skftn:
Closing in favor of #1502 and #1503

PR #1507, opened 2018-07-11
[Release] Support for legacy configuration files/registry keys (monero-core.conf)

skftn:
The application name was switched to `monero-gui` for 0.12.2. This led to the GUI trying to locate `~/.config/monero-gui.conf` for the settings, ignoring the old `~/.config/monero-core.conf` resulting in lost configuration. 

On Windows, the same happened in the registry `HKEY_LOCAL_USER\Software\monero-core\...`

This PR sets the application name depending on if it can find the old config file/key, allowing new users to adopt the new scheme and old users keep their config.

It also adds a line to Debug info on the settings page:

#1502 #1503 #1505

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -33,6 +33,7 @@
 #include <QIcon>
 #include <QDebug>
 #include <QObject>
+#include <QSettings>
 #include <QDesktopWidget>
 #include <QScreen>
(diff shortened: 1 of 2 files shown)


skftn:
Closing in favor of #1502 and #1503

PR #1515, opened 2018-07-17
Display the config path in the debug section

skftn:
Will show a registry path under Windows.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -104,6 +104,11 @@ int main(int argc, char *argv[])
     app.setOrganizationDomain("getmonero.org");
     app.setOrganizationName("monero-project");
 
+    QString configPath;
+    QSettings *settings = new QSettings(QSettings::NativeFormat, QSettings::UserScope, app.organizationName(), app.applicationName());
+    configPath = settings->fileName();
+    delete settings;
+
 #if defined(Q_OS_LINUX)
     if (isDesktop) app.setWindowIcon(QIcon(":/images/appicon.ico"));
 #endif
@@ -269,6 +274,7 @@ int main(int argc, char *argv[])
 
     engine.rootContext()->setContextProperty("defaultAccountName", accountName);
     engine.rootContext()->setContextProperty("applicationDirectory", QApplication::applicationDirPath());
+    engine.rootContext()->setContextProperty("configPath", configPath);
 
     bool builtWithScanner = false;
 #ifdef WITH_SCANNER
--- a/pages/Settings.qml
+++ b/pages/Settings.qml
@@ -738,6 +738,20 @@ Rectangle {
                 font.pixelSize: 14
                 text: walletLogPath
             }
+
+            TextBlock {
+                Layout.fillWidth: true
+                font.pixelSize: 14
+                text: qsTr("Config path: ") + translationManager.emptyString
+            }
+
+            TextBlock {
+                Layout.fillWidth: true
+                font.pixelSize: 14
+                text: {
+                    configPath
+                }
+            }
         }
     }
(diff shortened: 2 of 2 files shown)


main.cpp
```diff
@@ -104,6 +104,11 @@ int main(int argc, char *argv[])
     app.setOrganizationDomain("getmonero.org");
     app.setOrganizationName("monero-project");
 
+    QString configPath;
+    QSettings *settings = new QSettings(QSettings::NativeFormat, QSettings::UserScope, app->organizationName, app->applicationName);
```
stoffu: I had to change this line as below in order to get it compile:
```diff
-    QSettings *settings = new QSettings(QSettings::NativeFormat, QSettings::UserScope, app->organizationName, app->applicationName);
+    QSettings *settings = new QSettings(QSettings::NativeFormat, QSettings::UserScope, app.organizationName(), app.applicationName());
```

cryptochangements34:
Works as expected on ubuntu after applying the compile fix that stoffu mentioned

MaxXor:
Please fix the PR. It includes a bunch of other commits.

skftn:
@stoffu 

Do you have `/Users/stoffu/Library/Preferences/com.monero-project.monero-gui.plist` by any chance?

stoffu:
@skftn
No.

main.cpp
```diff
@@ -104,6 +104,11 @@ int main(int argc, char *argv[])
     app.setOrganizationDomain("getmonero.org");
     app.setOrganizationName("monero-project");
 
+    QString configPath;
+    QSettings *settings = new QSettings(QSettings::NativeFormat, QSettings::UserScope, app.organizationName(), app.applicationName());
```
pazos: AFAIK there is no need to do dynamic memory allocation here. 

`QSettings settings(QSettings::NativeFormat, QSettings::UserScope, app.organizationName(), app.applicationName());`

should work

skftn:
> This PR just allocate memory for a new QSettings instance and set its format and scope, but doesn't try to use that instance for read/store values from there.

This is correct. In the QML, we use `import Qt.labs.settings 1.0` which magically creates `QSettings` for us. I don't mean to replace `Qt.labs.settings` by creating our own in C++ and then passing it to QML. I just want to know where the path to our config file is. So I `delete` the `Qsettings` object after I retreived the filename.

main.cpp
skftn: @pazos We can't switch to `QSettings::IniFormat` without some sort of migration though.

On Windows, registry->.ini
On OSX, plist->.ini (?)
On Linux, .conf->.ini

Would say that's out of this scope for this PR.

pazos:
> In the QML, it uses import Qt.labs.settings 1.0 which magically creates QSettings for us with default values. I don't mean to replace that by creating our own in C++ and then passing it to QML. I only want to know where the path to our config file is. So I delete the Qsettings object after retreiving the config path.

This is risky because we are not sure if the settings are stored there (or in any other path magically :p). I see two options here:

1. Wait until qml implements a [filename method for settings](https://doc-snapshots.qt.io/qt5-dev/qml-qt-labs-settings-settings.html)
2. Implement our own qml module in c++ (with value, setValue and filename methods) and use that instead.

main.cpp
pazos: linux stores settings in ini, it doesn't have its own native format.

migration is ok, but can lead to bugs and add ugly code. IMO is better to do the switch in a major release and never look back 😄

skftn:
This seems tricky. I might come back to this later. Closing for the time being.

PR #1525, opened 2018-07-27
Windows installer: Updates for 0.12.3.0

rbrunner7:
Based on the contents of the official download file `monero-gui-win-x64-v0.12.3.0.zip` as released earlier today.

Mostly trivial adjustments to new release number, time passed, and some changes in the files to install.

With the confusion about the application name ("gui" versus "core") caused by 0.12.2.0 behind us, the installer can safely write any special blockchain location to the `monero-core` registry key where the GUI wallet will pick it up.

As an improvement, in the case of an *update* where the user could have modified the blockchain location in the GUI wallet already installed, the installer now picks up that location from the registry (instead of re-using stale data i.e. what the user said in the last installer run).

### DIFF
--- a/installers/windows/Monero.iss
+++ b/installers/windows/Monero.iss
@@ -8,7 +8,7 @@ AppName=Monero GUI Wallet
 ; Thus it's important to keep this stable over releases
 ; With a different "AppName" InnoSetup would treat a mere update as a completely new application and thus mess up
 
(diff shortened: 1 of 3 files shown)


skftn:
Hard for me to test; can only rubber stamp. Looks good to me though, thanks!

Issue #1512, opened 2018-07-17
[Bug] Suddenly returns to initial wizard when pressing ESC in History's detail window


stoffu:
https://streamable.com/68158

The first time I clicked on the "?" icon, I closed the window normally by clicking "OK". The second time I opened the tx details window, I pressed the Esc key, and somehow the window size shrunk a bit, and then after clicking "OK" I was back at the initial wizard screen.

skftn:
+bug

dEBRUYNE-1:
+resolved

PR #1527, opened 2018-07-28
fixed: disable password dialog hotkeys handling when it is hidden

xiphon:
#1512


skftn:
Very nice! will test later.

PR #1530, opened 2018-07-28
Fix build instructions for Linux

MaxXor:
`QT_SELECT=5` is still required on Debian 9 and it also doesn't harm on any other Linux system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Packaging for your favorite distribution would be a welcome contribution!
 
     ```
     cd monero-gui
-    ./build.sh
+    QT_SELECT=5 ./build.sh
     ```
 
 The executable can be found in the build/release/bin folder.


skftn:
I believe it's already enforced here:

https://github.com/monero-project/monero-gui/blob/254553f90ea7144ab25b123bfece98f720138cde/monero-wallet-gui.pro#L2

MaxXor:
That's not the same, this enforces the minimum version WHEN qt is found. When I run `./build.sh` on latest Debian 9 I tells me it doesn't know which qt version to chose.

This PR just restores the build instructions which were changed in this commit: https://github.com/monero-project/monero-gui/commit/08f61e25de378b2b37fd7a4374d7a085ea1ff69f#diff-04c6e90faac2675aa89e2176d2eec7d8L130

skftn:
Sounds good. Thanks.

PR #1532, opened 2018-07-28
Correctly maximize window

MaxXor:
Handle maximizing of the window just like any other application. Currently maximizing shows the wallet in fullscreen which also hides the taskbar.


skftn:
Nice changes, keep 'm coming :+1:

PR #1446, opened 2018-06-04
Add support for creating hardware wallet

stoffu:
In order for this to build, the Monero submodule needs to be on the master branch.

screen shot 2018-06-04 at 16 51 42

CC: @cslashm

PR #1535, opened 2018-07-31
[RELEASE] Some fixes for Ledger wallet creation

stoffu:
Not sure if there's going to be another point release. I'll update the corresponding pending PR #1446 for the master branch.


luigi1111:
There is not going to be another 12 release. Closing this in favor of #1446

PR #1540, opened 2018-08-08
[DO NOT MERGE] pro: link against libsodium on Linux

radfish:
This is needed to link against libmonero_wallet from monero
with PR #4159 merged.

Please merge AFTER: https://github.com/monero-project/monero/pull/4159 is merged

Not touching other platforms because can't test.

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -116,6 +116,7 @@ LIBS += -L$$WALLET_ROOT/lib \
         -lepee \
         -lunbound \
         -leasylogging \
+        -lsodium \
 }
 
 android {


cryptochangements34:
maybe mark [DO NOT MERGE] in the title then change it back after #4159 is merged?

stoffu:
I believe linking to libsodium is already necessary since the merge of #3999 which starts to use the library function `crypto_verify_32`.

monero-wallet-gui.pro
```diff
@@ -116,6 +116,7 @@ LIBS += -L$$WALLET_ROOT/lib \
         -lepee \
         -lunbound \
         -leasylogging \
+        -lsodium \
```
MoroccanMalinois: Since it's also required for android, could you please move that addition to line 292 ?

BigslimVdub:
@stoffu have you checked this addition on OS X when building?

stoffu:
@BigslimVdub 
Yes, this is needed for Aeon, too. https://github.com/aeonix/aeon-gui/pull/11

PR #1546, opened 2018-08-22
Corrected language code for Ukrainian and renamed localized file

ordtrogen:
Pinging @TheFuzzStone @LvMsterfild for verification

The file was called monero-core_ua.ts and it contained the attribute language="ua_UA", but the language code for Ukrainian is "uk" (UA being the country code). This must have been a typo, am I right?

NB. the file lang/languages.xml was already correct which supports the typo thesis: 

image


TheFuzzStone:
Yes, you're right.

PR #1553, opened 2018-09-04
Prevent accidental off-by-one decimal point transfers

skftn:
[From Reddit](https://www.reddit.com/r/Monero/comments/9ap96g/gui_v01230_cli_v01230_both_with_direct_ledger/e5d40gb/):

> I have a complaint about the GUI xmr client, hope this is an appropriate place to comment this. I just tried to send a smaller amount of monero .2 xmr. When you type in .2 only 2 xmr fills in, the period does not fill in and I almost didn't realize. You have to type zero first to send less then 1, 0.2, the first char has to be a number not a period. Can this be changed to either accept the period first, or if a period is pressed to fill in 0. before it. If I didn't notice I could have sent 2 xmr instead of .2.

I changed the regex to support amounts that begin with a period, however, the actual validation that checks the amount will not let you create a transaction with amounts that start with a period. Should solve some confusion.

### DIFF
--- a/components/LineEdit.qml
+++ b/components/LineEdit.qml
@@ -196,7 +196,7 @@ Item {
         MoneroComponents.Input {
             id: input
             anchors.fill: parent
-            anchors.leftMargin: inlineIcon.visible ? 38 : 0
+            anchors.leftMargin: inlineIcon.visible ? 44 * scaleRatio : 0
             font.pixelSize: item.fontSize
             font.bold: item.fontBold
             onEditingFinished: item.editingFinished()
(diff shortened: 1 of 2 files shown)


skftn:
@MaxXor Like the report from reddit specifies, someone pressed `.2` there and due to regex, it eats up the period and you'll end up with `2`. If you're not paying attention, you could send ten times as much as you intended.

dvsdan:
If a value that starts with a period is invalid why not check if a period is pressed as the first character in the field and just insert the chars "0." before the regex check, instead to make the user's transaction amount valid.

skftn:
@dvsdan PR welcome.

PR #1602, opened 2018-10-06
Add submodule 'monero-GUI-guide'

erciccione:
This PR add the GUI guide (https://github.com/monero-ecosystem/monero-GUI-guide) as a submodule, so that it can be included directly in the monero-gui repository.
Some issues need to be addressed:

- This PR updates the instructions in the README. Instead of the simple `git clone`, it instructs to use `git clone https://github.com/monero-project/monero-gui.git && cd monero-gui && git submodule init monero-GUI-guide && git submodule update`, so that the guide's submodule will be also cloned.
This to avoid to use `--recurse-submodules`, since it's not intended to clone the 'monero' submodule along with the GUI. Is this ok? Would you prefer to simply clone 'monero-gui' and give the option in the README to clone also the guide? Is there a better way to handle this? Should we just clone the GUI repo and include the guide only when the wallets is built? (as it happens for the 'monero' submodule).

- As @fluffypony suggested, the guide should be copied into the 'build' folder. I guess we need to edit the makefile to do that, but that goes beyond my knowledge. Could somebody take care of it?

@moneromooo-monero @rbrunner7

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = monero
 	url = https://github.com/monero-project/monero
 	ignore = all
+[submodule "monero-GUI-guide"]
+	path = monero-GUI-guide
+	url = https://github.com/monero-ecosystem/monero-GUI-guide.git
--- a/README.md
+++ b/README.md
@@ -106,14 +106,13 @@ Packaging for your favorite distribution would be a welcome contribution!
       `emerge dev-qt/qtmultimedia:5 media-gfx/zbar`
 
 
-3. Clone repository
+3. Clone repository (including the guide)
 
-    `git clone https://github.com/monero-project/monero-gui.git`
+    `git clone https://github.com/monero-project/monero-gui.git && cd monero-gui && git submodule init monero-GUI-guide && git submodule update`
 
 4. Build
 
     ```
-    cd monero-gui
     QT_SELECT=5 ./build.sh
     ```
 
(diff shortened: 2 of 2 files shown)


erciccione:
Taking out the WIP tag, since this could theoretically already be merged and a new PR to include the guide in the build process could be opened after.

rbrunner7:
Well, I am afraid I am not of great help here.

I don't fully understand the handling of the Monero submodule in `get_libwallet_api.sh`. I think not being able to simply use `--recurse-submodules` or `git submodule init/update` without parameters would be dangerous because that may be thing most people would try first after some missing files in reflex, and if that already messes up something ... But as I said, no idea how to organize or re-organize submodule handling here to get that out of the way.

As to the copying of the guide into the build directory, I am out of the loop here as well. I just wonder how the recent build directory reorg for Monero will influence the GUI build process, and maybe that has to go through first before it makes sense to add instructions somewhere to copy around the guide.

pazos:
@erciccione: do we need a markdown to html parser or is the guide intended to be in md? A md2html parser is easy to integrate in the code. Sadly a md2pdf is not an option since it requires a whole latex distribution.

How the user invokes the guide?, it is just a file to open with a text editor?

erciccione:
@pazos 
I'm open to every option here. At the end, the guide is already available in markdown online, and i will release the PDF version with every major release anyway, so the point is to have something as simple as possible to use for the user, so something easy to open with a text editor or a simple html page sound like the best options IMO.

skftn:
Is anyone open to the idea of having a very simple C++ program that converts markdown to HTML? I mean, not much needs to happen. And this would work without requirements. And I would have an excuse to code C++...

skftn:
@pazos hoedown looks nice but looks like someone is already working on a solution; https://github.com/monero-ecosystem/monero-GUI-guide/pull/56

Issue #1582, opened 2018-10-02
All languages need to be updated for 0.13 - Call for translators


erciccione:
Another hard fork is approaching, and with it, a mayor release: 0.13
Since 0.12.3, the GUI changed a bit. Some strings were added and some were removed. #1572 refreshed all language files, so now **they can all be updated with the missing translations**, before 0.13 is out.

**This is a call for translators**, down here can be found the list of the available languages. If the language is thicked, somebody worked on it and the translation is already PRd.

**If you wish to contribute, please make sure your language is still labelled as 'needed' on the [Issue tracker of the Localization Workgroup on Taiga](https://taiga.getmonero.org/project/erciccione-monero-localization/taskboard/test)**. The task related to this issue is "Update strings (0.13)". If the task is 

+ **Needed**: nobody worked on it and still need a translator.
+ **Work in progress**: somebody is actively working on the language, but haven't opened a PR yet. 
+ **Ready for test**: the translation was PRd on this repository (the link to the PR can be found in the description of the task)

**If you start to work on a language, please write here or let me know** using one of the communication channel listed below, so that i can update the issue tracker and avoid double works.

- [x] **Arabic** - #1591 - *Merged*
- [ ] **Bulgarian** - New! Merged in #1637 , but won't be included in 0.13
- [x] **Catalan** - #1620 (not sure it will make it for 0.13)
- [x] **Chinese (Simplified)** - #1600 - *Merged*
- [x] **Chinese (Taiwanese)** - #1581 - *Merged*
- [ ] **Croatian**
- [x] **Czech** - #1618 - *Merged*
- [x] **Danish** - #1556 - *Merged*
- [x] **Dutch** - ~~#1594~~ replaced by #1599 - *Merged*
- [ ] **Esperanto**
- [x] **Finnish** - #1593 - *Merged*
- [x] **French** - #1586 - *Merged*
- [x] **German** - #1588 - *Merged*
- [x] **Hebrew** - #1603 - *Merged*
- [x] **Hungarian** - #1612 - *Merged*
- [ ] **Indonesian**
- [x] **Italian** - #1584 - *Merged*
- [x] **Japanese** - #1597 - *Merged*
- [ ] **Korean**
- [x] **Lithuanian** - #1608 - *Merged*
- [ ] **Pirate**
- [x] **Polish** - #1589 - *Merged*
- [x] **Portuguese (pt)** - #1596 - *Merged*
[...]

PR #1603, opened 2018-10-06
Translation and corrections for Hebrew #1582

samuelgozi:
Translated whatever I could and also reviewed some of the already translated lines.
I didn't build the GUI, but Im confident that my translations are good.

### DIFF
--- a/translations/monero-core_he.ts
+++ b/translations/monero-core_he.ts
@@ -149,42 +149,43 @@
     <message>
         <location filename="../pages/History.qml" line="72"/>
         <source> selected: </source>
-        <translation type="unfinished"></translation>
+        <translation> נבחרים: </translation>
     </message>
     <message>
         <location filename="../pages/History.qml" line="161"/>
         <source>Search</source>
-        <translation type="unfinished"></translation>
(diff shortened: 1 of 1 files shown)


translations/monero-core_he.ts
```diff
@@ -1168,38 +1175,38 @@ The old wallet cache file will be renamed and can be restored later.
...
         <location filename="../pages/settings/SettingsNode.qml" line="323"/>
         <source>Password</source>
-        <translation type="unfinished">סיסמה</translation>
+        <translation>סיסמה</translation>
     </message>
     <message>
         <location filename="../pages/settings/SettingsNode.qml" line="347"/>
         <source>Connect</source>
-        <translation type="unfinished"></translation>
+        <translation type="unfinished">התחבר</translation>
```
erciccione: If the translation is correct, you can remove `type="unfinished"` here

samuelgozi:
Did my best to fix what you pointed out as well as adding some other small changes and translations.

samuelgozi:
@erciccione Fixed that.

samuelgozi:
I added you as a collaborator

PR #1614, opened 2018-10-08
Lithuanian translations

Medardas:
There are still 6 translations missing however they need some deep dive into Lithuanian semantics (bunch of new terms as well) and Monero technical understanding. Could not finish it before this release

### DIFF
--- a/translations/monero-core_lt.ts
+++ b/translations/monero-core_lt.ts
@@ -16,27 +16,27 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="73"/>
         <source>Payment ID &lt;font size=&apos;2&apos;&gt;(Optional)&lt;/font&gt;</source>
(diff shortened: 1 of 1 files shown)


erciccione:
Hi @Medardas do this PR integrate missing strings in #1608 ? if not, it's a duplicate and i'm afraid it cannot be merged

Medardas:
@erciccione it does integrate a lot of missing strings and fix some already available from #1608 but there are still 6 missing in total.
#1608 had more than 100 missing I believe.

erciccione:
+duplicate

erciccione:
I'm investigating why 'Monero' was translated to 'Bitlitas' the first time and i was wrong. He made that translation.

@kasperaitis Can you explain why it was translated like that in first place?

PR #1616, opened 2018-10-09
fix bad Indonesian translations.

ha-yi:
fix some malaysian language to Indonesian.


translations/monero-core_id.ts
```diff
@@ -11,12 +11,12 @@
     <message>
         <location filename="../pages/AddressBook.qml" line="51"/>
         <source>Qr Code</source>
-        <translation type="unfinished"></translation>
+        <translation type="unfinished">Kode QR</translation>
```
erciccione: If this is correct, please remove the `unfinished` tag

PR #1627, opened 2018-10-10
Update qml.qrc

whotheff:
Bulgarian translation


whotheff:
Please check for errors, this is my first upload.
I added bulgarian translation.

PR #1639, opened 2018-10-12
update outdated icu files

Lafudoci:
Update icu files from 61 to 62.


luigi1111:
If this isn't just relevant to the release, then it's not really a duplicate; both should be merged.

skftn:
@luigi1111 

#1562 into master
#1640 into releasev0.13

Should be correct?

luigi1111:
Oh, that's my fault. I misread the situation. Thanks.

PR #1662, opened 2018-10-16
Minor adjustments to the Swedish translation

snaggen:
When I audited the GUI i found some Swedish translations that were not optimal. 
* Layout was translated as Layout in Swedish, which is not really optimal but works. However, since it really is not about Layout, I changed the Swedish translation to "Utseende" since that better fits the Swedish language and what the page is really about. 
* Sweep Unmixable is a strange sentence to start with, but it was translated to "Svep upp omixbara" which feels like a word by word translation. I interpret the word "Sweep" here as the metaphor where you are sweeping the floor, so you are basically cleaning up the unmixables. So I have adjusted the swedish translation to better fit this metaphor. 
* Also I have changed a string on the Settings Info page to make it more consistent. 

@freminik @ordtrogen can you review this.

Also, I must say that the Swedish translation looks very good. So even if we don't have agreed on everything, the result is still really good... and after all, that is what matters. So kudos for your efforts.

### DIFF
--- a/translations/monero-core_sv.ts
+++ b/translations/monero-core_sv.ts
@@ -359,7 +359,7 @@
     <message>
         <location filename="../pages/Keys.qml" line="93"/>
         <source>WARNING: Do not reuse your Monero keys on another fork, UNLESS this fork has key reuse mitigations built in. Doing so will harm your privacy.</source>
-        <translation>VARNING: Återanvänd inte din Monero nyckel på en annan förgrening, SÅVIDA INTE denna förgrening har återanvändnings begränsningar inbyggda. Görs detta så kan din integritet skadas.</translation>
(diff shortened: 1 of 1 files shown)


snaggen:
I also sent a note on the swedish translator list tp-sv@listor.tp-sv.se

ordtrogen:
Do you mean "Layout" is incorrect in the English source?

snaggen:
Well, basically yes, since the only option there is about client side decorations. So, Appearance would probably be a better source. However, since Layout in Swedish use is imported from English it feels even worse in Swedish, since it is used in a much more narrow scope. Hence, I took the liberty to change it, but I should probably open a bug against the original translation.

snaggen:
Opened issue #1666

snaggen:
Updated the translation based on review by Anders on the translation list tp-sv@listor.tp-sv.se.

Issue #1661, opened 2018-10-16
Non-translatable strings


snaggen:
I looked at the latest GUI and looked for untranslated strings using the Swedish translation that should be complete. I then found some un translated strings, that were not available in the monero-core_sv.ts file. That gave me the following non-translatable strings.

On the Send page:
Automatic (in the transaction priority dropdown).

On the Recieve page it lists adresses, I have an entry saying
`#0  Primary address <adress>`
Where the text "Primary address" is not translatable

On the History page: 
The entries in the sort dropdown is not translatable.
In each entry "Sent", "To", "Transaction ID" is not translatable.

On the Advanced -> Sign/Verify page, these strings are not translatable
"Message from file"
"Verify message"

erciccione:
Thanks @snaggen , the absence of a code freeze make hard to have all strings synced for release. Will fix this asap

erciccione:
#1664 should fix this

erciccione:
+resolved

PR #1664, opened 2018-10-16
Fix strings not marked for translations and resync all languages

erciccione:
#1661


components/HistoryTable.qml
```diff
@@ -151,7 +151,7 @@ ListView {
                 anchors.leftMargin: 18 * scaleRatio
                 font.family: MoneroComponents.Style.fontLight.name
                 font.pixelSize: 14 * scaleRatio
-                text: isOut ? "Sent" : "Received"
+                text: isOut ? qsTr("Sent") + translationManager.emptyString : "Received"
```
moneromooo-monero: Why just one ?

erciccione: Because just one was noticed missing, but yes, make sense to mark both.
Donme, and resynced all files

erciccione:
Rebased

PR #1681, opened 2018-10-18
Rename "blackball" for clarity

moneromooo-monero:
Apparently some people seem to think it's a censorship list...


erciccione:
Translators will hate me

moneromooo-monero:
Right. Everyone knows you're me ^_^

PR #1685, opened 2018-10-19
New translations

jatorrents:
Hi, I'm doing it right?


erciccione:
Hi @jatorrents . Yes, that's perfect. Do you plan to translate more strings or only the present ones?

erciccione:
And the commit message need to be changed in something like 'updates catalan translation'

erciccione:
ping @jatorrents

erciccione:
abandonded. Anyway, all translation should be uploaded on translate.getmonero.org

+invalid

PR #1693, opened 2018-10-22
Fix daemon host/port setting to allow tab switching

sectioneight:
Previously, these were multiline editing (for some reason I'm not sure,
perhaps you can have multiple remote nodes?)

I tried various workarounds, but the simplest was to change them to
simply LineEdit (like username/password) and now my bug is fixed.

Fixes #1688

PR #1695, opened 2018-10-23
Fix daemon host/port setting to allow tab switching

sectioneight:
Previously, these were multiline editing (for some reason I'm not sure,
perhaps you can have multiple remote nodes?)

I tried various workarounds, but the simplest was to change them to
simply LineEdit (like username/password) and now my bug is fixed.

Repeat of #1693, this time against master. 

One day I will get the hang of contributing to this repo :)


MaxXor:
@sectioneight I hope so. ping @luigi1111

PR #1701, opened 2018-10-24
[DO NOT MERGE] build: static linking on Mac

xiphon:


pazos:
@xiphon: why???

Didn't macdeployqt/otool work for you? Frameworks are bundled inside the application and rpath should be changed automatically. Other libraries should be auto picked too.

xiphon:
@pazos

Nevermind. We had some conversation in #monero-gui and was wondering whether Mac static linking is even possible. Opened this PR just to trigger the build bot.

Issue #1610, opened 2018-10-07
[Discussion] Description of the View-Only wallet incorrect


Leza89:
Hey there,

 I wanted to discuss if you guys see a need to change the following description:

> `<location filename="../pages/settings/SettingsWallet.qml" line="185"/>`
> `        <source>Creates a new wallet that can only view transactions, cannot initialize transactions.`

Since a view-only wallet can initialize transactions; Just not sign them.

However my idea

> Creates a new wallet that can only view and send transactions, but cannot sign transactions.

might confuse new users.

Any ideas or do you think we should leave it as is?

skftn:
That indeed seems like a good correction to make. Could you PR for it?

stoffu:
Agreed for the need of change. Like this maybe?

> Creates a new wallet that can only view and initiate transactions, but cannot sign transactions.

Leza89:
Ah.. getting closer

What about

>Creates a new wallet that can only view and initiate transactions, but **requires a spendable wallet to sign and send** transactions.

Is that too convoluted already?

stoffu:
Technically, signing requires the spendable wallet, but sending (submitting) a signed transaction doesn’t. Maybe your version could cause some confusion?

Leza89:
Yes that is the problem. However do you think someone new to crypto (which at this point basically is everyone) will understand the concept of signing? I'm searching for a way to phrase it that'll make it obvious.

>Creates a new wallet that can only view and initiate transactions, but requires a spendable wallet to sign transactions for sending.

stoffu:
Just my guess, but the phrase “sign a transaction” seems to already literally imply (without technical context) giving a clear approval to some tentative action. So your above text seems to read naturally even for new users.

Also, the concept of view-only wallet and cold signing itself already assumes some minimal level of education on the user. If the user can’t understand the description, either look elsewhere for educational material, or not use the feature. After all, the problem being addressed here seems to be the accuracy of the description, not the approachability.

Leza89:
>After all, the problem being addressed here seems to be the accuracy of the description, not the approachability.

When I'm translating I always try to make it as descriptive as possible while keeping it as short as possible. I prefer a phrasing that'll explain itself in one sentence instead of making the user duckduckgo'ing for 2 minutes.

If I got you correct though, you'd say that [this](https://github.com/monero-project/monero-gui/issues/1610#issuecomment-430332666) is a good approach and you'd just omit the "before sending" part?

[...]

PR #1703, opened 2018-10-24
Correction of View-Only Wallet description (see Issue #1610)

Leza89:
* changed

<location filename="../pages/settings/SettingsWallet.qml" line="185"/>
        <source>Creates a new wallet that can only view transactions, cannot initialize transactions.</source>

to

Creates a new wallet that can only view and initiate transactions, but requires a spendable wallet to sign transactions before sending.

* Updated all translations and added "unfinished" tag where necessary

* Updated german translation

### DIFF
--- a/pages/settings/SettingsWallet.qml
+++ b/pages/settings/SettingsWallet.qml
@@ -182,7 +182,7 @@ Rectangle {
                     textMargin: 0
                     leftPadding: 0
                     topPadding: 0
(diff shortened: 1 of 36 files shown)


Leza89:
Link to Issue #1610 for convenience

Leza89:
I've tried MINGW64 (Git Bash) and GitHub Desktop → Open in Command line
Both report that git lupdate`/ `lupdate` and `git update`/`update` are an unknown command (or update is used in the wrong way / unknown command)

So I'll just do it manually, i guess.

Leza89:
Alright. Thanks again.

PR #1706, opened 2018-10-27
receive: add Payment URL

xiphon:
receive


selsta:
screenshot 2018-11-13 at 20 02 07

The color of “Save As” is the same as the background, white.

screenshot 2018-11-13 at 20 03 43

Looks good other than that.

PR #1709, opened 2018-10-28
settings: don't call parseFloat, isFinite to parse height 

xiphon:
`parseInt` is more than enough here


skftn:
Sure. Would be nice to keep the function in `Utils.js` however.

xiphon:
Updated

skftn:
Thanks!

PR #1724, opened 2018-11-03
Fix word in finnish translation

erciccione:
The original word was translated as "public" instead of "private" (at least according automatic translators).
As reported on reddit: https://www.reddit.com/r/Monero/comments/9ti2on/gui_v01304_beryllium_bullet_released/e8xp6c7/?context=3

@rpinola Could you review this?


rpinola:
Translation was wrong. Looks OK now.

PR #1743, opened 2018-11-18
pro: link protobuf

radfish:
Required on Arch Linux (Tested on x86_64), otherwise linking fails with undefined symbols referencing protobuf.

### DIFF
--- a/monero-wallet-gui.pro
+++ b/monero-wallet-gui.pro
@@ -119,7 +119,8 @@ LIBS += -L$$WALLET_ROOT/lib \
         -lepee \
         -lunbound \
         -lsodium \
-        -leasylogging
+        -leasylogging \
+        -lprotobuf
 }
 
 android {


skftn:
Are you sure `-lprotobuf` works on OSX/Win too? Maybe put it into the `linux {` block.

stoffu:
On Mac, protobuf is indeed available from Homebrew and this patch does fix the link error. Not sure about Windows.

xmrdsc:
@Lafudoci Perhaps if you have time you could test this patch on Windows :-)

mmbyday:
As is, this patch :boom: (breaks) build.sh on Windows. 

**Error**
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.1/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lprotobuf

Installing this seems to fix it but the compile time is greatly increased.
pacman -S mingw-w64-x86_64-protobuf

malbit:
> As is, this patch 💥 (breaks) build.sh on Windows.
> 
> **Error**
> C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.1/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lprotobuf
> 
> Installing this seems to fix it but the compile time is greatly increased.
> pacman -S mingw-w64-x86_64-protobuf

for me "pacman -S mingw-w64-x86_64-protobuf" gave nothing. I had to install "mingw-w64-x86_64-protobuf-c" package to get it working. and installing time was not longer than normally
at macOS 10.13.6 after installing "brew install protobuf" GUI does work. Library is called: libprotobuf.17.dylib

dEBRUYNE-1:
#1838 solved this.

dEBRUYNE-1:
+resolved

PR #1746, opened 2018-11-19
[Pootle] First test commit from Pootle

Monero-Pootle:
These translations come from translate.getmonero.org

This GitHub account is managed by @erciccione. The commit is manually created, so a review is always necessary.
Before every commit i run lupdate for all languages, to make sure all translations are synced with the last version of the code

This commit is a test, but if everything went as expected, this PR is ready to be merged as it is.

### DIFF
--- a/translations/monero-core.ts
+++ b/translations/monero-core.ts
@@ -422,42 +422,42 @@
 <context>
     <name>LeftPanel</name>
     <message>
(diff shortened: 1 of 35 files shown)


Monero-Pootle:
Some strings are marked as unfinished even if they are translated. That's not a problem with pootle but with lupdate. I will look for a solution that is not change manually string by string

PR #1749, opened 2018-11-23
transfer: prepend '0' to amount starting with '.' 

### LINKED ISSUE #1718: v13.0.3 GUI does not allow transaction under 1 XMR
OSX 10.13.6
monero-gui-mac-x64-v0.13.0.3.tar.bz2

Trying to send a transaction, GUI does not unlock the send button unless over 1 xmr is entered. If a decimal is the first point, I can not send a transaction. 
It does not change with custom decorations on or off. 

less than 1

more than 1

### LINKED ISSUE #1748: I downloaded the gui wallet but I can't get Daemon to run
I have now been able to download the new Gui wallet but I can't get daemon to sync. Any ideas, I keep getting an error and it says to monerod manually? How do I do that? Also, it says to check my wallet and daemon log for errors.
any ideas how to get the daemon to connect?

thanks,
Fred

xiphon:
@dvsdan's approach https://github.com/monero-project/monero-gui/pull/1553#issuecomment-418468076:
> If a value that starts with a period is invalid why not check if a period is pressed as the first character in the field and just insert the chars "0." before the regex check, instead to make the user's transaction amount valid.

Resolves #1718

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -144,6 +144,11 @@ Rectangle {
                       fontBold: true
                       inlineButtonText: qsTr("All") + translationManager.emptyString
                       inlineButton.onClicked: amountLine.text = "(all)"
+                      onTextChanged: {
+                          if (amountLine.text.startsWith('.')) {
+                              amountLine.text = '0' + amountLine.text;
+                          }
+                      }
 
                       validator: RegExpValidator {
                           regExp: /(.|)(\d{1,8})([.]\d{1,12})?$/


xiphon:
Yep, updated

MaxXor:
Ehm, please check: https://github.com/monero-project/monero-gui/pull/1553#issuecomment-418330506

ping @skftn

xiphon:
Read through https://github.com/monero-project/monero-gui/pull/1553 comments.
Going to update the PR and implement the @dvsdan's approach instead https://github.com/monero-project/monero-gui/pull/1553#issuecomment-418468076:
> If a value that starts with a period is invalid why not check if a period is pressed as the first character in the field and just insert the chars "0." before the regex check, instead to make the user's transaction amount valid.

xiphon:
Updated

PR #1750, opened 2018-11-24
transfer: fix amount validation regexp

xiphon:
**Bugs fixed**
1. A user can enter any character (`!`, `x`, etc.) as the first character in the amount field  
Ex: amount `!123` passes the validation regexp
2. Multiple dots in the amount field
Ex.: amount `.123.123` passes the validation regexp


xiphon:
Can't reproduce, could you double check that you compiled the patched version?

MaxXor:
To test a regular expression I don't need to compile it. I've just checked it locally and it allows the thing I've mentioned.

PR #1753, opened 2018-11-27
get_libwallet_api: configure cmake toolchain for msys

stoffu:
I fail to compile the current master on MSYS2 with this error:
```
C:/msys64/home/stoffu/github/monero-project/monero-gui/monero/external/unbound/validator/val_secalgo.c: In function 'verify_canonrrset':
C:/msys64/home/stoffu/github/monero-project/monero-gui/monero/external/unbound/validator/val_secalgo.c:665:35: error: dereferencing pointer to incomplete type 'EVP_MD_CTX' {aka 'struct evp_md_ctx_st'}
  ctx = (EVP_MD_CTX*)malloc(sizeof(*ctx));
                                   ^~~~
```
This seems related to the backward incompatibility in OpenSSL transitioning from 1.0.x to 1.1.x (and FWIW, the MSYS latest package uses 1.1.1), but strangely, adding these flags to cmake (borrowed from the CLI Makefile) seems to resolve it. I'm not sure why, and I'm surprised that this hasn't been spotted already AFAIK. Also strange is the fact that the buildbot is just fine with the current code. Am I missing something obvious?

### DIFF
--- a/get_libwallet_api.sh
+++ b/get_libwallet_api.sh
@@ -188,14 +188,14 @@ elif [ "$platform" == "mingw64" ]; then
     # Do something under Windows NT platform
     echo "Configuring build for MINGW64.."
     BOOST_ROOT=/mingw64/boost
-    cmake -D CMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE -D STATIC=ON -D BOOST_ROOT="$BOOST_ROOT" -D ARCH="x86-64" -D BUILD_GUI_DEPS=ON -D INSTALL_VENDORED_LIBUNBOUND=ON -D CMAKE_INSTALL_PREFIX="$MONERO_DIR" -G "MSYS Makefiles" ../..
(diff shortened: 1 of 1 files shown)


stoffu:
@xiphon 
Thank you for commenting.

Though I'm not familiar with OpenSSL, I'm not so sure if this is really a bug. AIUI, `EVP_MD_CTX_new` was introduced since 1.1.0, so the fact that it isn't found in the build environment (i.e. `HAVE_EVP_MD_CTX_NEW` is undefined) implies that we're using 1.0.x, and in 1.0.x, the type `EVP_MD_CTX` is not opaque and hence the existing code using `malloc` should compile just fine.

My above error seems to imply as if the linked OpenSSL is from 1.0.x while the headers used for compiling is from 1.1.x, which I find very strange and confusing.

I'm also not sure why specifying the `CMAKE_TOOLCHAIN_FILE` parameter would lead to this difference.

In any case, I believe the cmake build configuration for the wallet API should closely follow the one for the CLI. Or is there a reason not to do so?

xiphon:
Can't reproduce the issue. 

According to the error message `EVP_MD_CTX` is opaque, that mean you are compiling libunbound with OpenSSL 1.1 headers. But in that case `HAVE_EVP_MD_CTX_NEW` should be defined.

Could you check that it is not a build config cache problem? Basically checkout the repo into new folder and attempt to build it.
Could you post the complete build log or steps to reproduce it (MSYS version, installed packages, etc.) ?

PR #1762, opened 2018-11-29
keys: dynamic panel height and qrcode is no longer clipped

mmbyday:
**Before:**
a

**After:**
b

### DIFF
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -150,7 +150,7 @@ Rectangle {
             }, State {
                 name: "Keys"
                 PropertyChanges { target: root; currentView: keysView }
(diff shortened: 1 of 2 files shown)


selsta:
with the patch there is a lot of empty space beneath the QR code, is this necessary?

mmbyday:
Thanks! Open to other suggestions, but this was the best combo of settings that would also allow the mobile sized (aka minimum width window) window to also work :)

selsta:
Even on the smallest window there is this space beneath the QR code on macOS. Is this how it’s supposed to look?

screenshot 2018-11-29 at 23 34 14

mmbyday:
Good eye.
Updated with a 100px offset. 180px was added previously because that was the qr image's minimum height.
aft

selsta:
Looking good now. Can you squash your commits?

PR #1763, opened 2018-11-29
transfer: amount validation regexp improvement

selsta:
Any opinions on this? This would disallow amounts like:

`0003.313`

The CLI allows numbers like above.

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -151,7 +151,7 @@ Rectangle {
                       }
 
                       validator: RegExpValidator {
-                          regExp: /^(\d{1,8})?([\.]\d{1,12})?$/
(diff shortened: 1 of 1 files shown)


xiphon:
Is a completely valid number. Would like to allow allow leading/trailing zeroes.

selsta:
Kinda agree @xiphon, but there is still the problem now that if there are trailing zeros, you can’t add 8 digits of numbers.

`000011111.1` should be valid but is impossible to add.

Anyway this is minor.

selsta:
Updated to allow unlimited leading zeros, not sure if I like this or if I should just close this PR.

mmbyday:
Keep it simple, IMO. No change required. A for effort! 😆

PR #1766, opened 2018-11-30
sign: fix "signature" label layout for better UI consistency

mmbyday:
**Before:**
sb

**After:**
sa

### DIFF
--- a/pages/Sign.qml
+++ b/pages/Sign.qml
@@ -297,15 +297,10 @@ Rectangle {
 
             ColumnLayout {
                 id: verifySignatureRow
-                anchors.topMargin: 17 * scaleRatio
-
-                Label {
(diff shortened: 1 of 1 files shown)


pages/Sign.qml
```diff
@@ -297,15 +297,11 @@ Rectangle {
 
             ColumnLayout {
                 id: verifySignatureRow
-                anchors.topMargin: 17 * scaleRatio
-
-                Label {
-                    id: verifySignatureLabel
-                    text: qsTr("Signature") + translationManager.emptyString
-                }
+                anchors.topMargin: 18 * scaleRatio
```
xiphon:
```suggestion
```

mmbyday: Nice call. Thanks!

PR #1767, opened 2018-11-30
SettingsNode: make TextArea readOnly

mmbyday:
Prevents accidental tabbing and then editing of textArea fields. 😆 

glitch


dEBRUYNE-1:
Just wanted to say nice work. The PRs are much appreciated! If you're acquainted with IRC, we (the GUI contributors) hang out in #monero-gui on freenode.

PR #1768, opened 2018-11-30
transfer: paste Payment URL to fill the payment form fields

xiphon:
paste-payment-url

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -215,6 +215,18 @@ Rectangle {
               wrapMode: Text.WrapAnywhere
               addressValidation: true
               onInputLabelLinkActivated: { appWindow.showPageRequest("AddressBook") }
+              pasteButton: true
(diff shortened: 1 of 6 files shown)


pages/Transfer.qml
```diff
@@ -214,7 +214,19 @@ Rectangle {
               placeholderText: "4.. / 8.."
               wrapMode: Text.WrapAnywhere
               addressValidation: true
-              onInputLabelLinkActivated: { appWindow.showPageRequest("AddressBook") }
+              onInputLabelLinkActivated: { appWindow.showPageRequest("AddressBook") }
```
selsta: trailing whitespace

mmbyday:
Very nice!

Just two observations...

- Could fail more gracefully than pasting everything into address field by notifying user of error.
- Amount is padded with zeros. A whole number amount like 1 turns into 1.000000000000

PR #1770, opened 2018-12-01
Add hidapi for OSX

Blitzkrieg000:
Add brew install hidapi for osx installs of latest Monero GUI. (need this for ledger / trezor support of "usb to app")


selsta:
Trezor required `protobuf`.

Issue #1568, opened 2018-09-22
Various UX issues


Gingeropolous:
copied from some dude on reddit

Yup, without providing solutions then critique is useless. Here's a few issues.

    Onboarding: There's a fundamental lack of onboarding and the wallet literally requires a manual to understand how to use it. Don't even get me started on the Ledger integration, but that's not Moneros fault IMO. Moving forward, it's easy to forget how confusing this space was when you first started but keeping this first time user POV is the key to simplifying products. There should be a mix of onboarding, tooltips, explanations and potentially links to video tutorials (for visual learners)

    Information Hierarchy: The navigational components can (and should) be consolidated because they don't (IMO) justify their own nav items because there's simply not enough content, "Receive" being the main culprit here. As an example, Send and Receive can be consolidated into "Transfers" and address book can be a selector in the Send tab. Here, we went from 3 tabs to one. This type of consolidation is important because it lessens the cognitive load on users and provides the perception of simplicity which can be as important as simplicity itself.

    Semantics: This is the lowest hanging fruit IMO. Protocol semantics should not be presented at the application layer. A few examples of current semantics that should be abstracted into simpler phrases: node, daemon, ring signatures (even though this is awesome) etc.

    Graphics: This is a minor one but none of the images in the wallet are hi-res (svgs) they're all flat 1x pngs, so on retina devices all the assets are pixelated. This is such a small issue, but it's the small things that make an application feel safe, secure and trustworthy.

    Payment ID's: Probably the most confusing things for new users. What is this and why is it important? What happens if I don't add one? How do I add one? Can I type in random chars? Is their char limitations? What do you suggest as an ID structure? Admittedly, these are really asinine questions, but trust me, these are the questions the average user asks themselves.

    Addresses: When I create a new address what happens? Do I need a separate account for these wallets? Do all my balances show in this wallet? How many can I create? Again, such stupid questions but reflective of average user mental models.

    Layout: There's a lot of layout issues and there's UI conflicts where things overlap, have inconsistent paddings, margins etc. We're not using grids at all so things look very inconsistent. This isn't the worst thing in the world because thinking about interfaces spatially and responsively is kinda hard. It's a little easier for me by virtue of the systems I've built but it's these kinda of things that contribute to the overall trustworthiness of the product because everything looks like it was intentionally placed where it is and thought about.

[...]

PR #1774, opened 2018-12-04
Enable clipboard copy functionality on XMR amounts

mmbyday:
Improvement wrt #1568 :: "not being able to copy and paste basic stuff like the wallet balance makes things seem amateurish or buggy (even if that's not the case)."

hi


mmbyday:
Thanks for the improvement idea, @selsta . Updated and improved, thanks to you. :)

PR #1775, opened 2018-12-04
images: remove black background from History arrow

selsta:
`convert downArrow.png -fuzz 10% -transparent black downArrow.png`
`convert upArrow-green.png -fuzz 10% -transparent black upArrow-green.png`


selsta:
Not sure if I should increase the `-fuzz` option even more. It looks fine in the GUI but there is a slight black border left.

MaxXor:
Hm true, the other images don't have such borders.

mmbyday:
Looks good. Will be helpful for the "light" mode theme.

PR #1778, opened 2018-12-04
addressbook: improve readability by truncating long text and allowing…

mmbyday:
… wrapmode on address/description/paymentid fields. Scales nicely with mobile view.

**Before**
bef

**After**
after

### DIFF
--- a/components/AddressBookTable.qml
+++ b/components/AddressBookTable.qml
@@ -28,6 +28,7 @@
 
 import QtQuick 2.0
 import moneroComponents.Clipboard 1.0
+import "../js/TxUtils.js" as TxUtils
 
 ListView {
     id: listView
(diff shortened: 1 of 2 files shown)


selsta:
Not sure if you want to spend more time on the address book, but here are two things I found:

- The add button gets disabled after adding a new address or after only half clicking it (see GIF).
out
- Currently, short payment IDs are not allowed to get added. I don’t see a reason for this, as we want to get away from long payment IDs.

mmbyday:
I wonder if it's a qt version issue. I'm on Qt 5.9.5 and can't reproduce the same bug. What version are you on @selsta ? Thanks 
s

selsta:
I’m on 5.11.2.

selsta:
I’ve just noticed that I have this issue with all orange buttons, so this is not something to fix inside this PR.

mmbyday:
No problem. Tested it on windows with qt 5.11.1. All good too.

Issue #1273, opened 2018-04-05
Fix the default size of the wallet's window on high DPI screens


vp1111:
In a computer running an updated version of Windows 10, with a resolution of 2256 x 1504, when opening the GUI wallet it takes the default size of 971 x 839 (including the title bar). This leads to hiding some information from the bottom part of the screen, as you can see in the image below:

![01-send](https://user-images.githubusercontent.com/34245203/38365241-8e8aef1c-38a9-11e8-97d4-2cab7e77eb86.png)

Ideally, there would be an increase in 18% of the height size. The following image shows a 971 x 995 version (including the title bar):

![01-sendv2](https://user-images.githubusercontent.com/34245203/38365316-d0182242-38a9-11e8-9556-5dd9124fcbb7.png)

I'm unsure if this reproduces in every high DPI screen. Let me know if you need extra info.

curumimxara:
I've noticed in another computer that this only happens if you have the Advanced options section opened, which is stored somewhere in the wallet. When you re-open the wallet the advanced options keeps open, thus giving the impression that something is mixing.

This is very minor.

skftn:
+enhancement

skftn:
+easy

erciccione:
+hacktoberfest

selsta:
#1309 

+duplicate

PR #1781, opened 2018-12-04
info.plist: enable high-DPI support on macOS and iOS

mmbyday:
Please test on macOS and iOS if you can.

https://doc.qt.io/qt-5/highdpi.html

#1273


selsta:
This is not required for macOS. Not sure why, maybe high-DPI is enabled by default on newer systems.

mmbyday:
You're right. Same for Windows. Not needed unless on older versions. Thanks.

PR #1782, opened 2018-12-05
checkbox: remove a gap between rectangle and text, fix layout

xiphon:
Before:
fix-checkbox-before

After:
fix-checkbox


MaxXor:
Why does it allow clicking the checkbox on the right side when the cursor is clearly not on the checkbox+label?

xiphon:
Updated

PR #1789, opened 2018-12-07
PasswordDialog: add ability to toggle text echo mode

mmbyday:


MaxXor:
Yes, an eye icon would be better.

components/PasswordDialog.qml
```diff
@@ -136,12 +137,33 @@ Item {
...
+                            anchors.fill: parent
+                            cursorShape: Qt.PointingHandCursor
+                            hoverEnabled: true
+                            onClicked: {
+                                passwordInput.echoMode = isHidden ? TextInput.Normal : TextInput.Password;
+                                isHidden = !isHidden;
+                            }
+                            onEntered: {
+                                parent.opacity = 0.9
+                                parent.width = 28
```
xmrdsc: All integer constants should terminate with `* scaleRatio`.

mmbyday: Updated. Thank you.

selsta:
Do you remember where you found the eye icon? I’m looking for a 2x version of it.

PR #1790, opened 2018-12-08
submodules: update master to v0.13.0.4 to fix bad offset calculation …

mmbyday:
…otherwise unable to transfer funds :arrow_up:

When compiling GUI on current master branch, this error occurs when attempting to transfer funds on main/test/stagenet.

offset-send


MaxXor:
Yup, right now it's still on 0.13.0.2.

xmrdsc:
I remember the following script working in a such a way that it would pull latest master in:

https://github.com/monero-project/monero-gui/blob/0f05127262cf662ab549deae426c9deebf77967e/get_libwallet_api.sh#L18-L27

But it checks out an older master. I'm confused. Approving this PR.

PR #1791, opened 2018-12-08
standard-button: fix layout, add hover color animation

xiphon:
1. Fix layout
2. Add hover color animation
3. Remove redundant and unused `icon` property, use `rightIcon` with unset `text` instead

Before:
button-before

After:
button-after


selsta:
screenshot 2018-12-11 at 21 39 21
”Stop Daemon“ looks buggy now.

Issue #1559, opened 2018-09-14
GUI build breaking on Windows 8.1 with mingw64


cryptobarbossa:
**Setup:**

- Windows 8.1 x64

- 8GB RAM 

- 100GB+ storage

- Boost 1.68.0-1

- QT 5.11.1-3

**Instructions followed:**

- Install MSYS2 64-bit and open mingw64

- run 'pacman -Syuu' then exit mingw64

- open mingw64 and run 'pacman -Syuu' again

- 'pacman -S mingw-w64-x86_64-toolchain make mingw-w64-x86_64-cmake mingw-w64-x86_64-boost mingw-w64-x86_64-openssl mingw-w64-x86_64-zeromq mingw-w64-x86_64-libsodium'

- 'pacman -S mingw-w64-x86_64-qt5'

- 'pacman -S git'

- git clone --recursive https://github.com/monero-project/monero-gui.git

- cd monero-gui

- ./build.sh

**_Following errors occurred whilst building:_**

Link to full [log](https://pastebin.com/xwxK0MVN)
`Bdynamic -lwinscard -lws2_32 -lwsock32 -lIphlpapi -lgdi32 C:/msys64/mingw64/lib/libQt5Quick.dll.a C:/msys64/mingw64/lib/libQt5Widgets.dll.a C:/msys64/mingw64/lib/libQt5Gui.dll.a C:/msys64/mingw64/lib/libQt5Qml.dll.a C:/msys64/mingw64/lib/libQt5Network.dll.a C:/msys64/mingw64/lib/libQt5Core.dll.a
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/home/lad/monero-gui/monero/lib/libwallet_merged.a(wallet.cpp.obj):wallet.cpp:(.text+0x4ee): undefined reference to `__stack_chk_fail'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/home/lad/monero-gui/monero/lib/libwallet_merged.a(wallet.cpp.obj):wallet.cpp:(.text+0x55a): undefined reference to `__stack_chk_fail'
[...]

PR #1796, opened 2018-12-10
README: add MSYS2 workaround for qmlcachegen Bad address error with qt 5.11.2

stoffu:


xmrdsc:
Thanks for debugging.

stoffu:
Referencing https://github.com/monero-project/monero-gui/issues/1559

PR #1797, opened 2018-12-10
MiddlePanel: add scrollbars :lipstick:

mmbyday:
Scrollbars without requiring the previously not yet stable QtQuick.Controls 2.0 module.

newafterscroll2

Improves usability with lower res screens such as https://github.com/monero-project/monero-gui/issues/1580

PR #777, opened 2017-06-21
Disabled update check

Jaqueeee:
Attempt to fix https://github.com/monero-project/monero-core/issues/761
requires #2101

PR #1804, opened 2018-12-12
MiddlePanel: use QtQuick.Control scrollbar

selsta:
As there are problems with #1797 on macOS (see PR for more details), I decided to try out the QtQuick.Control scrollbar again. Seeing it is used in other places (e.g. [here](https://github.com/monero-project/monero-gui/blob/master/components/DaemonConsole.qml#L159)) I think this shouldn’t crash on Windows (#777) anymore.

Are additional changes to components/Scroll.qml required?

### DIFF
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -29,6 +29,7 @@
 
 import QtQml 2.0
 import QtQuick 2.2
+import QtQuick.Controls 2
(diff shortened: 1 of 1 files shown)


xmrdsc:
Works well on Linux/Win7. Thanks =)

PR #1807, opened 2018-12-12
qrscanner: update addressbook and transfer UI :lipstick:

mmbyday:
Before
beforeqr

After
qr


xmrdsc:
Nice functionality.



^ pretty bad design, does not look nice. Can I propose a button next to the 'Paste' above a lineedit? Just a suggestion, perhaps talk to kneuffelbund about it.

Also we must ensure #1816 compiles on all relevant platforms.

mmbyday:
@xmrdc Good point. Here's a better design making use of the inlinebutton. Code updated.

nwe

As for #1816, it's tested and seems to work on Windows. Can always roll out with #ifdef Q_OS_WIN if the other platforms cannot be implemented in time.

PR #1793, opened 2018-12-09
Merchant page

xmrdsc:
This PR introduces a dedicated merchant page. Current iteration of the merchant features present on the receive page are inadequate to be used as a point of sales system.

The changes in this PR will make it easier for merchants to accept Monero. Run the application in full-screen mode, set the amount to be received and customers scan the QR code.

Alternatively you can provide a payment URL.

Some notes:

- Introduces a menu under Receive called Merchant.
- Optimized for running in full-screen.
- Transaction history shows a 'X time ago' indication. i.e: "the transaction happened 5 minutes ago"
- Incoming transactions are monitored via tx pool.
- For now, clicking on transactions takes you to the History page.
- In the future merchants could optionally specify their to received amount in USD when #1714 is accepted & merged.

[Video](https://i.imgur.com/xk2dlEA.mp4)

PR #1810, opened 2018-12-13
Receive: hide payment URL in QR code menu

selsta:


xiphon:
Don't like this change. It forces even advanced users to read the help to find out that the have to use context menu to copy Payment URL.

Should avoid the context menu. It is counter-intuitive.

Would be better to always show QR code, but hide Payment URL and amount field under the "Advanced options" section. 
What are the downsides?

selsta:
@xiphon had the idea because of @dEBRUYNE-1 comment:

https://github.com/monero-project/monero-gui/pull/1793#issuecomment-446507569

xiphon:
> Would be also okay. I just don’t see a reason to have a huge textbox for the payment URL, all you can do with it is copy anyway.

Yep. Would like to implement "Save As" and "Copy Payment URL" buttons then:
receive-buttons

selsta:
These look good, can you PR them? I tried my own design of this page but failed as I don’t know enough about Qt.

xiphon:
Sure, #1836

PR #1811, opened 2018-12-13
StandardDropdown: fix variable name (> Qt 5.7)

selsta:
Qt 5.12

Newer Qt versions don't like the name "column"

### DIFF
--- a/components/StandardDropdown.qml
+++ b/components/StandardDropdown.qml
@@ -39,7 +39,7 @@ Item {
     property string pressedColor
     property string releasedColor
     property string textColor: "#FFFFFF"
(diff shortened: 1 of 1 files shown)


xmrdsc:
Shadowing built-in keyword or?

selsta:
> Shadowing built-in keyword or?

Not sure what that means.

Before:
before
After:
after

mmbyday:
Shadowing as in reserved word. Overlaps with an existing thing. In either case, this looks good to me :P Thanks

xmrdsc:
Works on 5.7.

PR #1813, opened 2018-12-13
settingsLog: fix layout glitch when loaded first time :bug:

### LINKED ISSUE #1806: Various small UI bugs / possible improvements
- The “Stop daemon” button has the wrong size.
screenshot 2018-12-11 at 21 39 21

- *Settings -> Log:* Looks wrong the first time the GUI is opened. Looks fine afterwards.
screenshot 2018-12-12 at 18 16 13

- *Settings -> Node:* “Stop local node” button has no mouseover hover effect.
- *Settings:* Grey buttons have no mouseover hover effect.
- The height of most of the pages is weird, e.g. why can scroll this far on the Transfer page?
screenshot 2018-12-12 at 18 40 29

mmbyday:
Resolves #1806 (2)

log


selsta:
Did something similar here: #1812

Not sure which is better :)

mmbyday:
Let's test out yours. There was another weird glitch I was encountering with the original layout due to that width / 2 interacting with the gridlayout and visibility logic.

PR #1817, opened 2018-12-13
settings-wallet: use StandardButton instead of custom buttons

xiphon:
#1806 (4)

fix-settings-wallet-buttons


selsta:
There’s also one on the Info page (“Copy to clipboard”).

PR #1820, opened 2018-12-14
Clear text fields upon switching wallets :bug:

### LINKED ISSUE #1402: Text Field is not cleared if switch wallets
Windows x64 GUI Wallet 0.12.0

If you type in an any field on a wallet, then in-wallet close that wallet and open another, the text in that field is still typed in that field in the different wallet.

Does not appear to be a serious issue, however could raise concerns regarding privacy if multiple users share same desktop and use wallet app.

mmbyday:
Fixes #1402


luigi1111:
@mmbyday please rebase.

PR #1814, opened 2018-12-13
Lock wallet on inactivity

xmrdsc:
Fixes #814

- Lock wallet after a configurable amount of minutes; shows password dialog
- **Enabled by default: 10 minutes**
- Added 2 persistent settings: `lockOnUserInActivity` (bool) and `lockOnUserInActivityInterval` (minutes)
- Slider from 1 to 60 minutes.
- Keypress or mousepress counts as activity. Mouse movement does not.

PR #1822, opened 2018-12-14
PasswordDialog: add CAPSLOCK checking

### LINKED ISSUE #881: [UX] Caps lock warning on wallet password window
Show a warning that caps lock is enabled on wallet password window.

mmbyday:
Fixes #881 
Improving user experience / usability.

paass

### DIFF
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -107,3 +107,22 @@ function filterNodes(nodes, port) {
 function epoch(){
     return Math.floor((new Date).getTime()/1000);
 }
+
+function isAlpha(letter){ return letter.match(/^[A-Za-z0-9]+$/) !== null; }
+
+function isLowerCaseChar(letter){ return letter === letter.toLowerCase(); }
+
+function isUpperLock(shift, letter){
+    if(!isAlpha((letter))) return false;
+    if(shift) {
+        if(isLowerCaseChar(letter))
+            return true;
+        else
+            return false;
+    } else {
+        if(isLowerCaseChar(letter))
+            return false;
+        else
+            return true;
+    }
+}
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -34,6 +34,7 @@ import QtQuick.Controls.Styles 1.4
 import QtQuick.Window 2.0
 
 import "../components" as MoneroComponents
+import "../js/Utils.js" as Utils
 
 Item {
     id: root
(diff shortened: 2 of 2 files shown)


selsta:
I like the idea but I wonder if there’s an easier way to add this.

js/Utils.js
```diff
@@ -48,4 +48,42 @@ function showSeedPage() {
...
+function isalpha(letter){
+    var alpha = String("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ");
+    var i;
+    for(i in alpha )
+        if(letter == alpha[i])
+            return true;
+    return false;
+}
+
+function isLowerCaseChar(letter){
```
xmrdsc: `function isLowerCaseChar(letter){ return letter === letter.toLowerCase(); }`

xmrdsc:
Forget my review, I suggest a different approach.

Add a keyfilter here:

https://github.com/monero-project/monero-gui/blob/773c8370d8cdc2d6a17c7c493dadd99f924a42c2/filter.cpp#L40-L43

Emit a new signal here:

https://github.com/monero-project/monero-gui/blob/773c8370d8cdc2d6a17c7c493dadd99f924a42c2/main.cpp#L325

Call the filter 'capslockToggle'.

Listen for it in `main.qml` and set an appWindow 'global' variable:

```js
function capslockToggle(state) {
    appWindow.capslockToggled = state;
}
```

Then in password dialog you can use `appWindow.capslockToggled`.

Superior solution because it would indicate capslock usage without having to first enter a character.

mmbyday:
Good idea, will wait for #1814 to be merged first before implementing this superior event/signal based approach ;-)

xmrdsc:
Do you want to give it a shot? I can pick this up too if you want :)

mmbyday:
@xmrdc I tried, but will need your help. Continue here, https://github.com/mmbyday/monero-gui/pull/2 , or start fresh :)

xmrdsc:
I looked into this. I got the signals to work, however, detecting capslock status is quite a pain to make in a cross-platform manner. For example, on OSX a capslock press is exactly that, a press - it and stays pressed until you click it again, at which point capslock releases - you'll end up with only 1 keydown event. Thus I am in favor of rebasing this ;)

js/Utils.js
mmbyday: Awesome 👍

js/Utils.js
```diff
@@ -48,4 +48,42 @@ function showSeedPage() {
...
     updateBalance();
-}
\ No newline at end of file
+}
+
+/*******************
+   CheckCapsLocks
+   author qyvlik email qyvlik@qq.com https://github.com/qyvlik/CheckCapsLock.qml/blob/master/CheckCapsLock.qml
+*******************/
+function isalpha(letter){
```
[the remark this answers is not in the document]
mmbyday: Awesome 👍 👍

mmbyday:
Rebased.

PR #1824, opened 2018-12-14
Transfer: fix amount label boldness

mmbyday:
Before
bef12

After
after234234


mmbyday:
Ok, should work now :) Lineedit and Lineeditmulti labels now have the same parameters.

PR #1826, opened 2018-12-14
Sign: redesign page with message/file toggle

### LINKED ISSUE #994: Add radio-buttons to sign/verify screen
**Problem**:
The sign/verify part is either/or in terms message/file -> the Signature generated and shown is either from the message or from the file. The GUI does not really reflect this logic, both options can be filled and the signature is created only from the one
image

**Solution**:
Add radio-buttons which enable to select only one of the candidate for the sign/verify process to make it cleared for which the process is applying

mmbyday:
Fixes #994
Before
old

After
test

### DIFF
--- a/pages/Sign.qml
+++ b/pages/Sign.qml
@@ -32,12 +32,14 @@ import QtQuick.Controls.Styles 1.4
 import QtQuick.Layouts 1.1
 import QtQuick.Dialogs 1.2
 
-import "../components"
 import moneroComponents.Clipboard 1.0
 import moneroComponents.WalletManager 1.0
+import "../components" as MoneroComponents
 
 Rectangle {
     id: mainLayout
+    property bool messageMode: true
+    property bool fileMode: false
 
     color: "transparent"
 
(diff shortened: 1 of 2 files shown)


xiphon:
> For fun

Makes sense. I was wondering if some malware infected your PC and is doing this ;-)

> If gitmoji's aren't welcome

Not that i'm strongly against them, but they seem to be useless. I'm not aware of any software that recognizes emoji in commit messages, just Github web site.

xmrdsc:
https://github.com/monero-project/monero-gui/blob/838ca44a6c6ea2d6ce85444dff1a6c345de7c3dc/pages/Sign.qml#L183

`qrc:///pages/Sign.qml:307: ReferenceError: MoneroComponents is not defined`

Fix with:

`import "../components" as MoneroComponents`, then make sure all components start with `MoneroComponents.`

In addition, top margin is too small, i.e: compare this page with other pages. The content starts too early.

mmbyday:
- [x] import "../components" as MoneroComponents, then make sure all components start with MoneroComponents

- [x] In addition, top margin is too small, i.e: compare this page with other pages. The content starts too early.

- [x] Lastly; I think the spacings around mode selection is awkward. Big 'Mode' text on the left, big buttons on the right side. Could probably improve here in terms of design.

- [x] Nice PR!

@xmrdc great eye, great recommendations. all included and improved :-) thanks!!

new

pages/Sign.qml
```diff
@@ -84,231 +86,342 @@ Rectangle {
...
+            radius: 2 * scaleRatio
+            border.color: Qt.rgba(255, 255, 255, 0.25)
+            border.width: 1 * scaleRatio
+            color: "transparent"
 
-            Label {
-                id: signTitleLabel
-                fontSize: 24 * scaleRatio
-                text: qsTr("Sign") + translationManager.emptyString
+            GridLayout {
```
xiphon:
```suggestion
            RowLayout {
```

pages/Sign.qml
```diff
@@ -84,231 +86,342 @@ Rectangle {
...
+                Layout.fillWidth: true
+                Layout.preferredHeight: instructionsText.height + 40
```
xiphon:
```suggestion
                 Layout.preferredHeight: instructionsText.height + 40 * scaleRatio
```

pages/Sign.qml
```diff
@@ -84,231 +86,342 @@ Rectangle {
...
 
-        // sign
-        ColumnLayout {
-            id: signBox
-            anchors.left: parent.left
-            anchors.right: parent.right
-            spacing: 20 * scaleRatio
+        Rectangle {
+            id: instructionsRect
+            Layout.preferredHeight: instructionsText.height + 26
```
xiphon:
```suggestion
            Layout.preferredHeight: instructionsText.height + 26 * scaleRatio
```

pages/Sign.qml
```diff
@@ -84,231 +86,342 @@ Rectangle {
...
-                        labelText: qsTr("Address") + translationManager.emptyString;
-                        addressValidation: true
-                        anchors.topMargin: 5 * scaleRatio
-                        placeholderText: "4.. / 8.."
+            MoneroComponents.LineEditMulti {
+                id: verifyAddressLine
+                Layout.fillWidth: true
+                labelText: qsTr("Address") + translationManager.emptyString;
+                addressValidation: true
+                placeholderText: "Enter the Monero Address (example: 44AFFq5kSiGBoZ...)"
```
xiphon: qsTr

mmbyday:
Thank you for your review suggestions @xiphon 👍  Updated.

Issue #1159, opened 2018-03-05
Open a wallet from file displays Settings page in the background


dEBRUYNE-1:
Open a wallet from file -> It will display the settings page of the GUI in the background.

skftn:
+bug

xmrdsc:
@mmbyday mentioned:
> Whatever page, such as the setting page, was open previously, will show in the background. 

### Quick solution:

Programmatically switch to the transfer page before closing a wallet (and call [completeTransition()](http://doc.qt.io/qt-5/qml-qtquick-controls-stackview.html#completeTransition-method) to prevent stacktransition).

### Better solution:

Find a way to properly construct and deconstruct QML components by utilizing `Component.onCompleted {` so a cleanup of text contents would happen automatically.

Just thinking out loud :-)

mmbyday:
I made a small, simple change. That'll fix this for now. #1837 ,

dEBRUYNE-1:
Looks good. I'll close this issue when the PR gets merged.

dEBRUYNE-1:
+resolved

PR #1832, opened 2018-12-17
main: hide Unlock time when switching wallets

### LINKED ISSUE #1403: Unlock time is Visible when switch to wallet
Windows x64 GUI Wallet 0.12.0

When staying in-app and you close one wallet and open another, if that wallet has recently received a transaction, it will show the balance unlock time prior wallet unlock to successful wallet password.

monero wallet balance unlock time visible circled

Does not appear to be a serious issue, however could raise concerns regarding privacy.

mmbyday:
resolves #1403

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -923,6 +923,7 @@ ApplicationWindow {
 
     // close wallet and show wizard
     function showWizard(){
+        clearMoneroCardLabelText();
         walletInitialized = false;
         closeWallet();
         currentWallet = undefined;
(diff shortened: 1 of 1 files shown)


dEBRUYNE-1:
Does this resolve #1159 too?

mmbyday:
No. Whatever page, such as the setting page, was open previously, is still showing in the background. What kind of behavior would you prefer to happen in #1159?

xmrdsc:
Nice!

dEBRUYNE-1:
@mmbyday - Perhaps simply the wizard page where the user can select the different options (create a new wallet, restore from the 25 word mnemonic seed etc.)?

mmbyday:
@dEBRUYNE-1 Terrific idea. Implemented by #1837 :)

dEBRUYNE-1:
@mmbyday - Nice one :+1:

PR #1833, opened 2018-12-17
SettingNode: Create explicit warnings and information when connecting to remote node

### LINKED ISSUE #1085: Create explicit warnings and information when connecting to remote node
Do some kind of checking to see if the IP isn't 127.0.0.1 and warn about risks.

mmbyday:
resolves #1085

- appends text, minimizes translation efforts :)

warnnn

### DIFF
--- a/pages/settings/SettingsNode.qml
+++ b/pages/settings/SettingsNode.qml
@@ -276,7 +276,7 @@ Rectangle{
             MoneroComponents.WarningBox {
                 Layout.topMargin: 26 * scaleRatio
                 Layout.bottomMargin: 6 * scaleRatio
-                text: qsTr("To find a remote node, type 'Monero remote node' into your favorite search engine. Please ensure the node is run by a trusted third-party.") + translationManager.emptyString
+                text: qsTr("To find a remote node, type 'Monero remote node' into your favorite search engine. Please ensure the node is run by a trusted third-party.") + "\<br><br>" + qsTr("WARNING: Using a remote node means you are trusting someone elses copy of the blockchain. To have a fully trustless interaction with the Monero network, you must synchronize your own node. It is highly recommended to run your own node.") + translationManager.emptyString
             }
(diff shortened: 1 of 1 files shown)


xiphon:
Consider to mention metadata leak as well: user's identity with corresponding transactions ids sent through remote node connections.

erciccione:
@xiphon I would add only a minimal amount of informations in the warning. Then users can make their own researches. FYI the guide embedded with the archives will be more verbose about this (monero-ecosystem/monero-GUI-guide#103)

xiphon:
> @xiphon I would add only a minimal amount of informations in the warning

Yep, the point is that using a remote node may weaken user's privacy under some circumstances. Not just blockchain contents trust, but overall.

I guess the warning may be even shorter and more general than it is now. Or/and contain a reference to the guide.

Any thoughts?

erciccione:
I guess the warning could me more specific then. Something like `WARNING: Using a remote node means you are trusting someone elses copy of the blockchain, this could weaken your privacy (Check LINK-TO-GUIDE for more info). It is highly recommended to run your own node."`. It would also be shorter.

About linking to the guide, there was this idea of using tooltips or something like that to help the user, but it wasn't further discussed. I plan to make the guide available online in HTML format, that would make easier to link to it. Anyway, i think linking to the guide could be a good solution (#1602 is still open waiting for a good solution to embed it in the monero-gui repo)

xmrdsc:
NACK.

Current message should suffice if you take into account the wizards already warn for this. If they do not, that's the place for more elaborate messages. However, if consensus is such that we should warn the user with at every step of the way, I suggest to make this one a bit shorter.

xmrdsc:
The new wizards will show the relevant messages regarding remote nodes. I think this is the place for elaborate warnings. I suggest we close this.

xmrdsc:
+invalid

PR #1836, opened 2018-12-17
receive: add "Download QR-Code image" and "Copy Payment URL" buttons

xiphon:
receive-buttons


selsta:
The only discussion that remains now is what to show by default and what to hide under advanced.

xiphon:
> The only discussion that remains now is what to show by default and what to hide under advanced.

Yep, but should be done in a separate PR

xiphon:
> Did you also self design these? @2x would be super nice :)

Nope, these icons are from https://useiconic.com/open `data-transfer-download` and `external-link`

xiphon:
To be merged after #1793

selsta:
@xiphon can you rebase this?

xiphon:
Updated
receive-buttons

PR #1838, opened 2018-12-17
build: fixes build with Trezor, requires #4945, on current monero master

ph4r05:
Fixes compilation with the Trezor enabled with the current monero master.

Requires https://github.com/monero-project/monero/pull/4945 which is not yet merged. 

To disable Trezor compilation run:

```bash
BUILD_TREZOR=false ./build.sh 
```


ph4r05:
I've tested it in practice only on OSX. It would be great if somebody can test it on win, linux, android platforms.

PR #1818, opened 2018-12-13
SettingsInfo: use MoneroComponents.StandardButton :lipstick:

mmbyday:

PR #1840, opened 2018-12-18
settings-info: use StandardButton instead of custom button

xmrdsc:
#1817


xmrdsc:
Whoops, duplicate of #1818

PR #1841, opened 2018-12-18
images: add @2x high DPI versions

selsta:


xmrdsc:
Very nice. Some context on what this PR does:

> On macOS and iOS, you can provide alternative resources with double the size and the @2x identifier for icons and images and place them in the resource file. On Retina displays, the @2x versions are used automatically.

Reviewed.

PR #1845, opened 2018-12-19
MiddlePanel: fix scrollbar margin top/bottom

selsta:
Before:
screenshot 2018-12-19 at 15 33 23
After:
screenshot 2018-12-19 at 15 34 30


selsta:
Force-pushed so that it works well with #1846

PR #1849, opened 2018-12-19
openalias: handle address validation correctly

mmbyday:
before
beforeope

after
afteddd


js/TxUtils.js
```diff
@@ -56,3 +56,11 @@ function checkSignature(signature) {
     return false;
 }
 
+function isValidOpenAliasAddress(address) {
+  address = address.trim()
```
xmrdsc: Should be 4 space indentation

PR #1854, opened 2018-12-21
CheckBox: iconOnTheLeft property, optional uncheckedIcon and border, rich text format

xiphon:
* Property to control icon alignment (left/right)
* Optional `uncheckedIcon` support (different icons for checked and unchecked states)
* Option to hide checkbox border

### DIFF
--- a/components/CheckBox.qml
+++ b/components/CheckBox.qml
@@ -38,8 +38,10 @@ Item {
     property string uncheckedIcon
     property bool checked: false
     property alias background: backgroundRect.color
+    property bool border: true
     property int fontSize: 14 * scaleRatio
     property alias fontColor: label.color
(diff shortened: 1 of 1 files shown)


components/CheckBox.qml
```diff
@@ -51,6 +53,8 @@ Item {
 
     RowLayout {
         id: checkBoxLayout
+        layoutDirection: iconOnTheLeft ? Qt.LeftToRight : Qt.RightToLeft
+        spacing: !isMobile ? 10 : 8
```
mmbyday:
```suggestion
        spacing: (!isMobile ? 10 : 8) * scaleRatio
```

xiphon: Fixed, thanks

PR #1855, opened 2018-12-21
Transfer: hide PaymentId and Description by default

xiphon:
Requires #1853 #1854 

Feedback and comments on proposed change are welcome.

transfer-hide-payment-id


rbrunner7:
Looks reasonable to me. A good compromise between slowly phasing out payment ids but still supporting them for the time being.

selsta:
I’m like this feature but I’m still unhappy abhout how Payment IDs is now not aligned with the other labels.

Maybe it’s possible to add the + to the left of the Payment ID label and align it like it was before.

SamsungGalaxyPlayer:
We go even further and say (optional, will be deprecated March/April 2019). See Monero [issue 3772](https://github.com/monero-project/monero/issues/3772).

Of course, this is contingent on agreeing to remove them in the next upgrade.

xiphon:
Updated

xiphon:
Updated to hide Description field too.

dEBRUYNE-1:
I'd remove the `will be deprecated March/April 2019` part for now, as there's no consensus yet on when the long, unencrypted payment ID will be deprecated.

xiphon:
> I'd remove the `will be deprecated March/April 2019` part for now, as there's no consensus yet on when the long, unencrypted payment ID will be deprecated.

Yep. Updated the PR.

PR #1866, opened 2018-12-27
Transfer: disable long payment ids by default

selsta:
monero-project/monero#5020

Not sure how this should be implemented, this is just an idea.

Default looks like this: 

screenshot 2018-12-27 at 19 55 52

Can be re-enabled in settings.

screenshot 2018-12-27 at 19 55 48


selsta:
Right, forgot about the address book. Should I hide it the same way as on the transfer page?

mmbyday:
> Right, forgot about the address book. Should I hide it the same way as on the transfer page?

Sounds good @selsta.

selsta:
Updated.

PR #1871, opened 2018-12-29
Compact margins for transfer optionals

xmrdsc:
Compact margins for the optional buttons.

@mmbyday I think this looks better. What do you think?

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -293,56 +293,66 @@ Rectangle {
       }
 
       ColumnLayout {
-          CheckBox {
-              id: paymentIdCheckbox
-              border: false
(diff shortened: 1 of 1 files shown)


selsta:
ou2
Is this working as intended?

xmrdsc:
Yes. You dont like that margin change I take it? Personally I dont mind.

selsta:
Yep, I don’t really like it, I think the current margins are fine.

mmbyday:
@xmrdsc nice in thought. although at first impression, before looks more balanced. bigger picture, i think this page can be improved with a layout redesign. amount, destination, send button are primary; everything else gets de-prioritized / tucked away.

PR #1876, opened 2018-12-30
PasswordDialog: cancel button text is more descriptive and dynamic

mmbyday:
cancel

### DIFF
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -191,7 +191,7 @@ Item {
                 MoneroComponents.StandardButton {
                     id: cancelButton
                     small: true
-                    text: qsTr("Cancel") + translationManager.emptyString
+                    text: root.walletName.length > 0 ? qsTr("Change wallet") + translationManager.emptyString : qsTr("Cancel") + translationManager.emptyString
                     KeyNavigation.tab: passwordInput
                     onClicked: {
                         root.close()


selsta:
Just so that I understand it, how can it be possible that root.walletname.length = 0?

dEBRUYNE-1:
What happens if a user clicks the `Change wallet` button? Is the file dialog opened where the user can choose his desired wallet? If so, perhaps we should have three buttons, i.e., a `Cancel` button where a user can go back to the wizard, a `Change wallet` button were the user can open a different existing wallet, and a `Continue` button to open the most recent wallet.

dEBRUYNE-1:
Ping @mmbyday, any opinion on my suggestion?

mmbyday:
@selsta , whenever a wallet is opened, passwordDialog is passed a root.walletName . Only the view seed scenario doesn't pass a walletName so that's why this works.

https://github.com/monero-project/monero-gui/blob/36eb1f80e1c070f753746519e2c9e4f01784f6c2/main.qml#L1011

@dEBRUYNE-1 Change wallet just goes back to the wizard, as the cancel button currently does. For sake of user convenience, I'll make a subsequent PR to skip the language selection screen so the user sees the screen that is most relevant, which is the new wallet/restore wallet/open a wallet from file/create from new hw device.

selsta:
> @selsta , whenever a wallet is opened, passwordDialog is passed a root.walletName . Only the view seed scenario doesn't pass a walletName so that's why this works.

Makes sense, also works together with #1886 :)

> @dEBRUYNE-1 Change wallet just goes back to the wizard, as the cancel button currently does. For sake of user convenience, I'll make a subsequent PR to skip the language selection screen so the user sees the screen that is most relevant, which is the new wallet/restore wallet/open a wallet from file/create from new hw device.

I guess it’s better to wait until after the wizard redesign for this.

PR #1893, opened 2019-01-10
main: skip language page when logging out or changing wallets

mmbyday:
before
eee

after
aft

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -933,6 +933,7 @@ ApplicationWindow {
         closeWallet();
         currentWallet = undefined;
         wizard.restart();
+        wizard.switchPage(true); //skips select language page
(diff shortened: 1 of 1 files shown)


erciccione:
Is there a specific reason for skipping the language page? How will be possible to switch to another language then?

xmrdsc:
Note that the wizards are currently being redone, fyi in case you work on something that would maybe be replaced next week!

xiphon:
> Is there a specific reason for skipping the language page? 

In most cases is a redundant step harming UX.

> How will be possible to switch to another language then?

Use `prev page` button on the left.

xmrdsc:
Fixed in #1909

+duplicate

PR #1894, opened 2019-01-10
merchant: small bug fixes and improvements

selsta:


pages/merchant/Merchant.qml
```diff
@@ -603,7 +603,7 @@ Item {
             var isout = model.data(idx, TransactionHistoryModel.TransactionIsOutRole);
             var timeDate = model.data(idx, TransactionHistoryModel.TransactionDateRole);
             var timeHour = model.data(idx, TransactionHistoryModel.TransactionTimeRole);
-            var timeEpoch = new Date(timeDate + "T" + timeHour + "Z") .getTime() / 1000;
+            var timeEpoch = new Date(timeDate + "T" + timeHour) .getTime() / 1000;
```
mmbyday:
```suggestion
            var timeEpoch = new Date(timeDate + "T" + timeHour).getTime() / 1000;
```

selsta: Already merged :/ I think the space is cosmetics only, I tested both and saw no difference.

PR #1907, opened 2019-01-17
IconButton: fix mousearea and remove animation

selsta:
The animation was buggy if the mouse got released mouse outside of the mouse-area or outside of the whole program. Maybe someone has an idea how this can get fixed instead of removing it?

### DIFF
--- a/components/IconButton.qml
+++ b/components/IconButton.qml
@@ -35,7 +35,6 @@ Item {
 
     signal clicked(var mouse)
 
-
     id: button
     width: parent.height
     height: parent.height
@@ -47,13 +46,13 @@ Item {
         id: buttonImage
         source: ""
         x : (parent.width - width) / 2
-        y : (parent.height - height)  /2
+        y : (parent.height - height) / 2
         z: 100
     }
 
     MouseArea {
         id: buttonArea
-        anchors.fill: parent
+        anchors.fill: buttonImage
         hoverEnabled: true
         cursorShape: Qt.PointingHandCursor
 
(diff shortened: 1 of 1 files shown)


mmbyday:
This will keep the animation intact and fix the release bug.
```
diff --git a/components/IconButton.qml b/components/IconButton.qml
index 801c01e..de82660 100644
--- a/components/IconButton.qml
+++ b/components/IconButton.qml
@@ -35,7 +35,6 @@ Item {

     signal clicked(var mouse)

-
     id: button
     width: parent.height
     height: parent.height
@@ -47,13 +46,13 @@ Item {
         id: buttonImage
         source: ""
         x : (parent.width - width) / 2
-        y : (parent.height - height)  /2
+        y : (parent.height - height) / 2
         z: 100
     }

     MouseArea {
         id: buttonArea
-        anchors.fill: parent
+        anchors.fill: buttonImage
         hoverEnabled: true
         cursorShape: Qt.PointingHandCursor

@@ -61,14 +60,14 @@ Item {
             buttonImage.x = buttonImage.x + 2
             buttonImage.y = buttonImage.y + 2
         }
+
         onReleased: {
-            buttonImage.x = buttonImage.x - 2
-            buttonImage.y = buttonImage.y - 2
+            buttonImage.x = (parent.width - width) / 2
+            buttonImage.y = (parent.height - height) / 2
         }

         onClicked: {
             parent.clicked(mouse)
         }
     }
-
 }
```

selsta:
thanks, updated.

selsta:
@mmbyday Somehow this is still buggy for me :/

components/IconButton.qml
```diff
@@ -47,28 +46,29 @@ Item {
...
-        y : (parent.height - height)  /2
+        y : (parent.height - height) / 2
         z: 100
     }
 
     MouseArea {
         id: buttonArea
-        anchors.fill: parent
+        anchors.fill: buttonImage
+        anchors.margins: -5
```
mmbyday:
```suggestion
```
This was the culprit :b

selsta: That’s to make the clickable area a little big bigger than the icon. :)

PR #1905, opened 2019-01-15
add account support

selsta:
Updated and rebased version of #793 

Known bugs:

- Copied addresses are truncated. (monero-project/monero#5074)
- Account table buttons are kinda buggy. (#1907)

I’d appreciate feedback!

PR #1914, opened 2019-01-22
Receive: update subaddress table design

selsta:
Same design as in #1905. Not sure if the item height is too high now, I’ve used the same design as the address book table from zeplin.io

screenshot 2019-01-22 at 15 51 19


dEBRUYNE-1:
As a side note, perhaps the orange buttons below the QR code should be centered?

mmbyday:
@dEBRUYNE-1 done in #1915 :)

@selsta Looks good for design consistency. Agree with item height which could be lower. Too much white/black space per row.

GBKS:
monero-expandable-addresses-gbks-190128

How about using expandable rows like we do in the transactions screen? The problem I see with the static detail display as shown in the GIF in #1915 is that it gets pushed down if you have a lot of addresses. With expandable rows, you avoid this.

SamsungGalaxyPlayer:
@GBKS that's one way to handle it. Another option is to simply only scroll the subaddresses, not the QR code too. So the top scrollbar would allow you to go through the different subaddresses while the QR code stayed in the same place.

I have no particular opinion on which is better, I'll leave it up to the great UX team :)

GBKS:
@SamsungGalaxyPlayer we also had a chat about this on IRC yesterday. @xmrdsc pointed out we already do what you mention on the [Receive screen](https://i.imgur.com/toHKGjb.png).

I find it a bit cluttered, but both approaches should work well on desktop. If the wallet needs to work on mobile, we might have to adjust in one way or another since we'll be very tight space-wise. But we can probably cross that bridge when we get to it.

Issue #1903, opened 2019-01-15
Accounts support - request for improvement


kayront:
(breaking down #1864 as requested)

Previous work seems to have happened here - #763 

See last/final paragraphs of #1864 for commentary on why this would be a worthy addition in my opinion.

xmrdsc:
Support for accounts was recently added.

+resolved

PR #1915, opened 2019-01-22
receive: ux/ui improvements

mmbyday:
rrr

- current address is more apparent in orange color
- full address visible
- hovering over button describes a button's action
- qr code size is down a notch, more balanced imo.

Your thoughts?

### DIFF
--- a/components/LineEditMulti.qml
+++ b/components/LineEditMulti.qml
@@ -38,6 +38,7 @@ ColumnLayout {
     Layout.preferredHeight: childrenRect.height
 
     property alias text: input.text
+    property alias textHorizontalAlignment: input.horizontalAlignment
     property alias labelText: inputLabel.text
     property alias labelButtonText: labelButton.text
     property alias placeholderText: placeholderLabel.text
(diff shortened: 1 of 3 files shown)


selsta:
I’m in favour of a smaller QR code and centred buttons.

Against full address visible, as we removed the full address some time ago to make this page less cluttered :)
Regarding orange color, it would be best if this could wait as I still experiment with the table design in both #1914 and #1903.

maogo:
full address visible+1

xiphon:
Don't center the address. This way it looks like it consists of 2 separate entries `51y ... PEH4` and `PSZ ... Q6f`

mmbyday:
Thanks for the feedback. Address is no longer centered.

tficharmers:
Now that 'Accounts' have been merged, should it be more obvious which account these receive addresses relate too? Also, like on the 'Send' page, should there be a link/button to quickly switch accounts?

selsta:
Thanks for this PR mmbyday but I’ll close this, Receive page changed a lot since this PR and I personally didn’t like the look of these changes.

selsta:
+close

selsta:
+resolved

PR #1916, opened 2019-01-24
Bring back paymentID to tranfer page and force hide paymentId on a settings toggle

xmrdsc:
This was reported by charuto on https://git.pwned.systems/monero-project/monero-gui/issues/1

> If payment ID option is expanded, upon removing the payment ID setting, it will remain on the send page.
> Tested on master branch.

**Edit**: Brought back payment id to the transfer page, albeit collapsed by default, as was decided in the [dev meeting](https://www.reddit.com/r/Monero/comments/akmf5q/logs_of_yesterdays_dev_meeting/).

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -300,7 +300,7 @@ Rectangle {
               fontSize: paymentIdLine.labelFontSize
               iconOnTheLeft: false
               Layout.fillWidth: true
-              text: qsTr("Payment ID <font size='2'>( Optional )</font>") + translationManager.emptyString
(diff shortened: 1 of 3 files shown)


pages/Transfer.qml
```diff
@@ -290,7 +290,7 @@ Rectangle {
       }
 
       ColumnLayout {
-          visible: appWindow.persistentSettings.showPid || paymentIdCheckbox.checked 
+          visible: appWindow.persistentSettings.showPid
```
xiphon: Don't have to change this

xmrdsc: :+1:

PR #1920, opened 2019-01-29
Dynamic --max-concurrency, max 6

xmrdsc:
`--max-concurrency` based on 50% of threads available with a max of `6`. ~In addition, `--db-sync-mode` is forced on `safe`. This guards against corruption on unclean shutdowns.~

I'm expecting this PR to solve some of the "my machine is slow while syncing" complaints.

### DIFF
--- a/src/daemon/DaemonManager.cpp
+++ b/src/daemon/DaemonManager.cpp
@@ -1,5 +1,6 @@
 #include "DaemonManager.h"
 #include <QFile>
+#include <QThread>
 #include <QFileInfo>
 #include <QDir>
 #include <QDebug>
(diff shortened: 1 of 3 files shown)


xmrdsc:
```
< moneromooo> I'd say don't touch db sync mode. If the defaults aren't good, they should be tuned upstream.
< moneromooo> max concurrency is set to (IIRC) the number of hw threads. You may want to set it to half that, or whatever.
 < moneromooo> Maybe default to half threads, and have an option to drive it up.
```

src/daemon/DaemonManager.cpp
```diff
@@ -72,7 +73,14 @@ bool DaemonManager::start(const QString &flags, NetworkType::Type nettype, const
 
     arguments << "--check-updates" << "disabled";
 
+    // --max-concurrency based on threads available. max: 6
+    int32_t concurrency = QThread::idealThreadCount() / 2;
+    if(concurrency > 6)
```
selsta: You could use qBound() here, otherwise looking good.

xmrdsc: Thanks, fixed.

PR #1921, opened 2019-01-29
History: omit redundant 'Cancel' button in confirmation dialog

### LINKED ISSUE #1919: Redundant 'Cancel' button in CSV Export confirmation screen
I think just the 'OK' button is required here? Both buttons seem to just dismiss the screen.

screen shot 2019-01-29 at 11 53 34

xiphon:
Fixed #1919

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -384,17 +384,18 @@ Rectangle {
             var written = currentWallet.history.writeCSV(currentWallet.currentSubaddressAccount, dataDir);
 
             if(written !== ""){
-                confirmationDialog.title = qsTr("Success") + translationManager.emptyString;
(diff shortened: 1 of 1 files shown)


components/StandardDialog.qml
```diff
@@ -152,6 +152,7 @@ Rectangle {
             MoneroComponents.StandardButton {
                 id: cancelButton
                 text: qsTr("Cancel") + translationManager.emptyString
+                visible: root.cancelVisible
```
xmrdsc: Doesn't it loop like this? `root.cancelVisible` is an alias to this line.

xiphon: Yep, it does. Fixed

PR #1922, opened 2019-01-30
Fix daemon flags

xmrdsc:
I've discovered that the daemon flags never get saved to `persistentSettings`. This PR fixes that. In addition, I've made the input box multiline.

### DIFF
--- a/pages/settings/SettingsNode.qml
+++ b/pages/settings/SettingsNode.qml
@@ -362,6 +362,7 @@ Rectangle{
                     if (appWindow.daemonRunning) {
                         appWindow.stopDaemon();
                     } else {
+                        persistentSettings.daemonFlags = daemonFlags.text;
                         appWindow.startDaemon(persistentSettings.daemonFlags);
                     }
                 }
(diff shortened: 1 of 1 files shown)


pages/settings/SettingsNode.qml
```diff
@@ -362,6 +362,7 @@ Rectangle{
                     if (appWindow.daemonRunning) {
                         appWindow.stopDaemon();
                     } else {
+                        persistentSettings.daemonFlags = daemonFlags.text;
```
selsta: Is this required? `onEditingFinished: persistentSettings.daemonFlags = daemonFlags.text;` should do this anyway. I don’t know if the signal gets emitted before the daemon starts though.

selsta: @mmbyday makes sense :) thanks

xmrdsc: @selsta Unfortunately, signal `onEditingFinished` is only emitted when user leaves the input. In the situation that the user is still focused (editing) the input box and clicks `Start Daemon`, the event is not emitted :(

PR #1931, opened 2019-02-02
warningbox: make text readonly

mmbyday:
Fixes this from happening.

sda


xmrdsc:
Could you explain the fix?

PR #1939, opened 2019-02-15
settingsnode: enable trusted daemon option

### LINKED ISSUE #1938: Monero GUI: Trusted Daemon?
https://www.reddit.com/r/Monero/comments/aquwbj/monero_gui_trusted_daemon/

mmbyday:
fixes #1938 
tttt

### DIFF
--- a/pages/settings/SettingsNode.qml
+++ b/pages/settings/SettingsNode.qml
@@ -298,7 +298,12 @@ Rectangle{
                 daemonPortText: rna.search(":") != -1 ? (rna.split(":")[1].trim() == "") ? "18081" : rna.split(":")[1] : ""
                 onEditingFinished: {
                     persistentSettings.remoteNodeAddress = remoteNodeEdit.getAddress();
-                    console.log("setting remote node to " + persistentSettings.remoteNodeAddress)
+                    console.log("setting remote node to " + persistentSettings.remoteNodeAddress);
+                    if (persistentSettings.is_trusted_daemon) {
+                        persistentSettings.is_trusted_daemon = !persistentSettings.is_trusted_daemon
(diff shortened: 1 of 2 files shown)


pages/settings/SettingsNode.qml
```diff
@@ -330,6 +330,16 @@ Rectangle{
...
             }
 
+            MoneroComponents.CheckBox {
+                id: setTrustedDaemonCheckBox
+                checked: persistentSettings.is_trusted_daemon
+                onClicked: {
+                    persistentSettings.is_trusted_daemon = !persistentSettings.is_trusted_daemon
+                    currentWallet.setTrustedDaemon(persistentSettings.is_trusted_daemon)
+                }
+                text: qsTr("Enable as Trusted Daemon") + translationManager.emptyString
```
xmrdsc: `Mark as trusted daemon`

xmrdsc:
The option should probably reset itself when the remote node changes.

pages/settings/SettingsNode.qml
mmbyday: I dig it 👍

mmbyday:
> The option should probably reset itself when the remote node changes.

Good idea, now implemented.

A status message will flash the following
Remote node updated. Trusted daemon has been reset. Mark again, if desired.

Setzler22:
This doesn't appear to be working on the Linux GUI client. No matter what I do it still doesn't let me rescan the wallet balance

Setzler22:
Actually, it's even worse. There have been times that it is unchecked and it is acting like a trusted node, and there are times that it is checked and it is not. In other words, the check box on the GUI has no connection to whether the node is actually trusted or not.

This is a serious security issue.

Issue #1940, opened 2019-02-17
Remote node port autofills with default 18081 even if not mainnet


binaryFate:
In Settings -> Node, when using a remote node and entering its information, the `port` input field autofills to "18081" even if wallet is a stagenet one.

It should autofill to correct default port for mainnet / stagenet / testnet, depending what type of wallet is open.

PR #1943, opened 2019-02-17
Fix default daemon/remote node RPC port based on the network type

xiphon:
Fixes https://github.com/monero-project/monero-gui/issues/1940

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -78,7 +78,7 @@ ApplicationWindow {
     property bool androidCloseTapped: false;
     property int userLastActive;  // epoch
     // Default daemon addresses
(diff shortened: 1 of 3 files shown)


main.qml
```diff
@@ -1891,6 +1891,17 @@ ApplicationWindow {
         passwordDialog.open();
     }
 
+    function getDefaultDaemonRpcPort(networkType) {
```
xmrdsc: Probably best to put this function in `js/Utils.js` and use it via `Utils.getDefaultDaemonRpcPort()`. You will need `import "../js/Utils.js" as Utils`

PR #1946, opened 2019-02-23
Ease the pain of people that don't recursively clone

NewEraCracker:
- Update instructions to recursively clone
- Combines submodule init and update into one command and does so recursively


xiphon:
We don't need it, everything clones and builds well without `--recursive`.

dEBRUYNE-1:
@NewEraCracker - As @xiphon says, everything works properly without this PR. I am thus going to close this. 

+resolved

PR #1948, opened 2019-02-23
wizard: add flags to language sidebar

mmbyday:
eeee

Better with flags, like before.


xmrdsc:
This is very nice, but we're having a lot of problems with the sidebar in combination with the build bot. Sidebar might need to be removed and replaced with a dedicated screen (seems to be a Qt bug).

Test out a buildbot binary to see the bug.

xiphon:
Merged via https://github.com/monero-project/monero-gui/pull/2848

PR #1952, opened 2019-02-24
StandardButton: fix button appearance on Qt 5.7

xiphon:


xmrdsc:
Tested, works on buildbot. Very nice.

PR #1960, opened 2019-02-25
transfer: make simple mode more simple

mmbyday:
Less is more in simple mode. 

Only the absolute required fields should be visible. Compared to other beginner-friendly wallets, only the essentials remain. Those are amount and address. Currently, transfer page in simple has two extra optional touch points causing undue user thoughts and doubt. Removed transaction priority and description based on some user's feedback. Thoughts?

After
feed

Before

afttt

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -127,7 +127,7 @@ Rectangle {
       }
 
       GridLayout {
-          columns: (isMobile)? 1 : 2
+          columns: (isMobile || !(appWindow.walletMode >= 2)) ? 1 : 2
           Layout.fillWidth: true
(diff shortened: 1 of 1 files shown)


xmrdsc:
I'd argue description has a place in simple mode as simple mode also includes the tx history page - on which it is nice to view transactions accompanied with their respective description - even more so for beginners.

xmrdsc:
What do others thinks? :-)

selsta:
I generally agree with this PR.

mmbyday:
Fair points all around. So if I understand correctly, the transaction priority should be dropped faster than a misbehaving node connection. 

Here's what remains.
faaa

PR #1961, opened 2019-02-25
Dont show message on simple mode disconnect/reconnect

xmrdsc:
I've noticed on Windows 10, it will show this warning message while connecting to a remote node, regardless if it is (eventually) successful or not. This is bad because users will think something is up, while it's working.

I loosely assume this is due to the Javascript code that does the XHR request to the node aggregator, but I don't know for sure. This behaviour has not happened to me on Linux. sgp observed the same problem.

For now, it's best to replace this message with `console.log()`. We can have decent status messages when xiphon's node magic gets merged for 14.1.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1791,7 +1791,7 @@ ApplicationWindow {
                 appWindow.disconnectedEpoch = 0;
                 return;
             }, function(){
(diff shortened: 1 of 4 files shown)


xmrdsc:
Exactly, I think this is due to how these XHR (HTTP) requests sometimes fail or otherwise return prematurely, tricking the GUI into thinking the request was unsuccessful, however when the result callback eventually fires, it will just connect.

^ speculation.

PR #1965, opened 2019-02-25
transfer: messaging for disabled send button

### LINKED ISSUE #1795: Display message on why send button is disabled
When sending monero to multiple addresses, it is easy to encounter an unlocked balance of zero. However, the reason why isn't apparent when trying to click the send button (nothing happens). Maybe a hint can be given when the send button is clicked on why it is disabled?

This may be related to #67...

mmbyday:
Fixes #1795
erroror

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -48,6 +48,7 @@ Rectangle {
     color: "transparent"
     property int mixin: 10  // (ring size 11)
     property string warningContent: ""
+    property string sendButtonWarning: ""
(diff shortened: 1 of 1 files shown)


selsta:
Nice idea, I think we should wait for v0.14.1.0 with this change as it’d introduce new strings and translations for v0.14.0.0 did already start.

mmbyday:
Agreed. No rush. Let the translators focus on the current monumental task.

PR #1985, opened 2019-02-28
wizard: allow restore height when creating wallet from device

### LINKED ISSUE #1984: Why can I not choose a restore date when restoring from Ledger, but only when restoring from seed?
https://www.reddit.com/r/Monero/comments/avqcwp/gui_v01400_prerelease_test_binaries_buildbot/ehho3k6

mmbyday:
Fixes #1984

### DIFF
--- a/wizard/WizardCreateDevice1.qml
+++ b/wizard/WizardCreateDevice1.qml
@@ -89,12 +89,14 @@ Rectangle {
                 MoneroComponents.LineEdit {
                     id: restoreHeight
                     Layout.fillWidth: true
-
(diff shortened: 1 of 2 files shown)


dEBRUYNE-1:
As this creates new strings, probably best to include this in the point release?

mmbyday:
I'd push it through. The cost of not doing so does not outweigh a few missing translations.

dEBRUYNE-1:
@mmbyday - Just spoke to @erciccione and it should not really be a problem for the translations. Thus, let's add this PR.

PR #1986, opened 2019-02-28
Translations for 0.14.0

Monero-Pootle:
Translations coming from Pootle: translate.getmonero.org

The list of translators can be found on Pootle itself.

About 2/3 of the languages have been updated, some have been completed, some still need a lot of work, but since the limited amount of time we had, i consider it a big win. To be more specific:

- 18 full translated languages
- 3 around 80% complete
- 10 around 60-70% complete

Detailed statistics on Pootle

### DIFF
--- a/translations/monero-core.ts
+++ b/translations/monero-core.ts
@@ -2,77 +2,162 @@
 <!DOCTYPE TS>
 <TS version="2.1">
 <context>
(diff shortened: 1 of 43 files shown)


Monero-Pootle:
Rebased on #1985  + resync

PR #1994, opened 2019-03-04
main: enable slow-hash JIT

selsta:
```
18:31 <+moneromooo> dEBRUYNE: can you call putenv in the GUI before running monerod ?
18:31 <+moneromooo> ie, forcing JIT on.
18:32 <+moneromooo> putenv("MONERO_USE_CN_JIT=1");
```

Doesn’t seem to work for me though on macOS. ~114H/s with and without this patch.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -88,7 +88,8 @@ int main(int argc, char *argv[])
 #ifdef Q_OS_WIN
     bool isWindows = true;
 #endif
-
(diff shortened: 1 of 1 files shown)


mmbyday:
Tested. No change to hashrate on Windows.

selsta:
https://github.com/monero-project/monero/pull/5213

PR #1996, opened 2019-03-07
Fix simple mode (boostrap) translation string

MaxXor:


xmrdsc:
I did this on purpose to simplify terminology and not impose the task of having to translate 'bootstrap' in all languages.

Issue #1738, opened 2018-11-12
Error opening wallet in Windows GUI


G-Bigo:
The Windows x64 GUI wallet contains a massive bug. It is impossible to re-open a wallet.

Situation: 
• Open program first time
• Choose language (German for me)
• Create a wallet
• Choose a password (in my case containing spaces)
• Start service
• Sync blockchain
  (Everything standard up to here)
• Close program
• Re-open program
• Enter password
• Wallet couldn't be opened: basic_string::_M_replace_aus

The same happens when trying to open the (definitely existing) wallet from a file.

When restarting the whole process and creating a new wallet the wallet specs show a possible reason:
The alleged location (on a windows system!) alway ends with something like (without quotation marks)
'/username/username'
I.e. the windows path contains slashes (instead of backslashes), which of course must evoke an error. Seems like some Linux code splashed over unaltered.

I don't know whether the slashes within the wallet file path actually are the reason for the described bug. Anyway, the procedure described above reproducible yields the wallet open error.

G-Bigo:
I first run into this issue with version 0.13.0.3, then downloaded 0.13.0.4 to see whether it works. And no, nothing is fixed in v0.13.0.4 — exactly the 'Can't open wallet' error described above.

dEBRUYNE-1:
@G-Bigo - You will, after upgrading to GUI v0.13.0.4, still have to remove the wallet cache once. Thereafter it should work properly though.

dEBRUYNE-1:
Ping @G-Bigo - Did you manage to resolve your issue?

G-Bigo:
Yepp, deleting old wallet, restoring (aka recreating) it with mnemonic phrase, re-synching wallet. No it works as expected. Thanks.

erciccione:
+resolved

Issue #1730, opened 2018-11-07
SOLVED  Couldn't open wallet: std::bad_alloc  Monero Gui v0.13.0.4 Beryllium Bullet and Windows 10


fouadr:
Hi there,

Just downloaded, extracted and opened Monero Gui v0.13.0.4 Beryllium Bullet in Windows 10, after opening I used my seed to restore the wallet ( to create new wallet/cache files) and this went fine.
Once the wallet was synchronised I closed the Monero Wallet Gui and a few minutes later I tried to open the wallet file, right after providing my passphrase the dreaded isse that I was hoping not to see appered:

"Couldn't open wallet: std::bad_alloc" 

Are these builds not tested on the different platforms for where they are build for??
Guys, after opening the wallet also close and open it again after build!

And yes, deleting the cache file is a work around, just wait a long time to download the whole blockchain again... but this is not a desired situation.

dEBRUYNE-1:
Are you absolutely certain you downloaded GUI v0.13.0.4? Because that particular issue has been resolved in v0.13.0.4 (you may have to delete the wallet cache once though). If not, please use the direct download link from here:

https://reddit.com/r/Monero/comments/9ti2on/gui_v01304_beryllium_bullet_released/

fouadr:
Thanks, I'll try to download again from the official Monero website ...

fouadr:
Thanks, I'll try that..

fouadr:
I had downloaded the right version ( build date of monero-wallet-gui is:  ‎31 ‎oktober ‎2018, ‏‎13:24:30).
I deleted the cache file and restarted.
At the moment it is downloading the Blocks.... will take a while, even through a remote node...

fouadr:
Solved by deleting the cache file again.

PR #2012, opened 2019-03-12
main: automatically fix common wallet cache errors

mmbyday:
Gracefully fixes wallet opening issues such as #1738 #1730 

Clears wallet cache for common error messages which normally require manual user intervention.
Handles cases involving basic_string::_M_replace_aux and std::bad_alloc error.
Old wallet cache remains and is renamed by walletManager.clearWalletCache.

Before
bef

After
aft1

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2018, The Monero Project
+// Copyright (c) 2014-2019, The Monero Project
 //
 // All rights reserved.
 //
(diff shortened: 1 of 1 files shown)


selsta:
Good idea! Does it still create a backup of the old wallet cache before clearing it?

mmbyday:
> Good idea! Does it still create a backup of the old wallet cache before clearing it?

Yes! Old wallet cache remains and is renamed by walletManager.clearWalletCache.

mmbyday:
> Nice! It can be simplified as the wallet repair procedure is the same:
> 
> ```
> case "basic_string::_M_replace_aux":
> case "std::bad_alloc":
>     ...
> ```
@MaxXor 
Nice idea! Feel free to submit PR. I tried that a similar idea. Didn't like it.

selsta:
>Yes! Old wallet cache remains and is renamed by walletManager.clearWalletCache.

Nice :) @dEBRUYNE-1 does anything speak against this change?

MaxXor:
@mmbyday Just include the suggested change in your PR. Why another?

dEBRUYNE-1:
@selsta - Should be fine as long as the file is renamed and retained. Any private transaction keys needed in the future can still be obtained then by loading the cache into a compatible version.

xmrdsc:
Very nice!

xmrdsc:
@MaxXor Feel free to create PR on top of this :-)

Issue #2023, opened 2019-03-17
Support moneroseed: scheme


xmrdsc:
Moo:

> I was playing around with a `moneroseed:` scheme to allow someone to start using monero without relying on the seed modes. Basically, Alice would tell Bob "download monero from this repo, run this setup thingie that sets up moneroseed: handling, then click on this link".

The intent is that someone new to Monero/GUI can get the source from wherever, and get a `moneroseed://a.b.c.d` from someone, click the link and the GUI starts via this seed. This bypasses the centralization inherent in the seed servers.

Scheme format is `monerod:127.0.0.1` but could be a FQDN as well, it essentially depends on what `--seed-node` from monerod can parse. GUI could decide to extend on this idea and go with a custom scheme that supports multiple seed node definitions.

### To-do's

- Make the OS recognize the scheme (Windows/Linux/Mac)
- Handle incoming moneroseed via a yet to be created GUI flag in `main.cpp`, pass it to QML
- Save this string in `appWindow.persistentSettings`, as it probably should be persistent.
- Upon spawning `monerod` from within the GUI, pass ipv4 of FQDN parsed from the moneroseed as `--seed-node` in `DaemonManager.cpp`.

Relevant moo branch: https://github.com/moneromooo-monero/bitmonero/commits/moneroseed

xdg-open example:

```
[Desktop Entry]
Name=Monero
GenericName=Monero
X-GNOME-FullName=Monero
Comment=Monero
Keywords=Monero;
Exec=/home/user/src/bitmonero/test-monero.sh %U
Terminal=true
Type=Application
Icon=monero
Categories=GNOME;
MimeType=x-scheme-handler/moneroseed;
StartupNotify=true
X-GNOME-Bugzilla-Bugzilla=GNOME
X-GNOME-UsesNotifications=true
```

moneromooo-monero:
> Save this string in appWindow.persistentSettings, as it probably should be persistent.

It should probably not be. It's meant as an entry point into the network, so you can get more peers via P2P comms. Once you've got that, it's probably best to do the "normal" thing.

xmrdsc:
Gotcha.

xmrdsc:
cross-ref https://github.com/monero-project/monero/issues/5426, waiting on that

kpcyrd:
[...]

Issue #222, opened 2016-11-28
Hook into URI scheme on, eg, Android, iOS


moneromooo-monero:
I've no idea how that'd technically work, but:

- the GUI now shows a QR code for simple payment
- there are now RPC and wallet2 API for parsing a monero: URI

So there needs to be a way for the wallet to register itself as a handler for monero: URIs. When invoked, this should call the wallet2 API (tools::wallet2::parse_uri) to get back the constituents, then call the transfer function with those, and ensure a user prompt will be made.

johnalanwoods:
This is trivial on iOS and macOS. Will investigate on Android.

erciccione:
@moneromooo-monero is still also present in new GUI or can be closed?

erciccione:
@moneromooo-monero This looks fixed in #2029. Please reopen if necessary.

+resolved

PR #2029, opened 2019-03-22
IPC and custom protocol handler for monero://

### LINKED ISSUE #760: [Feature Request] Uri scheme for Monero Core
Hello,
I'm developing integrations for Monero. I think I can include in checkout page a button that open monero-gui with address-field precompiled. This should a good idea, then the client can click with a button :)
For this, we need uri scheme support.
I can provide some resources about how to develop this 
Windows : https://msdn.microsoft.com/en-us/library/aa767914(v=vs.85).aspx

xmrdsc:
Adds support for the `monero://` scheme. allowing the GUI to launch from a (but not limited to) a browser.

Following parameters are supported:

- `monero://<address>`
- `monero://<address>?tx_amount=5.0`
- `monero://<address>?tx_description=foo`
- `monero://<address>?tx_payment_id=foo`
- `monero://<address>?tx_amount=5.0&tx_description=foo&tx_payment_id=foo`

This is also what MyMonero supports; [link](https://github.com/mymonero/mymonero-app-js/blob/fe8028556d4582d1e747e52871e068fd19b886ee/local_modules/MoneroUtils/monero_requestURI_utils.js#L61)

IPC stands for inter process communication. We use [QLocalServer](https://doc.qt.io/qt-5/qlocalserver.html) to communicate URI commands whenever an instance of the GUI is already running in order to avoid starting the GUI each time an user clicks on a monero link.

Fixes #760
#2023
#222

[Video example](https://www.youtube.com/watch?v=W_8dGIB7Kio).

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -399,6 +399,49 @@ ApplicationWindow {
         leftPanel.balanceLabelText = qsTr("Balance (#%1%2)").arg(currentWallet.currentSubaddressAccount).arg(accountLabel === "" ? "" : (" – " + accountLabel));
     }
 
+    function onUriHandler(uri){
+        if(uri.startsWith("monero://")){
+            var address = uri.substring("monero://".length);
+
+            var params = {}
+            if(address.length === 0) return;
+            var spl = address.split("?");
+
+            if(spl.length > 2) return;
+            if(spl.length >= 1) {
+                // parse additional params
+                address = spl[0];
+
+                if(spl.length === 2){
+                    spl.shift();
+                    var item = spl[0];
+
+                    var _spl = item.split("&");
+                    for (var param in _spl){
+                        var _item = _spl[param];
+                        if(!_item.indexOf("=") > 0) continue;
+
+                        var __spl = _item.split("=");
+                        if(__spl.length !== 2) continue;
+
+                        params[__spl[0]] = __spl[1];
+                    }
+                }
+            }
+
+            // Fill fields
+            middlePanel.transferView.sendTo(address, params["tx_payment_id"], params["tx_description"], params["tx_amount"]);
+
+            // Raise window
+            appWindow.raise();
+            appWindow.show();
+
+            // @TODO: remove after paymentID deprecation
+            if(params.hasOwnProperty("tx_payment_id"))
+                persistentSettings.showPid = true;
+        }
+    }
+
     function onWalletConnectionStatusChanged(status){
         console.log("Wallet connection status changed " + status)
         middlePanel.updateStatus();
@@ -489,6 +532,12 @@ ApplicationWindow {
 
         // Force switch normal view
         rootItem.state = "normal";
+
+        // Process queued IPC command
+        if(typeof IPC !== "undefined" && IPC.queuedCmd().length > 0){
+            var queuedCmd = IPC.queuedCmd();
+            if(/^\w+:\/\/(.*)$/.test(queuedCmd)) appWindow.onUriHandler(queuedCmd); // uri
+        }
     }
 
     function onWalletClosed(walletAddress) {
@@ -1086,6 +1135,7 @@ ApplicationWindow {
         walletManager.deviceButtonPressed.connect(onDeviceButtonPressed);
         walletManager.checkUpdatesComplete.connect(onWalletCheckUpdatesComplete);
         walletManager.walletPassphraseNeeded.connect(onWalletPassphraseNeeded);
+        IPC.uriHandler.connect(onUriHandler);
 
         if(typeof daemonManager != "undefined") {
             daemonManager.daemonStarted.connect(onDaemonStarted);
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -708,10 +708,20 @@ Rectangle {
     }
 
     // Popuplate fields from addressbook.
-    function sendTo(address, paymentId, description){
-        addressLine.text = address
-        setPaymentId(paymentId);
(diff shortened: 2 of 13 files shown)


xmrdsc:
@rbrunner7 

Registering protocols on Windows so that an user can click on `monero://` links (and taken to the GUI) is done inside the registry by adding keys/values in `HKEY_CLASSES_ROOT` (which is an alias to `HKEY_LOCAL_MACHINE\Software\Classes`).

Take this registry file that you can execute for example: `test.reg`:

```reg
REGEDIT4

[HKEY_CLASSES_ROOT\xx]
@="URL:xx Protocol"
"URL Protocol"=""

[HKEY_CLASSES_ROOT\xx\DefaultIcon]
@="\"C:\\Program Files\\xx\\xx.exe\""

[HKEY_CLASSES_ROOT\xx\shell]

[HKEY_CLASSES_ROOT\xx\shell\open]

[HKEY_CLASSES_ROOT\xx\shell\open\command]
@="\"D:\\app.exe\" \"%1\""
```

It will create the following layout;



After such entry,  a link like `xx://foo` on a website would launch `D:\app.exe`.

As for access control, Windows does:

#### Read/Write:

- HKEY_CURRENT_USER

#### Read Only:

- HKEY_LOCAL_MACHINE
- HKEY_CLASSES_ROOT

So, to write here we'll need admin access. Which brings me to your installer, it seems like during installation is the right time to register these protocols. Is this something you'll be able to incorporate? The following 2 schemes would need to get registered: `monero://foo` and `moneroseed://foo`

If not, I can create a mechanism that at runtime 1) checks if the entry already exists 2) if not, create it (and prompt Windows authentication thingy).

rbrunner7:
Yes, I would say such things definitely tend to be done with installers on Windows.

I will have a look.

selsta:
```
diff --git a/share/Info.plist b/share/Info.plist
index 5b7a0b3..438443d 100644
--- a/share/Info.plist
+++ b/share/Info.plist
@@ -31,5 +31,25 @@
 
   <key>NSRequiresAquaSystemAppearance</key>
   <string>True</string>
+
+  <key>CFBundleURLTypes</key>
+  <array>
+    <dict>
+      <key>CFBundleURLName</key>
+      <string>monero Handler</string>
+      <key>CFBundleURLSchemes</key>
+      <array>
+        <string>monero</string>
+      </array>
+    </dict>
+    <dict>
+      <key>CFBundleURLName</key>
+      <string>moneroseed Handler</string>
+      <key>CFBundleURLSchemes</key>
+      <array>
+        <string>moneroseed</string>
+      </array>
+    </dict>
+  </array>
 </dict>
 </plist>
```

selsta:
macOS: Clicking on e.g. monero://44AFFq5kSiGBoZ4NMDwYtN18obc8AemS33DBLWs3H7otXft3XjrpDtQGv7SqSsaBYBb98uNbr2VBBEt7f2wfn3RVGQBEP3A opens the GUI but doesn’t fill anything out like in your video.

xmrdsc:
@selsta 

Nice one. Argument should go into `--ipc`, not sure if you're able to do that in the Info.plist.

xmrdsc:
Could you try again @selsta ?

selsta:
Same as before :/ Maybe you can give me tips on how to debug this :P

rbrunner7:
> The following 2 schemes would need to get registered: monero://foo and moneroseed://foo

Will you create new command-line parameters for the GUI, to feed it a Monero address or a Monero seed? Or, what would be the corresponding commands to issue if somebody clicks on a link that uses one of those new schemes?

rbrunner7:
After reading more about all this I think I can answer my own question: The full URI will be added to the command line as parameter when the GUI wallet is started through a link using the new schemes. It should be trivial to scan the command line for `monero:` or `moneroseed:` there, probably no need to add anything else like `--uri` or similar.

xmrdsc:
@rbrunner7 Exactly, it'll be a positional argument that ill parse and sanitize. I've already made this (see code in `main.cpp`), example: `./monero-wallet-gui "monero://address?tx_amount=3.5&tx_description=foo"`.

I've come across [QDesktopServices::setUrlHandler](https://doc.qt.io/Qt-5/qdesktopservices.html#setUrlHandler) which allegedly automatically registers such protocol schemes on Windows and MacOS. We could go that route but I prefer from the installer on Windows.

xmrdsc:
Rebased.

pages/Transfer.qml
```diff
@@ -708,10 +708,20 @@ Rectangle {
     }
 
     // Popuplate fields from addressbook.
```
rbrunner7: Little typo in "Popuplate" ...

main.qml
```diff
@@ -399,6 +399,49 @@ ApplicationWindow {
...
+                        if(!_item.indexOf("=") > 0) continue;
+
+                        var __spl = _item.split("=");
+                        if(__spl.length !== 2) continue;
+
+                        params[__spl[0]] = __spl[1];
+                    }
+                }
+            }
+
```
rbrunner7: I have read (e.g. [here](https://msdn.microsoft.com/en-us/windows/desktop/aa767914#MainContent) that on Windows it's possible to encounter "percent encoding" sometimes in the arguments in connection with custom URI schemes, like `%20` for blanks somewhere, e.g. in `tx_description`. Did you test whether this actually occurs, and if yes, do you care about this somewhere i.e. decode?

PR #2019, opened 2019-03-15
device: Trezor support added, requires #5211

ph4r05:
Trezor support enabled.
All the work was done in Wallet::API and https://github.com/monero-project/monero/pull/5211 

Screenshot 2019-03-15 at 14 29 17

## Building

In order to test this PR you need to fetch my original branch as it requires newer Monero submodule

```bash
git clone --recursive https://github.com/monero-project/monero-gui.git
cd monero-gui
git remote add ph4r05 https://github.com/ph4r05/monero-gui.git
git fetch --all
git checkout -b trezor-dev ph4r05/trezor-dev

cd monero
git remote add ph4r05 https://github.com/ph4r05/monero.git
git fetch --all
git checkout -b trezor/api_pass ph4r05/trezor/api_pass
git submodule update --recursive
cd -
./build.sh
```

PR #2037, opened 2019-03-27
Async device open and create from device, passphrase entry, requires #5355

ph4r05:
- Requires https://github.com/monero-project/monero-gui/pull/2019
- Passphrase entry on host added, requires early listener setting https://github.com/monero-project/monero/pull/5355
- Wallet open and create from device shows splash to indicate a possible long process
- Create from a device is async to support passphrase entry

Screenshot 2019-03-27 at 09 27 35
Screenshot 2019-03-27 at 09 27 48
Screenshot 2019-03-27 at 09 28 22

## Building

In order to build this PR for testing you need to fetch my original branch

```bash
git clone --recursive https://github.com/monero-project/monero-gui.git
cd monero-gui
git remote add ph4r05 https://github.com/ph4r05/monero-gui.git
git fetch --all
git checkout -b trezor-pass ph4r05/trezor-pass

cd monero
git remote add ph4r05 https://github.com/ph4r05/monero.git
git fetch --all
git checkout -b trezor/api_pass ph4r05/trezor/api_pass
git submodule update --recursive
cd -
./build.sh
```

## Troubleshooting

If you experience `Bad offset calculation` error when creating a transaction the remote node you are using is probably running too old version.

The code is known to work with version `131076`. For testing purposes, try node `173.255.205.142:18089`, it should work smoothly.

### DIFF
--- a/components/PassphraseDialog.qml
+++ b/components/PassphraseDialog.qml
@@ -1,0 +1,327 @@
+// Copyright (c) 2014-2019, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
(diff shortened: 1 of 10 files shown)


components/PassphraseDialog.qml
```diff
@@ -0,0 +1,327 @@
...
+                Layout.fillWidth: true
+
+                font.pixelSize: 16 * scaleRatio
+                font.family: MoneroComponents.Style.fontLight.name
+
+                color: MoneroComponents.Style.defaultFontColor
+            }
+
+            Label {
+                text: qsTr("Warning: passphrase entry on host is a security risk as it can be captured by mallware. It is advised to prefer device-based passphrase entry.");
```
xmrdsc: malware

components/PassphraseDialog.qml
```diff
@@ -0,0 +1,327 @@
+// Copyright (c) 2017, The Monero Project
```
xmrdsc: `> // Copyright (c) 2014-2019, The Monero Project`

components/PassphraseDialog.qml
ph4r05: 😓

xmrdsc:
Nice PR. I have one small complaint with the QML, that is, `components/PassphraseDialog.qml` seems to be based off `components/PasswordDialog.qml`. They share similar functionality. I can attempt to get that into one file to avoid code duplication - and you wont have to deal with QML ;)

ph4r05:
> Nice PR. I have one small complaint with the QML, that is, `components/PassphraseDialog.qml` seems to be based off `components/PasswordDialog.qml`. They share similar functionality. I can attempt to get that into one file to avoid code duplication - and you wont have to deal with QML ;)

Thanks! 

The dialog is based on `NewPasswordDialog.qml`. We need to have 2 text boxes for passphrase entry as it is not checked by Trezor (on purpose, changes seed, plausible deniability, different wallets).

Moreover we need to have a warning there so user knows that host-based entry should be only fallback.

If you can factor out common code that would be nice. I didn't want to make huge changes to existing QMLs so this seemed to me as a path of minimal resistance :)

xmrdsc:
Ill look at it soon^tm, great work.

selsta:
Before this patch, it would display the following if the Ledger version was outdated:
Screenshot 2019-04-03 at 23 13 21
After this change, it shows “Creating wallet from device” for a short time and then fails without any error.

While the logs show the following, it would be great if this behaviour could be like before.
```
2019-04-03 21:13:50.097	  0x70000b235000	ERROR	device.ledger	src/device/device_ledger.cpp:322	Unsupported device application version: 1.2.2 At least 1.3.1 is required.
2019-04-03 21:13:50.098	     0x1135755c0	WARNING	frontend	src/wallet/api/wallet.cpp:403	qrc:/wizard/WizardController.qml:420: Error: Cannot assign [undefined] to QString
```

ph4r05:
I've added "Proceed to the device" splash logic to make it more explicit that some user action on the device is required. Otherwise user can just wait for operation to complete not knowing Trezor wants a confirmation.

ph4r05:
> the following, it would be great if this behaviour could be like before.

Ah thanks for info! I will check that

ph4r05:
@selsta could you pls try again?

selsta:
@ph4r05 Works now :)

xmrdsc:
Few days ago I argued we should avoid having to include `PassphraseDialog.qml` in this PR, however, I think we (GUI) should re-do password prompts anyway, possibly make it a dedicated screen as per @GBKS design (on zpl.io).

Current PR is fine by me.

PR #2041, opened 2019-03-28
Windows installer: Set up 'monero:' and 'moneroseed:' custom URI schemes

rbrunner7:
Together with the code for the GUI wallet itself from PR #2029 this implements custom URI schemes on Windows: The installer will register "monero:" and "moneroseed:" with Windows by writing corresponding registry entries.

As an advantage of letting the installer do this the two URI schemes will be deleted when uninstalling.

### DIFF
--- a/installers/windows/Monero.iss
+++ b/installers/windows/Monero.iss
@@ -70,12 +70,6 @@ Source: "FinishImage.bmp"; Flags: dontcopy
 Source: "bin\monero-wallet-gui.exe"; DestDir: "{app}"; Flags: ignoreversion
 Source: "bin\monero-GUI-guide.pdf"; DestDir: "{app}"; Flags: ignoreversion
 
-; Monero GUI wallet log file
-; The GUI wallet does not have the "--log-file" command-line option of the CLI wallet and insists to put the .log beside the .exe
-; so pre-create the file and give the necessary permissions to the wallet to write into it
-; Flag is "onlyifdoesntexist": We do not want to overwrite an already existing log
-Source: "monero-wallet-gui.log"; DestDir: "{app}"; Flags: onlyifdoesntexist; Permissions: users-modify
-
 ; Monero CLI wallet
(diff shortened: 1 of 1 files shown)


rbrunner7:
I tried the URI schemes under Windows 7: They worked well in Chrome, Internet Explorer and even in Opera 12.x, but sadly only partially in Firefox. Firefox did recognize `monero:` and `moneroseed:` as valid and opened a dialog box asking which external program to start, but for some mysterious reason failed to list the GUI wallet exe there as a choice automatically. I could however specify that exe as responsible manually and make the schemes fully working also on Firefox.

I tried to find out what Firefox "does not like" when it sees the GUI wallet exe but was not successful.

xmrdsc:
Nice work, good to hear it's working and well :-)

xmrdsc:
@rbrunner7 Could you confirm clicking the link works as expected when the GUI is already opened? Because that would test the IPC functionality of #2029, which on Windows is named pipes.

rbrunner7:
Ok, I can test with already-running GUI wallet this evening. Can you please point me to the right Buildbot build of the wallet to check that - lately I had no luck compiling it myself anymore ...

xmrdsc:
I reset the build on #2029 but looks like it failed.

xmrdsc:
Try build from #2061

rbrunner7:
I made an update today because in the latest Buildbot builds there are 2 new top-level files `libeay32.dll` and `ssleay32.dll` from OpenSSL that must be mentioned in the .iss installer script file in order to get installed.

xmrdsc:
I tested #2029 on a win10 Qt 5.10.1 machine, works fine. See [this image](https://i.imgur.com/p0mJViR.png). It does not start a second instance, but fills in the first upon running the command again.

I think the special branch I prepared for you had an unrelated error. We can merge this.

rbrunner7:
I updated yet again: I deleted the mechanism that forced the GUI wallet log to be written in the same directory as the exe. This is not necessary anymore, and now with the added way to start the GUI wallet using links even not working properly anymore.

PR #2055, opened 2019-04-05
AddressBook: redesign

selsta:
Basically exactly @GBKS design.

Screenshot 2019-04-05 at 19 51 30

~~Known bugs: If there is no address book entry, the headline wraps into the text below.~~


mmbyday:
Very nice. I like the changes @selsta 

A few quick observations

- delete an address book entry not yet working
- edit an address book entry creates an additional new entry
- spacing looks a little tight.
spacing

selsta:
@mmbyday Deleting and editing works fine on my system. (macOS)

What Qt version are you using?

mmbyday:
@selsta 5.11.1, which was the last known working QT version on Windows. Tried QT 5.7 on Windows in the past but no such luck.

selsta:
Note that editing uses a hack (deletes the selected entry + adds a new one with edited information) because there is no edit address book API function. The delete function uses only the API and should work without any problems.

selsta:
I can reproduce the issue with entries not getting deleted when using the buildbot build.

selsta:
Editing and deleting is now working.

xmrdsc:
Working great :+1:

luigi1111:
@selsta please rebase.

PR #2056, opened 2019-04-05
wizard: fix clearing of restore wallet fields

### LINKED ISSUE #2053: Monero GUI remembered mnemonic seed after logout of wallet
When you `Restore wallet from keys or mnemonic seed` then lock/logout of the wallet (not closed the wallet entirely) and try to `Restore wallet from keys or mnemonic seed` again then in the UI it will shows every detail that you've typed into the `Restore wallet` from the first import, including the seed, I think this is a security risk and the application should've cleared the UI data.

mmbyday:
Fixes #2053 
Fixes viewName variable name wizardRestoreWallet1
Now clears sensitive fields at end of wizard restore workflow.

### DIFF
--- a/wizard/WizardRestoreWallet1.qml
+++ b/wizard/WizardRestoreWallet1.qml
@@ -37,7 +37,7 @@ Rectangle {
     id: wizardRestoreWallet1
 
     color: "transparent"
-    property string viewName: "wizardCreateWallet1"
+    property string viewName: "wizardRestoreWallet1"
 
     function verify() {
         if (restoreHeight.text.indexOf('-') === 4 && restoreHeight.text.length !== 10) {
@@ -305,7 +305,7 @@ Rectangle {
     function onPageCompleted(previousView){
         if(previousView.viewName == "wizardHome"){
             // cleanup
-            walletInput.reset();
+            wizardWalletInput.reset();
             seedInput.text = "";
             addressLine.text = "";
             spendKeyLine.text = "";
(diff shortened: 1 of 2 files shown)


xmrdsc:
Nice one! I have one small adjustment to suggest, looking at `WizardCreateWallet1.qml` we can see a cleanup mechanism there:

https://github.com/monero-project/monero-gui/blob/19c2208dc447ccb5ba9ff990cbdd6ecf6b96f23d/wizard/WizardCreateWallet1.qml#L148-L152

Which means: if the user came from the home screen, clear the fields. Solution proposed here is to clear after writing a wallet. Either also change `WizardCreateWallet1.qml` (and possibly other files), or use the `previousView.viewName == "wizardHome"` technique.

mmbyday:
@xmrdc Thanks! In fact, I initially tried to use the previousView.viewName == "wizardHome" technique to clear the restore wallet fields on `WizardRestoreWallet1.qml` but there was no effect. However at a quick glance, it does seem to work on `WizardCreateWallet1.qml` I'll take a look at changing WizardCreateWallet1 to match but the intent of this PR was to fix issue 2053 😄

xmrdsc:
Ah I see the problem now. Current code in `WizardRestoreWallet1.qml`:

https://github.com/monero-project/monero-gui/blob/d2c47606ca8446e58ef122117f8f84b578691bab/wizard/WizardRestoreWallet1.qml#L305-L314

Errors with: `qrc:///wizard/WizardRestoreWallet1.qml:308: ReferenceError: walletInput is not defined`.

Simply change `walletInput.reset()` to `wizardWalletInput.reset()` and it should work.

PR #2057, opened 2019-04-08
flickable for wizardController

### LINKED ISSUE #2048: Scrolling for the wizards
> I'm on a tiny screen and I'm trying to connect to a remote node in the GUI. But the connect button is not visible in the new version of the monero software. There is no scroll button.
> Does anybody know what to do?

xmrdsc:
Scrolling like in `middlePanel.qml`

Fixes #2048

### DIFF
--- a/wizard/WizardController.qml
+++ b/wizard/WizardController.qml
@@ -26,13 +26,15 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import QtQml 2.0
 import QtQuick 2.7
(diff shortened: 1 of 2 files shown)


selsta:
Somehow I don’t see a scrollbar on macOS.

xmrdsc:
@selsta good catch, it fell outside of the window. Fixed.

mmbyday:
Works pretty good!

There might be a height issue during the create new wallet flow in the daemon settings screen.

errrrr

Incidentally, the language settings screen is not scrollable yet.

xmrdsc:
Oh, good point @mmbyday. Ill have to look into that. Thanks!

xmrdsc:
Seems to be a problem with the `RemoteNodeEdit{}` component regarding calculation of heights (?). Did not want to dive into it right now, rather fix it pragmatically.

As for language screen, it was a quick addition due to language sidebar not working. Will fix in another PR ;)

Scrolling for wizards are now working.

PR #2061, opened 2019-04-08
Fix master, remove unused image, requires #5355

xmrdsc:
Removes missing/unused image `dropdownDel.png` and fixes master by including PR 5355 for `get_libwallet_api.sh`


ph4r05:
Thanks for fixing it 🙏. 
My bad, I forgot to add `requires #5355` to the commit message.

PR #2065, opened 2019-04-11
PasswordDialog: don't always show()

### LINKED ISSUE #2054: GUI window keeps jumping into focus when it logs out automatically
There's a weird issue with the GUI which makes it annoying to keep it opened: when it logs me out of the wallet, it jumps into focus and asks me to enter my password. When I try to minimize it, it jumps back into view immediately.

The only way to minimize it is using Windows+M to minimize all windows. But then it returns in the same way after the period set for logging out.

GUI version: 14.0.0 x64 binaries. I use an old wallet, but I removed all files from C:\ProgramData\Monero and did a clean install of the client software.
OS: 64-bits Windows 10.
Expected behavior: log out of the wallet silently and keep the window minimized.

Video: https://imgur.com/0H3zVvt

mmbyday:
Seems to fix #2054

Please test.
Let inactivity timer expire. Minimize window. Window stays minimized after timer is reached.

### DIFF
--- a/components/NewPasswordDialog.qml
+++ b/components/NewPasswordDialog.qml
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, The Monero Project
+// Copyright (c) 2014-2019, The Monero Project
 //
 // All rights reserved.
 //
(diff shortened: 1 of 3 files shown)


xmrdsc:
Nice. You could also remove them for `NewPasswordDialog.qml` and `PassphraseDialog.qml`.

mmbyday:
Good idea. Updated.

PR #2066, opened 2019-04-12
readme: update link to pootle and add guide for translators

erciccione:
Done also in  monero-project/monero#5412


erciccione:
Edited @mmbyday

PR #2067, opened 2019-04-13
transfer: use checkbox2

mmbyday:
Spot the difference. Better with the same design elements IMO.

before:
befx

after:
aftttt

### DIFF
--- a/components/CheckBox2.qml
+++ b/components/CheckBox2.qml
@@ -29,12 +29,12 @@
 import QtQuick 2.0
 import QtQuick.Layouts 1.1
 import QtGraphicalEffects 1.0
-import "." 1.0
+import "../components" as MoneroComponents
 
 RowLayout {
     id: checkBox
     property alias text: label.text
-    property string checkedIcon: "../images/checkedIcon-black.png"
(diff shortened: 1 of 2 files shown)


xmrdsc:
I've always hated the plus signs. This looks way better to me.

selsta:
@GBKS Opinion? IMO we use arrows a lot already, so I preferred it the old way, but I’d be okay with this change to.

GBKS:
This is a good solution. The only thing I'd tweak is to tighten up the vertical spacing when the options are collapsed.

Alternatively, we could tweak the icon and change the labels to "Add payment ID" and "Add description", like this here:
monero-send-gbks-190415

This might be a different issue, but we also need to a title to this page. In user-testing, people didn't realize this was the send page when they first landed here after setting up their wallets.

selsta:
Screenshot 2019-04-16 at 11 27 03

IMO that is rather unintuitive, the arrows have a different meaning. Closing advanced options just hides things, closing payment ID / description removes what what was written inside the textbox. **Different behaviour** for the same icon.

I prefer your design @GBKS, adding “add” to the label in combination with the + icon makes its behaviour more clear.

Issue #2026, opened 2019-03-19
Create seperate QML application specifically for mobile


xmrdsc:
We make an effort in trying to ensure the QML we code runs on desktop, mobile, and tablet via stuff like `isMobile()` and `* scaleRatio`. This is not particularly pretty and more importantly, I don't think it would be wise to pursue [true responsiveness](https://www.w3schools.com/html/html_responsive.asp) within a single application,  at least not in QML.

I propose we create a second QML application specifically for mobile that's totally separate from our existing codebase. We should simply spawn a QML engine in `main.cpp` based on `isMobile`.

To sum up;

- Don't have to create crazy logic to support all kinds of screen sizes
- More performance, as we can optimize our QML specifically for mobile
- This way, the mobile application is not at the mercy of the 'flow' of the GUI, in terms of existing logic and views. GUI is made desktop-first, so it would be silly to try and 'make everything smaller' and hope this approach is good UX on mobile.

Thus, I conclude current situation is not reasonable and we'll need something custom to give mobile support the proper attention it needs. Doesn't have to be a complicated QML application.

PR #2069, opened 2019-04-14
[Qt 5.9] Remove scaleRatio

xmrdsc:
Removing `scaleRatio`, it is not needed. To scale the GUI; use environment variable `QT_SCALE_FACTOR`. There is commented code in [main.cpp](https://github.com/monero-project/monero-gui/blob/master/main.cpp#L99) that deals with DPI. And for mobile #2026 is a better idea.

Requires white-theme https://github.com/monero-project/monero-gui/pull/2060

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -120,8 +120,8 @@ Rectangle {
                 anchors.topMargin: 20
                 anchors.leftMargin: 20
                 anchors.verticalCenter: parent.verticalCenter
-                height: 490 * scaleRatio
-                width: 260 * scaleRatio
+                height: 490
+                width: 260
 
                 Image {
                     id: card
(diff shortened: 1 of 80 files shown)


twiuglufosfa:
after removing scaleratio i enabled highdpi on main.cpp and QT_SCALE_FACTOR on pro

 but is not working on android.

before removing scaleratio everything was ok.

i try all on QT_SCALE_FACTOR (1,2,1.3 etc)

xmrdsc:
@twiuglufosfa Android is unsupported, please see #2026

twiuglufosfa:
android was working very well with scaleratio

the only changed needed was ref_dpi on main.cpp

after removing scaleratio and inserting QT_SCALE_FACTOR on pro file

 resolution wont change so maybe i dont put right prices on QT_SCALE_FACTOR

or QT_SCALE_FACTOR dont work!

selsta:
Can you try adding `qputenv("QT_SCALE_FACTOR", "1.5");` to main.cpp?

twiuglufosfa:
worked!

thnx a lot.

PR #2073, opened 2019-04-15
 wizard/device: hide splash screen

selsta:
Else it will look like this in combination with white theme PR.

Screenshot 2019-04-15 at 08 20 05

@ph4r05 Does this look correct? Is it getting hidden in a different place?

### DIFF
--- a/wizard/WizardCreateDevice1.qml
+++ b/wizard/WizardCreateDevice1.qml
@@ -227,6 +227,7 @@ Rectangle {
     }
 
     function onCreateWalletFromDeviceCompleted(written){
+        hideProcessingSplash();
         if(written){
             wizardStateView.state = "wizardCreateWallet2";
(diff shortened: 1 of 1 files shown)


ph4r05:
Hmm I will need a closer look once I am on PC. 

Do you want to disable splash screen in white theme? I mean, splash screen serves a nice purpose, maybe just splash screen redesign for white theme would be better?

selsta:
@ph4r05 I think with this patch it works fine with white theme.

ph4r05:
Aah OK! Sry. I will check it out properly as soon as I am on PC

ph4r05:
@selsta your commit looks ok! Thanks for fixing it! :)

selsta:
@ph4r05 np, thanks for testing. The PR title could have been worded better.

PR #2080, opened 2019-04-16
Transfer: warn if description is a valid PID

selsta:
Screenshot 2019-04-16 at 13 23 59


xmrdsc:
Very nice :-)

xmrdsc:
Maybe add `// @TODO: remove after pid removal hardfork`comment on top of the warning box component, else we might forget to remove it in the future. I regularly grep for TODOs.

selsta:
Good idea, updated.

PR #2081, opened 2019-04-16
[Qt 5.9] Dynamic height settings wallet

xmrdsc:
> You could also look into Settings -> Wallet alignment. It messes up if the window width is small.

Requires #2060


luigi1111:
@xmrdsc please rebase.

Issue #2089, opened 2019-04-16
Check if updates are available


xmrdsc:
GUI has functionality to check if a newer version of the GUI is available, via DNS. I believe this was disabled some time ago. We should get this functionality working. 

This also has overlap with a yet-to-be-made issue about redesigning all popups/modals.

Electricsheep01:
Wasn’t this deemed a security risk? Big + on UX but

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

erciccione:
> Wasn’t this deemed a security risk? Big + on UX but

Yes, but AFAIK the CLI check for updates, it just doesn't install them. I think something like that could be done for the GUI as well.

xmrdsc:
Update mechanism leverages DNS to fetch latest version but not auto-download.

GBKS:
Here's a design for this modal, just in case you'd like to have one. Would it be possible to add a link to a summary of what's new in the update?

![monero-new-version-gbks-190525](https://user-images.githubusercontent.com/695901/58366052-fb890580-7ecc-11e9-8db5-e756218143d4.png)

selsta:
@GBKS Can you upload the modal to Zeplin? Looks good.

> We should get this functionality working.

Working again.

+resolved

GBKS:
Sorry, I don't believe I ever uploaded this to Zeplin. I recently moved my design files to Figma, which should allow you to directly inspect the designs from there. If you still want to work with this design, here's the [link to Figma](https://www.figma.com/file/DplJ2DDQfIKiuRvolHX2hN/Monero-GUI?node-id=0%3A41).

selsta:
Does Figma show the exact measurements like Zeplin does?

GBKS:
Yep, works almost the same way. You can click, hover and command-hover things to see all kinds of spacing. There's also a code panel on the right to see font sizes and other values. But I'm sure there are some small details that are different, let me know if anything is a big issue for you.

selsta:
I had to signup to see all the details, nice, looks like a good Zeplin alternative. I wonder if we should add the link to the README.

GBKS:
We can definitely do that. I published this Figma file as a community file that I can regularly push updates to and people can freely duplicate. If you want to add a link, here's the one to the [public file](https://www.figma.com/c/file/775423808468574409). I'm also planning to update my Github repo where I had the Sketch files to point to Figma. Just seems a little better suited for open-source design collaboration.

[...]

PR #2100, opened 2019-04-21
Fix checking for gui updates

mmbyday:
#2089
old

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -2001,15 +2001,23 @@ ApplicationWindow {
         print("Update found: " + update)
         var parts = update.split("|")
         if (parts.length == 4) {
-          var version = parts[0]
-          var hash = parts[1]
-          var user_url = parts[2]
-          var auto_url = parts[3]
-          var msg = qsTr("New version of monero-wallet-gui is available: %1<br>%2").arg(version).arg(user_url) + translationManager.emptyString
-          notifier.show(msg)
-        }
-        else {
-          print("Failed to parse update spec")
+            var version = parts[0]
+            var hash = parts[1]
+            //var user_url = parts[2]
+            //var auto_url = parts[3]
+            var osBuildTag = isMac ? "mac-x64" : isWindows ? "win-x64" : isLinux ? "linux-x64" : "unknownBuildTag"
+            var extension = isMac || isLinux ? ".tar.bz2" : isWindows ? ".zip" : ".unknownExtension"
+            var base_url = "https://downloads.getmonero.org/gui/monero-gui-"
+            var download_url = base_url + osBuildTag + "-v" + version + extension
+            var msg = ""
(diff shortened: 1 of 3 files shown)


xmrdsc:
`function onWalletCheckUpdatesComplete(update) {`

update is always empty for me. This is the bug I meant in #2089

erciccione:
I'm not sure the direct link is the best option.Maybe would be better something more mininmalistic? like: "a new version is available, check out getmonero.org".

xmrdsc:
To verify, how were you able to test the update mechanism? Can you confirm the GUI:

1. Makes a DNS request on startup
2. `WalletManager::checkUpdatesAsync` emits the `checkUpdatesComplete` with a non-empty `update` parameter when it is done checking.
3. So that `onWalletCheckUpdatesComplete(update)` in `main.qml` is not empty
4. So that it shows the dialog

^ cannot reproduce

mmbyday:
@xmrdsc Good questions. I tested this PR against the v0.13.0.4 branch, as you may see from clues in the GUI (missing accounts, openalias placeholdertext)

Yes, your understanding sounds almost right. In step 2, `checkUpdatesComplete` can be empty when there is no update.

You can look under the hood of `WalletManager::checkUpdates` and see that the `tuple `of update data comes from `Monero::WalletManager::checkUpdates` here:
https://github.com/monero-project/monero/blob/29073f65e8816d4c32b6ffef514943a5650b8d3b/src/wallet/api/wallet_manager.cpp#L334

When there is no new update, this is why `update` parameter is empty.
```
    if (!tools::check_updates(software, buildtag, version, hash))
      return std::make_tuple(false, "", "", "", "");
```

Hope that helps you.

mmbyday:
@erciccione That's a good option. It's simple and minimalistic as you say. Ultimately however, I think the typical user wants a one-click update though. That's what most user friendly wallets do for updates, like ledger-live.

mmbyday:
Rebased.

main.qml
```diff
@@ -1953,9 +1953,13 @@ ApplicationWindow {
         if (parts.length == 4) {
           var version = parts[0]
           var hash = parts[1]
-          var user_url = parts[2]
-          var auto_url = parts[3]
-          var msg = qsTr("New version of monero-wallet-gui is available: %1<br>%2").arg(version).arg(user_url) + translationManager.emptyString
+          //var user_url = parts[2]
+          //var auto_url = parts[3]
+          var osBuildTag = isMac ? "mac-x64" : isWindows ? "win-x64" : isLinux ? "linux-x64" : "unknownBuildTag"
```
xiphon: Return or throw a error in the unknown/unrecognized OS case.

mmbyday: Excellent idea. Thanks.

mmbyday:
Updated to handle unknown/unrecognized OS case with generic message.

PR #2102, opened 2019-04-21
wizardhome: fix network dropdown

### LINKED ISSUE #2101: Can't select different networks on wizardhome

mmbyday:
Resolves #2101

### DIFF
--- a/wizard/WizardController.qml
+++ b/wizard/WizardController.qml
@@ -107,7 +107,7 @@ Rectangle {
     property string walletRestoreMode: 'seed'  // seed, keys, qr
 
     // flickable margin
-    property int flickableHeightMargin
+    property int flickableHeightMargin: 200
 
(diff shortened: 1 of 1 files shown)


xmrdsc:
Cannot reproduce the problem, don't think this fix is needed.

mmbyday:
It's happening on the nightly builds as well.

xmrdsc:
Sorry, I can reproduce. This is since the new `Flickable` for `WizardController`.

https://github.com/monero-project/monero-gui/blob/bd2687c22f5d147e382e1440fd2f999b83f8dcb4/wizard/WizardController.qml#L110

Set this property to `200` :-)

mmbyday:
Nice call. Updated :-)

xmrdsc:
Nice :-)

PR #2128, opened 2019-04-29
logging: initial log level set by ENV var

ph4r05:
Enables to setup logging level from the start so the Wizard logs are visible for better debugging.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -140,7 +140,12 @@ int main(int argc, char *argv[])
 
 
     // loglevel is configured in main.qml. Anything lower than
-    // qWarning is not shown here.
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -141,6 +141,14 @@ int main(int argc, char *argv[])
 
     // loglevel is configured in main.qml. Anything lower than
     // qWarning is not shown here.
+    QString logLevelStr = qgetenv("MONERO_LOG_LEVEL");
```
xiphon: Use
```c++
int qEnvironmentVariableIntValue(const char *varName, bool *ok=nullptr) noexcept`
```

ph4r05: Aah that looks good, thanks for suggestion!

PR #2131, opened 2019-04-29
Transfer: warn when sending to a known long PID service

selsta:
Screenshot 2019-04-29 at 17 31 28


luigi1111:
@selsta please rebase.

PR #2142, opened 2019-05-01
wizardcreatedevice: cleanup layout

mmbyday:
hw

### DIFF
--- a/wizard/WizardCreateDevice1.qml
+++ b/wizard/WizardCreateDevice1.qml
@@ -46,8 +46,9 @@ Rectangle {
 
     ListModel {
         id: deviceNameModel
-        ListElement { column1: "Ledger"; column2: "Ledger"; }
-        ListElement { column1: "Trezor"; column2: "Trezor"; }
+        ListElement { column1: qsTr("Choose your hardware device"); column2: "";}
+        ListElement { column1: "Ledger"; column2: "Ledger";}
+        ListElement { column1: "Trezor"; column2: "Trezor";}
     }
 
     function update(){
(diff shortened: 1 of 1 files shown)


dEBRUYNE-1:
Would it perhaps be better if we set a default value for the subaddress lookahead box for hardware wallets? The default value for normal wallets is 50:200, but that would be too high for hardware wallets in my opinion, as it would take a long time to generate the wallet.

mmbyday:
@dEBRUYNE-1 Solid idea. What would you say is a reasonable default value?

selsta:
50:200 is only a problem for Ledger, Trezor works fine.

dEBRUYNE-1:
@mmbyday @selsta - What if we use 3:100 for Ledger (the value being used in the guides) and 50:200 for Trezor then? Also, the values should be overridden if the user manually enters a value in the box.

mmbyday:
@dEBRUYNE-1 Updated with your recommendations. Thanks.

dEBRUYNE-1:
Thanks!

selsta:
Somehow 3:100 takes longer on my Ledger than 50:200? I’m confused.

xmrdsc:
Nice work, love it.

selsta:
`3:100` takes 72s on my system, `50:200` takes 27s.

selsta:
@mmbyday Can you revert your latest changes? (https://github.com/monero-project/monero-gui/compare/fc8c00287aa097d05993cfa1c75550ee9218a8f9..c0619c37fab399451d36185856428269f5d15879)

We found out that Ledger already uses a lower value by default (5:20), so this change isn’t necessary.

mmbyday:
Thanks for testing! Last change reverted.

Issue #2123, opened 2019-04-28
[Trezor] - Proceed to device not shown after TX signature


ph4r05:
After the spending transaction is signed with the Trezor, GUI prompts for a wallet password and freezes, as Trezor device expects user input (accept live refresh request). GUI does not show "Proceed to the device" splash from some reason.

After the user clicks the Trezor button process continues normally.
I will try to address this but if you happen to know why is it happening it might be useful. Thanks! :)

ph4r05:
I think I found the reason - `transaction.commit()` is blocking call, this prevents the popup from showing.

ph4r05:
@selsta @xmrdsc @xiphon do you think it would be beneficial to make an async wrapper for the PendingTransaction.commit() and commit in async way?

selsta:
I know nothing about async, would there be any downsides? If not, it probably would be beneficial.

xiphon:
> do you think it would be beneficial to make an async wrapper for the PendingTransaction.commit() and commit in async way?

Sure, async wrapper would definitely be beneficial.

selsta:
Seems resolved.

PR #2147, opened 2019-05-02
transaction: async commit

ph4r05:
Addresses #2123, reduces UI freezing on long-lasting operation, improves Trezor experience


selsta:
This doesn’t seem to work with Ledger, after a successful transaction it begins with the sending process again. I’ll have to investigate further.

ph4r05:
> This doesn’t seem to work with Ledger, after a successful transaction it begins with the sending process again. I’ll have to investigate further.

Do you pls have a log from the signing? Currently, I don't see why it behaves differently for Ledger. It should work in general :/

Did you test also against the master?

selsta:
After trying it again, I can’t reproduce it anymore and seems to work fine with Ledger.

ph4r05:
Just a note: https://github.com/trezor/trezor-firmware/pull/129 significantly improves UX of the Trezor hardware wallet. The device prompts live refresh permission only once.

PR #2149, opened 2019-05-02
Check before disconnecting wallet

xmrdsc:
Else it errors here:

https://github.com/xmrdsc/monero-core/blob/dbe5b00ae5acf65c6c497014337c71ae98430f27/main.qml#L597

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -588,6 +588,9 @@ ApplicationWindow {
     }
 
     function disconnectRemoteNode() {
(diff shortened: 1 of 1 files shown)


main.qml
```diff
@@ -588,6 +588,9 @@ ApplicationWindow {
     }
 
     function disconnectRemoteNode() {
+        if (typeof currentWallet !== "undefined" || currentWallet !== null)
```
xiphon: Shouldn't it be reversed?

PR #2158, opened 2019-05-03
wallet: fix deadlock, omit invoking 'refresh' from callback

xiphon:
https://github.com/monero-project/monero/issues/5515


selsta:
I guess this explains why it refreshed twice after sending.

ph4r05:
This fixes the problem 👍

PR #2168, opened 2019-05-15
wizard: Fix restore height when using stagenet/testnet

### LINKED ISSUE #2150: Restore height as date is wrong when using stagenet/testnet
Solution: Remove this function for stagenet/testnet or add a getApproximateBlockchainHeight function for stagenet/testnet.

mmbyday:
Resolves #2150


mmbyday:
Updated, thanks! :-)

PR #2174, opened 2019-05-22
use https to fetch remote nodes list on Windows

xiphon:


xiphon:
@mmbyday 

Make sure your Qt is build with openssl support and you have openssl dynamic libs installed on the system

mmbyday:
@xiphon Thanks for the info and your patience. I think my openssl is setup accordingly. Can you verify the [readme ](https://github.com/monero-project/monero-gui/blob/master/README.md)and build process already includes building with openssl support and installation of dynamic libs? Or is there additional steps required and the readme would need amending.

xiphon:
@mmbyday 

Please checkout #2177 changes, run `make deploy`, test the GUI and report back the results.

PR #2175, opened 2019-05-25
Expose get/set attrs for wallet cache in libwalletqt

xmrdsc:
```C++
#include "WalletManager.h"
#include "Wallet.h"

WalletManager *wallet_manager = WalletManager::instance();
Wallet *wallet = wallet_manager->openWallet("/home/foobar/Monero/wallets/foobar/foobar.keys", "passwd", NetworkType::MAINNET);
    
wallet->setCacheAttribute("foo_key", "bar_value");
qDebug() << wallet->getCacheAttribute("foo_key");
```

Requires https://github.com/monero-project/monero/pull/5534

### DIFF
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -634,6 +634,15 @@ void Wallet::setPaymentId(const QString &paymentId)
     m_paymentId = paymentId;
 }
 
+QString Wallet::getCacheAttribute(const QString &key) const {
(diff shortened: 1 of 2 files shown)


src/libwalletqt/Wallet.cpp
```diff
@@ -634,6 +634,15 @@ void Wallet::setPaymentId(const QString &paymentId)
     m_paymentId = paymentId;
 }
 
+QString Wallet::getCacheAttribute(const QString &key) const {
```
selsta: Nitpick: `{` should be on the next line.

PR #2180, opened 2019-05-28
WalletManager: fix compilation warnings (unused args, members order)

xiphon:


xiphon:
@MaxXor 

> The functions `moveWallet` and `walletLanguage` are not even used in the code. It looks like the wallet language is handled somewhere else already. Maybe remove them altogether?

Somehow missed your comment. Yeah, looks like we can drop both these functions.

PR #2196, opened 2019-06-04
transfer: add do-not-relay mode

### LINKED ISSUE #2130: [UX] Add toggle to disable send button
At the moment the only way to ensure a full wallet does not relay a transaction is with the CLI's start option `--do-not-relay`
For the GUI (with full wallet) there is an advanced button to sign a TX file, but there is no option to disable the send button.

That means in the CLI (with `--do-not-relay`) you can't accidentally relay a transaction.
With the GUI you *can* accidentally relay if you're not careful.

**Suggestion:**
Add a setting for the GUI that diables the "Send" button or alternatively replace it with the "Create TX file" button from the "Advanced" menu.

This way one does not have to remember *not* to click "Send" everytime, but only needs to set an option once.

EDIT: Changed the wording to clarify a bit more.

mmbyday:
Resolves #2130
create

- changed layout of create txn button on transfer page
- fixed bug with transaction file path variable
- added status messages to keep user informed during signTxDialog
- added new persistentSettings.doNotRelayMode


pages/Transfer.qml
```diff
@@ -396,15 +396,16 @@ Rectangle {
...
           StandardButton {
               id: sendButton
               rightIcon: "qrc:///images/rightArrow.png"
               rightIconInactive: "qrc:///images/rightArrowInactive.png"
-              Layout.topMargin: 4
               text: qsTr("Send") + translationManager.emptyString
               enabled: {
                 updateSendButton()
               }
+              visible: !appWindow.viewOnly && (!appWindow.viewOnly === !persistentSettings.doNotRelayMode)
```
selsta: Could you do `!(appWindow.viewOnly || persistentSettings.doNotRelayMode)` here or am I missing something?

pages/Transfer.qml
```diff
@@ -396,15 +396,16 @@ Rectangle {
...
       RowLayout {
+          Layout.topMargin: 4
           StandardButton {
               id: sendButton
               rightIcon: "qrc:///images/rightArrow.png"
               rightIconInactive: "qrc:///images/rightArrowInactive.png"
-              Layout.topMargin: 4
               text: qsTr("Send") + translationManager.emptyString
               enabled: {
                 updateSendButton()
```
selsta:
```diff
diff --git a/pages/Transfer.qml b/pages/Transfer.qml
index 63560b2..56ea0cd 100644
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -740,7 +740,7 @@ Rectangle {
         root.sendButtonWarning = "";
 
         // Currently opened wallet is not view-only
-        if(appWindow.viewOnly){
+        if(appWindow.viewOnly && persistentSettings.doNotRelayMode){
             root.sendButtonWarning = qsTr("Wallet is view-only and sends are not possible.") + translationManager.emptyString;
             return false;
         }
```

pages/Transfer.qml
mmbyday: Nice and simple 👍 Thanks

pages/Transfer.qml
mmbyday: 👍  nice one

Mansarde:
Nice, thanks for implementing this feature!
Hope it makes it into upstream. 👍

xmrdsc:
I would prefer such a flow:

```                               /------ broadcast
Create transaction ---/
                                  \------ create tx file  
```

Send -> Create transaction -> popup window -> 2 options here.

Can be integrated in https://user-images.githubusercontent.com/695901/57525290-94931a80-732a-11e9-9928-af69da552464.png

selsta:
@xmrdsc Do you want this as default behaviour? I’d argue that creating a tx file is something 99% of users don’t do, so asking every time when sending if you want to broadcast the tx or not is redundant. Also it doesn’t address #2130.

The current UI without do-not-relay mode could be improved, though I guess that’s out of scope for this PR.

Mansarde:
What about this?

Do-not-relay **off**:
Behaviour remains as it already is.
I.e. send button can be used and advanced menu has "create TX file" button.

Do-not-relay **on**:
A) Send button is disabled and only the "create TX file" button of the advanced menu works.
OR
B) Send button is simply replaced with "create TX file" button.

selsta:
If we add this, I’d say your first suggestion. Everything else just adds more unmaintainability to the code.

PR #2199, opened 2019-06-06
fontawesomeness button/icon refactoring

mmbyday:
eee

Tested against start-low-graphics-mode 👍


selsta:
@mmbyday Can you maybe hop on #monero-irc (freenode)? :)

xmrdsc:
Firstly, this PR is quite nice. I am wondering one thing; 

Maybe instead of introducing `FontAwesomeButton.qml` we could incorporate it inside `IconButton.qml` - since essentially they try to achieve the same thing, that is, It's an icon ... and it's a button :-) 

The only difference is that one is an actual image, the other is a font-awesome character. This difference should not matter to the person writing QML, he just wants to use a QML component with a clean API that is able to work with both images and FA.

Loving the new icons 👍

GBKS:
@selsta to be honest, I don't like the FontAwesome icons very much. They are a too chunky and feel unrefined. If you look at the Pro icons of FontAwesome, they are much nicer. Probably intentionally that way so people upgrade to their paid accounts. I'd be happy to design any icons we need, just let me know what's missing.

PR #2201, opened 2019-06-08
DaemonManager: disable SSL for local daemon

xiphon:


xiphon:
Fixed it in on the daemon side https://github.com/monero-project/monero/pull/5618

PR #2220, opened 2019-06-17
get_tx_key fixes for Trezor

ph4r05:
- do not display tx key for Trezor-based wallets in the Transaction history by default. It required device roundtrip with confirmation which made browsing the history unusable.  
- async behavior: getTxKey, get tx info, get tx proof
- fixes issue https://github.com/monero-project/monero-gui/issues/2183
- fixes the deadlock issues
- requires https://github.com/monero-project/monero/pull/5662 to work correctly (otherwise, zeros are returned, async nature cannot be tested, proofs don't work, etc...). However, it compiles without it and this PR can be merged before the monero-core PR is merged and submodule updated.

Screenshot 2019-06-17 at 16 05 09

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1003,11 +1003,21 @@ ApplicationWindow {
                     ", address: ", address,
                     ", message: ", message);
 
-        var result;
+        function spendProofFallback(txid, result){
+            if (!result || result.indexOf("error|") === 0) {
+                currentWallet.getSpendProofAsync(txid, message, txProofComputed);
+            } else {
+                txProofComputed(txid, result);
+            }
+        }
+
         if (address.length > 0)
-            result = currentWallet.getTxProof(txid, address, message);
-        if (!result || result.indexOf("error|") === 0)
-            result = currentWallet.getSpendProof(txid, message);
+            currentWallet.getTxProofAsync(txid, address, message, spendProofFallback);
+        else
+            spendProofFallback(txid, null);
+    }
+
+    function txProofComputed(txid, result){
         informationPopup.title  = qsTr("Payment proof") + translationManager.emptyString;
         if (result.indexOf("error|") === 0) {
             var errorString = result.split("|")[1];
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -1576,16 +1576,16 @@ Rectangle {
 
     function getTxKey(hash, elem){
         if (elem.parent.state != 'ready'){
-            var txKey = currentWallet.getTxKey(hash)
-            elem.parent.text = txKey ? txKey : '-';
-            elem.parent.state = 'ready';
+            currentWallet.getTxKeyAsync(hash, function(hash, txKey) {
+                elem.parent.text = txKey ? txKey : '-';
+                elem.parent.state = 'ready';
+            });
         }
 
         toClipboard(elem.parent.text);
     }
 
     function showTxDetails(hash, paymentId, destinations, subaddrAccount, subaddrIndex){
-        var tx_key = currentWallet.getTxKey(hash)
         var tx_note = currentWallet.getUserNote(hash)
         var rings = currentWallet.getRings(hash)
         var address_label = subaddrIndex == 0 ? (qsTr("Primary address") + translationManager.emptyString) : currentWallet.getSubaddressLabel(subaddrAccount, subaddrIndex)
@@ -1593,10 +1593,12 @@ Rectangle {
         if (rings)
             rings = rings.replace(/\|/g, '\n')
 
-        informationPopup.title = qsTr("Transaction details") + translationManager.emptyString;
-        informationPopup.content = buildTxDetailsString(hash, paymentId, tx_key, tx_note, destinations, rings, address, address_label);
-        informationPopup.onCloseCallback = null
-        informationPopup.open();
+        currentWallet.getTxKeyAsync(hash, function(hash, tx_key) {
+            informationPopup.title = qsTr("Transaction details") + translationManager.emptyString;
+            informationPopup.content = buildTxDetailsString(hash, paymentId, tx_key, tx_note, destinations, rings, address, address_label);
+            informationPopup.onCloseCallback = null
+            informationPopup.open();
+        });
     }
 
     function showTxProof(hash, paymentId, destinations, subaddrAccount, subaddrIndex){
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -1099,17 +1099,13 @@ Rectangle {
                             MoneroComponents.TextPlain {
                                 font.family: MoneroComponents.Style.fontRegular.name
                                 font.pixelSize: 15
-                                text: {
-                                    var txKey = currentWallet.getTxKey(hash)
-                                    if(txKey) return txKey;
-                                    else return "-"
-                                }
-
+                                text: qsTr("Click to reveal")
                                 color: MoneroComponents.Style.defaultFontColor
                                 anchors.verticalCenter: parent.verticalCenter
+                                state: "txkey_hidden"
 
                                 MouseArea {
-                                    state: "copyable"
+                                    state: "copyable_txkey"
                                     anchors.fill: parent
                                     hoverEnabled: true
                                     onEntered: parent.color = MoneroComponents.Style.orange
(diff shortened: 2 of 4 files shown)


ph4r05:
@xiphon what do you think? It is PoC but it works quite nice already. I am open to improvement suggestions :)

src/libwalletqt/Wallet.h
```diff
@@ -178,6 +179,10 @@ class Wallet : public QObject
     Q_INVOKABLE QString getSubaddressLabel(quint32 accountIndex, quint32 addressIndex) const;
     Q_INVOKABLE void setSubaddressLabel(quint32 accountIndex, quint32 addressIndex, const QString &label);
 
+    //! hw-device backed wallets
+    Q_INVOKABLE bool isHwBacked() const;
```
xiphon: Seems it is not used anywhere

pages/History.qml
```diff
@@ -1095,11 +1095,20 @@ Rectangle {
                             color: "transparent"
                             Layout.fillWidth: true
                             Layout.preferredHeight: 20
+                            state: "ready"
 
                             MoneroComponents.TextPlain {
                                 font.family: MoneroComponents.Style.fontRegular.name
                                 font.pixelSize: 15
                                 text: {
+                                    if (currentWallet.isTrezorBacked()){
```
xiphon: Consider to move this to `Component.onCompleted`

src/libwalletqt/Wallet.cpp
```diff
@@ -657,9 +668,24 @@ QString Wallet::getUserNote(const QString &txid) const
   return QString::fromStdString(m_walletImpl->getUserNote(txid.toStdString()));
 }
 
-QString Wallet::getTxKey(const QString &txid) const
+QString Wallet::getCachedTxKey(const QString &txid) const
 {
-  return QString::fromStdString(m_walletImpl->getTxKey(txid.toStdString()));
+    return m_txKeyCache.contains(txid) ? m_txKeyCache[txid] : QString();
```
xiphon:
```suggestion
    return m_txKeyCache.value(txid, "");
```

pages/History.qml
ph4r05: do you mean just the `currentWallet.isTrezorBacked()` check?

xiphon:
@ph4r05 

> * cache tx_keys recovered by Trezor for smooth user experience (don't do the same work twice, e.g., when user is browsing transaction history)
> * do not display tx key for Trezor-based wallets in the Transaction history by default. It required device roundtrip with confirmation which made browsing the history unusable.

definitely like the idea

pages/History.qml
xiphon: Nvm, let's keep it as is

src/libwalletqt/Wallet.cpp
```diff
@@ -657,9 +662,24 @@ QString Wallet::getUserNote(const QString &txid) const
...
-QString Wallet::getTxKey(const QString &txid) const
+QString Wallet::getCachedTxKey(const QString &txid) const
+{
+    return m_txKeyCache.value(txid);
+}
+
+QString Wallet::getTxKey(const QString &txid)
 {
-  return QString::fromStdString(m_walletImpl->getTxKey(txid.toStdString()));
+    if (m_txKeyCache.contains(txid))
```
xiphon: reuse getCachedTxKey

xiphon:
Have another idea based on this one.

Taken security measures in consideration i think we should always display "Click to reveal" link instead of tx keys for all wallets (not just Trezor-based).
Using this approach we don't need to cache the keys, it is highly unlikely that a user would like to fetch the same tx key twice.

@ph4r05 
Would you like to do this? If no, just let me know

ph4r05:
@xiphon: I've updated the PR to the form you proposed. It is nice to have uniform UX.

ph4r05:
xiphon: I've been thinking about making it async already so it has proper UX and come to this...

Pls check this out whether it is OK for you, if yes, I can squash it and alter PR description and commit message. 

I've used a different async concept rather than signals, which are quite hard to use with UI elements. The resulting code would be a bit complicated. So I've created async methods which take `QJSValue`  as the last parameter in the `Wallet.h`. If the parameter is callable it is considered as a callback. This simplified async design quite a lot IMO. 

Pls let me know what do you think.

src/libwalletqt/Wallet.cpp
```diff
@@ -662,6 +662,26 @@ QString Wallet::getTxKey(const QString &txid) const
...
+    auto watcher = new QFutureWatcher<QString>();
+
+    connect(watcher, &QFutureWatcher<QString>::finished,
+            this, [this, watcher, txid, ref]() {
+          QFuture<QString> future = watcher->future();
+          watcher->deleteLater();
+
+          auto txKey = future.result();
+          if (ref.isCallable()){
+              QJSValue cb(ref);
```
ph4r05: Copy is required as param is a const reference and `call()` method is not const in Qt5 (there is a comment it should be const in Qt6). AFAIK this is usual practice.

src/libwalletqt/Wallet.cpp
```diff
@@ -662,6 +662,26 @@ QString Wallet::getTxKey(const QString &txid) const
...
+              cb.call(QJSValueList {txid, txKey});
+          }
+          emit txKeyReceived(txid, txKey, ref);
```
ph4r05: For greater flexilibity I've added standard signal which is triggered anyway. The last parameter can be aux value (e.g., UI element, related information) if it is not used as a callback.

src/libwalletqt/Wallet.h
```diff
@@ -348,6 +352,9 @@ class Wallet : public QObject
     void deviceButtonPressed();
     void transactionCommitted(bool status, PendingTransaction *t, QStringList txid);
     void heightRefreshed(quint64 walletHeight, quint64 daemonHeight, quint64 targetHeight) const;
+    void txKeyReceived(const QString &txId, const QString &txKey, const QJSValue &ref) const;
```
xiphon: Would prefer to drop these unused signals.

src/libwalletqt/Wallet.cpp
xiphon: Think we won't ever need this. Callbacks are flexible enough

src/libwalletqt/Wallet.cpp
```diff
@@ -699,6 +739,26 @@ Q_INVOKABLE QString Wallet::getSpendProof(const QString &txid, const QString &me
     return QString::fromStdString(result);
 }
 
+void Wallet::getSpendProofAsync(const QString &txid, const QString &message, const QJSValue &ref) const
+{
+  QFuture<QString> future = QtConcurrent::run(this, &Wallet::getSpendProof, txid, message);
+  auto watcher = new QFutureWatcher<QString>();
```
xiphon:
```suggestion
  auto watcher = new QFutureWatcher<QString>(this);
```

src/libwalletqt/Wallet.cpp
```diff
@@ -680,6 +700,26 @@ QString Wallet::getTxProof(const QString &txid, const QString &address, const QS
     return QString::fromStdString(result);
 }
 
+void Wallet::getTxProofAsync(const QString &txid, const QString &address, const QString &message, const QJSValue &ref) const
+{
+    QFuture<QString> future = QtConcurrent::run(this, &Wallet::getTxProof, txid, address, message);
+    auto watcher = new QFutureWatcher<QString>();
```
xiphon:
```suggestion
    auto watcher = new QFutureWatcher<QString>(this);
```

src/libwalletqt/Wallet.cpp
```diff
@@ -662,6 +662,26 @@ QString Wallet::getTxKey(const QString &txid) const
   return QString::fromStdString(m_walletImpl->getTxKey(txid.toStdString()));
 }
 
+void Wallet::getTxKeyAsync(const QString &txid, const QJSValue &ref) const
+{
+    QFuture<QString> future = QtConcurrent::run(this, &Wallet::getTxKey, txid);
+    auto watcher = new QFutureWatcher<QString>();
```
xiphon:
```suggestion
    auto watcher = new QFutureWatcher<QString>(this);
```

main.qml
```diff
@@ -1003,11 +1003,21 @@ ApplicationWindow {
                     ", address: ", address,
                     ", message: ", message);
 
-        var result;
+        var spendProofFallback = (txid, result) => {
```
xiphon: Qt 5.9 doesn't support arrow functions

src/libwalletqt/Wallet.cpp
ph4r05: QFutureWatcher takes non-const QObject as a parent. In order to do this I had to make getSpendProofAsync non-const

ph4r05:
@xiphon I've made the modifications, pls check if it is OK

pages/History.qml
```diff
@@ -1577,19 +1574,32 @@ Rectangle {
         }
     }
 
+    function getTxKey(hash, elem){
+        if (elem.parent.state != 'ready'){
+            currentWallet.getTxKeyAsync(hash, function(hash, txKey) {
+                elem.parent.text = txKey ? txKey : '-';
+                elem.parent.state = 'ready';
+                toClipboard(txKey);
```
xiphon:
```suggestion
                toClipboard(txKey);
```

xiphon:
@ph4r05 

Looks okay, just one last thing https://github.com/monero-project/monero-gui/pull/2220#discussion_r295783325

ph4r05:
Done :)

PR #2227, opened 2019-06-18
[WIP] concurrency: always await all futures to complete

xiphon:


xiphon:
Closed in favor of more robust async operations handling approach #2229

PR #2229, opened 2019-06-20
qt: implement FutureScheduler, always await async code to complete

xiphon:


xiphon:
Updated. Implemented js callbacks support in FutureScheduler class (they were introduced in https://github.com/monero-project/monero-gui/pull/2220)

Issue #2204, opened 2019-06-08
Create a new password dialog


xmrdsc:
There are 3 QML components for displaying a password dialog.

```bash
$ ls components | grep -i pass
NewPasswordDialog.qml
PassphraseDialog.qml
PasswordDialog.qml
```

Should be combined into one, modular QML component.

Realchacal:
Not 100% sure if this belongs here:
![image](https://user-images.githubusercontent.com/46682965/59261898-87758e00-8c3e-11e9-8999-cb995a412fde.png)

Reasons:

A) Clicking on "Change wallet" doesn't change your wallet, it just takes you back to the window before, where you can do many things, FOR EXAMPLE changing your wallet.

B) The wording can be misinterpreted like this: When a user actually closes his wallet to open another wallet, he may think "Does I have to click 'change wallet' to confirm that I actually want to change my wallet?" (Of course not; He has to click "Continue" to change his wallet! How strange.)

selsta:
+resolved

PR #2240, opened 2019-07-01
PasswordDialog: merge components and bug fixes

selsta:
This commit merges PasswordDialog, PassphraseDialog and NewPasswordDialog. Also the following bugs were fixed:

- Wizard pages still being active when opening a wallet from wizard. (https://streamable.com/zqryj)
- Capslock detection was buggy when copy pasting password, I replaced it with native implementations for each platform.
- isAlpha could throw errors when using special characters.


xmrdsc:
Tested in Windows.

xmrdsc:
Digging into our issue tracker:

- Have you read the comment at #2204 ?
- Does this fix/close #2271 ?

PR #2246, opened 2019-07-01
QML: fix warnings

selsta:


pages/Keys.qml
```diff
@@ -120,7 +120,7 @@ Rectangle {
                 copyButton: true
                 labelText: qsTr("Block #") + translationManager.emptyString
                 fontSize: 16
-                text: currentWallet.walletCreationHeight
+                text: currentWallet ? currentWallet.walletCreationHeight : 0
```
MaxXor: When can `currentWallet` be null?

selsta: When no wallet is opened.

selsta: Fixes this warning:

```
2019-07-02 12:56:22.075	W qrc:/pages/Keys.qml:123: TypeError: Cannot read property 'walletCreationHeight' of undefined
```

MaxXor: It's interesting, because all other text fields are initialized here:
https://github.com/monero-project/monero-gui/blob/fc3a62688592aadc7f0f32788eca5d2ebcfe1fa3/pages/Keys.qml#L264-L269

Maybe move it to the same location to keep this consistent?

selsta: updated

MaxXor: Thanks.

PR #2247, opened 2019-07-02
WalletManager: async close with splash screen

xiphon:


selsta:
This change breaks the password screen when entering a false password.

Screenshot 2019-07-15 at 21 46 57

PR #2255, opened 2019-07-03
NetworkStatusItem: simple mode has no node settings page

selsta:


components/NetworkStatusItem.qml
```diff
@@ -133,7 +134,8 @@ Rectangle {
 
                 MouseArea {
                     anchors.fill: parent
-                    cursorShape: Qt.PointingHandCursor
+                    enabled: appWindow.walletMode >= 2
```
xiphon:
```suggestion
                    visible: appWindow.walletMode >= 2
```

xmrdsc: Yes.

selsta: Thanks, updated.

PR #2258, opened 2019-07-03
Tails: support persistence

xmrdsc:
Requires #2257 

This PR adds support for Tails and [its persistent volume feature](https://tails.boum.org/doc/first_steps/persistence/index.en.html).

At launch it will:
1. Detect Tails OS
    - expose bool `isTails` to QML
2. Detect if a Tails encrypted and persistent volume is mounted
    - If not, warn the user about settings that will not be saved
3. Ask if the user wants to point the GUI to this persistent volume:
    - expose bool `tailsUsePersistence` to QML
    - Write log files to `~/Persistent/Monero/monero-core.conf`
    - Create and use wallet directory: `~/Persistent/Monero/wallets`
    - Write the `monero-core.conf` configuration file to `~/Persistent/Monero/`
    - Register a persistent desktop startup entry

To sum up, this PR detects Tails and sets the correct paths.

When Tails is detected, but persistence is disabled, show a warning:

When Tails is detected with persistence enabled, ask the user to write to it:

(typo in image has been fixed)

This will set the correct paths inside QML:


src/qt/TailsOS.h
```diff
@@ -0,0 +1,20 @@
+#ifndef TAILSOS_H
+#define TAILSOS_H
+
+#include <QApplication>
+
+extern bool tailsUsePersistence;
```
selsta: Maybe something like this to avoid global variables?

```diff
diff --git a/Logger.cpp b/Logger.cpp
index 6c2ac9a..8b650e3 100644
--- a/Logger.cpp
+++ b/Logger.cpp
@@ -63,7 +63,7 @@ static const QString defaultLogName = "monero-wallet-gui.log";
 
 
 // return the absolute path of the logfile and ensure path folder exists
-const QString getLogPath(const QString logPath)
+const QString getLogPath(const QString logPath, const bool tailsUsePersistence)
 {
     const QFileInfo fi(logPath);
 
diff --git a/Logger.h b/Logger.h
index 7674a43..a178119 100644
--- a/Logger.h
+++ b/Logger.h
@@ -29,7 +29,7 @@
 #ifndef LOGGER_H
 #define LOGGER_H
 
-const QString getLogPath(const QString logPath);
+const QString getLogPath(const QString logPath, const bool tailsUsePersistence);
 void messageHandler(QtMsgType type, const QMessageLogContext &context, const QString &message);
 
 #endif // LOGGER_H
diff --git a/main.cpp b/main.cpp
index ded6c8c..817d2f6 100644
--- a/main.cpp
+++ b/main.cpp
@@ -137,6 +137,7 @@ int main(int argc, char *argv[])
     // - QML Settings file location (monero-core.conf)
     // - Default wallets path
     // Target directory is: ~/Persistent/Monero
+    bool tailsUsePersistence = false;
     if (isTails) {
         if (!TailsOS::detectDataPersistence())
             TailsOS::showDataPersistenceDisabledWarning();
@@ -181,7 +182,7 @@ int main(int argc, char *argv[])
     Monero::Utils::onStartup();
 
     // Log settings
-    const QString logPath = getLogPath(parser.value(logPathOption));
+    const QString logPath = getLogPath(parser.value(logPathOption), tailsUsePersistence);
     Monero::Wallet::init(argv[0], "monero-wallet-gui", logPath.toStdString().c_str(), true);
     qInstallMessageHandler(messageHandler);
 
diff --git a/src/qt/TailsOS.cpp b/src/qt/TailsOS.cpp
index ba68bef..c50a35c 100644
--- a/src/qt/TailsOS.cpp
+++ b/src/qt/TailsOS.cpp
@@ -6,8 +6,6 @@
 #include "TailsOS.h"
 #include "utils.h"
 
-bool tailsUsePersistence = false;
-
 bool TailsOS::detect()
 {
     if (!fileExists("/etc/os-release"))
diff --git a/src/qt/TailsOS.h b/src/qt/TailsOS.h
index ca62eb9..bb48614 100644
--- a/src/qt/TailsOS.h
+++ b/src/qt/TailsOS.h
@@ -3,8 +3,6 @@
 
 #include <QApplication>
 
-extern bool tailsUsePersistence;
-
 class TailsOS
 {
 public:
```

src/qt/TailsOS.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+bool TailsOS::detectDotPersistence()
+{
+    return QDir("/live/persistence/TailsData_unlocked/dotfiles").exists();
+}
+
+void TailsOS::showDataPersistenceDisabledWarning()
+{
+    QMessageBox msgBox;
```
selsta: The design of these doesn’t match with the rest of the GUI. While we use `MessageDialog` in some minor places, it’s just because they haven’t been replaced yet, the user doesn’t see them during normal use.

What’s your opinion on this?

src/qt/TailsOS.h
xmrdsc: No, I did that on purpose.

src/qt/TailsOS.cpp
xmrdsc: I think it is fine, this is only for Tails - before QML has started.

selsta: Right, it’s before QML even started. Ok then.

src/qt/TailsOS.h
selsta: Why?

selsta: If you don’t want to change existing functions then I would suggest at this: https://paste.debian.net/hidden/a96a06f9/

xmrdsc: Yes, rather not change function signature of `getLogPath`

Nice suggestion, cheers.

xmrdsc:
PR is ready for testing/review

PR #2272, opened 2019-07-05
use XDG_CACHE_DIR for the log file

rnhmjoj:
I maintain the monero and monero-gui packages in NixOS: this is a patch I have been using and wanted to upstream for quite some time.
I'm not sure about now, but the log defaulted to being written in the same directory of the binary (monero-wallet-gui), which is read-only in NixOS and probably other distribution.
This patch moves the log to the XDG cache directory.

### DIFF
--- a/Logger.cpp
+++ b/Logger.cpp
@@ -56,8 +56,8 @@ static const QString defaultLogName = "monero-wallet-gui.log";
     static const QString appFolder = "Library/Logs";
 #else // linux + bsd
     //HomeLocation = "~"
-    static const QString osPath = QStandardPaths::standardLocations(QStandardPaths::HomeLocation).at(0);
-    static const QString appFolder = ".bitmonero";
+    static const QString osPath = QStandardPaths::standardLocations(QStandardPaths::CacheLocation).at(0);
+    static const QString appFolder = "bitmonero";
 #endif
 
 
--- a/main.cpp
+++ b/main.cpp
@@ -152,7 +152,9 @@ int main(int argc, char *argv[])
     QCommandLineOption logPathOption(QStringList() << "l" << "log-file",
         QCoreApplication::translate("main", "Log to specified file"),
         QCoreApplication::translate("main", "file"));
-
+    logPathOption.setDefaultValue(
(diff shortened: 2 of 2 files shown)


MaxXor:
Hm, this has been fixed, I remember this was the case a few releases prior to 0.14. The GUI log is written to `~/.bitmonero/monero-wallet-gui.log` on Ubuntu and Arch Linux as far as I've tested these two.

Can you re-check with the latest version if it is still happening?

rnhmjoj:
You are right, looks like I'm late. It would be better to use XDG_CACHE_DIR, though.
That way the configuration directory doesn't get cluttered with unimportant files.

xmrdsc:
Thanks for bringing this up @rnhmjoj. As @MaxXor pointed out, I believe this is indeed already fixed.

I've been wanting to move the following files to `~/Monero` so that this directory contains:
- The log file(s)
- The wallet files
- The wallet configuration file

The above is for the sake of having as much stuff in one directory as possible (as opposed to scattered around the system). This makes removing the GUI easier.

Then again, a proper uninstaller (or uninstall procedure) can also be a solution.

I'm not hyped about `XDG_CACHE_DIR`. Seems like this is used for other purposes (?).

Let me know if the current tagged release still has this problem for you, so we can look for a solution.

rnhmjoj:
The idea of the XDG [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is to allow the separation of configuration files, user generate data and cache files. The main advantage is that it makes the locations configurable; also it's easier backup by simply selecting the directory (like $XDG_DATA_HOME/monero), without worrying about excluding temporary files.

I get the point about fragmentation but monero-gui is already using XDG_CACHE_HOME for the qml cache, and the log file certainly belong to this category. An "uninstall" procedure could simply be to delete:
`$XDG_{CACHE,DATA,CONFIG}_HOME/monero-project`

xmrdsc:
Thanks for the information @rnhmjoj. Sounds like a good idea. I've created an issue/proposal over at #2291.

PR #2236, opened 2019-06-28
Wizard: language and dialog updates

notmike-5:
#2211 
A more perfected solution would dynamically assign dialog
only when wallet creation from hw device is confirmed.

Issue #2208, opened 2019-06-10
"Mode selection"-window: Wording overhaul


Realchacal:
![monero3](https://user-images.githubusercontent.com/46682965/59207445-5eee8500-8ba7-11e9-8be7-ab2343bec5d0.jpg)

Issues:

1.) Relating to blockchain downloads the term "bootstrap" is commonly used for fast blockchain downloads over a 3rd-party (instead over the P2P protocol). Contrary to that, in Monero-GUI the term is used to describe the temporary connection to a remote node. This is unnecessary confusing for everybody in the blockchain space.

2.) The explanations for the Simple mode and Simple mode (bootstrap) fail to explain the real difference. For example, a user has no chance to know, that Simple mode (bootstrap) will allow him to use his wallet instantly. 

3.) It is not indicated, what the EFFECTIVE consequences of the different modes are. But that's what's most important for (new!) users.

4.) In the description of the Advanced mode it says "The blockchain is downloaded to your computer" - which is wrong, since the user has full control over the corresponding settings and can choose to use a remote node.

![monero 3 vorschlag](https://user-images.githubusercontent.com/46682965/59210146-2a7dc780-8bad-11e9-97e3-c71a94276c33.jpg)

xmrdsc:
I like the proposed changes.

notmike-5:
#2236

xmrdsc:
People IRC (`#monero-gui`) do not seem to like this change. Specifically; dEBRYUNE/sgp/ErCiccione. In order for this to go in, you'll have to convince multiple people - I don't have the energy for it so I'm closing my PR #2280

notmike-5:
I have also closed #2273 so as to satisfy the Monero Politboro and Central Committee.

xmrdsc:
That's kind of stretching it @notmike-5 ;-)

Realchacal:
That's why I gave 4 reasons why the changes are important, so that people can argue against it (instead of having to say "they don't like it" on IRC.)

Seriously: That bootstrap-mode is ANYTHING but self-explaining. And currently it is not explained at all! Sure, veterans don't care. But for newcomers it's a slap in the face.

dEBRUYNE-1:
@Realchacal - I was opposed to the changes mainly because it included the terms remote node and local node. New users are not acquainted with those terms and therefore they arguably should not be used. Do you have any suggestions on new wording without these terms?

Realchacal:
[...]

PR #2273, opened 2019-07-06
Wizard ask password dialog

notmike-5:
Moved to its own branch.
See #2236 
#2208

### DIFF
--- a/wizard/WizardModeSelection.qml
+++ b/wizard/WizardModeSelection.qml
@@ -57,16 +57,16 @@ Rectangle {
 
             WizardHeader {
                 title: qsTr("Mode selection.") + translationManager.emptyString
-                subtitle: qsTr("Please select the statement that best matches you.") + translationManager.emptyString
+                subtitle: qsTr("Please select the statement that best matches your desired setup.") + translationManager.emptyString
             }
 
             WizardMenuItem {
                 opacity: appWindow.persistentSettings.nettype == 0 ? 1.0 : 0.5
                 Layout.topMargin: 20
-                headerText: qsTr("Simple mode") + translationManager.emptyString
+                headerText: qsTr("Simple mode (quick)") + translationManager.emptyString
                 bodyText: {
                     if(appWindow.persistentSettings.nettype == 0){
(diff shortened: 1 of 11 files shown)


wizard/WizardModeSelection.qml
```diff
@@ -57,16 +57,16 @@ Rectangle {
 
             WizardHeader {
                 title: qsTr("Mode selection.") + translationManager.emptyString
-                subtitle: qsTr("Please select the statement that best matches you.") + translationManager.emptyString
+                subtitle: qsTr("Please select the statement that best matches your desired setup.") + translationManager.emptyString
```
erciccione: Is this change really needed? I'm fine with editing the text if necessary, but keep in mind that everytime a string is changed even slightly, the matching string in all the language files will be resetted and will have to be retranslated. This for *more than 30 languages*.

wizard/WizardModeSelection.qml
```diff
@@ -121,8 +121,8 @@ Rectangle {
             }
 
             WizardMenuItem {
-                headerText: qsTr("Advanced mode") + translationManager.emptyString
-                bodyText: qsTr("Includes extra features like mining and message verification. The blockchain is downloaded to your computer.") + translationManager.emptyString
+                headerText: qsTr("Advanced mode (maximum privacy)") + translationManager.emptyString
+                bodyText: qsTr("Includes extra features like mining and message verification. By default, the blockchain is downloaded to your computer.") + translationManager.emptyString
```
erciccione: See my comment above

wizard/WizardModeSelection.qml
notmike-5: I don't really have a preference. This came from an issue #2208. I didn't create a branch initially and now my downstream master has this modded file in it. I'm happy to discard it, close the issues, and reclone to the latest upstream master, but didn't want to do that before resolving.

wizard/WizardModeSelection.qml
notmike-5: If this is closed, it is related to #2236 which could also be closed.

PR #2276, opened 2019-07-08
Fixes #2275, transactionCommited txid list

### LINKED ISSUE #2275: ver 0.14.1.0 Wallet shows "Monero sent succesfully: 0 transaction(s)" after sending a transaction
Tried to send a tx with Monero-gui using Trezor. 
The process went normally and the transaction was actually sent, but the wallet showed the following:
Screenshot 2019-07-08 at 09 07 48

ph4r05:
txid list was empty when emitted

### DIFF
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -524,7 +524,8 @@ bool Wallet::submitTxFile(const QString &fileName) const
 void Wallet::commitTransactionAsync(PendingTransaction *t)
 {
     m_scheduler.run([this, t] {
-        emit transactionCommitted(t->commit(), t, t->txid());
+        auto txIdList = t->txid();  // retrieve before commit
+        emit transactionCommitted(t->commit(), t, txIdList);
     });
 }
 
--- a/src/libwalletqt/Wallet.h
+++ b/src/libwalletqt/Wallet.h
@@ -351,7 +351,7 @@ signals:
     void walletCreationHeightChanged();
     void deviceButtonRequest(quint64 buttonCode);
     void deviceButtonPressed();
-    void transactionCommitted(bool status, PendingTransaction *t, QStringList txid);
+    void transactionCommitted(bool status, PendingTransaction *t, const QStringList& txid);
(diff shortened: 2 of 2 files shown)


xiphon:
Does the issue happen every time you send a tx (without the fix)?

ph4r05:
Everytime I tried.

xiphon:
Can't reproduce it with and without Ledger. Don't have Trezor device.

Could you please test the fix without the following change?

```diff
diff --git a/src/libwalletqt/Wallet.cpp b/src/libwalletqt/Wallet.cpp
index 46f843de8..2df148854 100644
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -524,7 +524,8 @@ bool Wallet::submitTxFile(const QString &fileName) const
 void Wallet::commitTransactionAsync(PendingTransaction *t)
 {
     m_scheduler.run([this, t] {
-        emit transactionCommitted(t->commit(), t, t->txid());
+        auto txIdList = t->txid();
+        emit transactionCommitted(t->commit(), t, txIdList);
     });
 }
 ```

**Comment by ph4r05** [comment]
I can give it another try, but without the change you mentioned the bug is present.

**Comment by ph4r05** [comment]
@xiphon I could reproduce this with software wallet only with the officially released v0.14.1.0 for OSX.

Screenshot 2019-07-15 at 15 11 30

**Comment by selsta** [comment]
```
15:16 <selsta> xiphon: ph4r05: I can reproduce the 0 transactions sent issue with and without Trezor.
15:17 <selsta> It didn’t happen a few days ago so it probably is some kind of race condition.
15:28 <selsta> #2276 seems to solve the issue for me.
```

**Comment by ph4r05** [comment]
Honestly I still don't really understand the nature of the problem. 

When I remove the PR changes and add logging to the `main.qml`:

```js
function onTransactionCommitted(success, transaction, txid) {
    console.log('res: ', txid, success);
```

The logging shows an empty array. After the PR fix it shows the array correctly. 

I figured out the txid has to be retrieved before commit.

ph4r05:
@selsta @xiphon I've updated the comment explaining why the PR works

Issue #1309, opened 2018-04-12
Support 4K displays


Warchant:
This is how `monero-gui` (`v0.12.0.0`, Windows 10) looks like on my laptop's 4K display (Dell XPS 15 9560):

![image](https://user-images.githubusercontent.com/1867551/38661994-7b73c1fa-3e3b-11e8-8d5a-a026bebca1df.png)

![image](https://user-images.githubusercontent.com/1867551/38662013-8eae9c22-3e3b-11e8-9128-721c19618883.png)

Please, add scaling.

SamsungGalaxyPlayer:
Yes, follow the instructions as @medusadigital said, selecting "System" or "System (Advanced)".

pazos:
[HiDPI support for windows](https://vicrucann.github.io/tutorials/osg-qt-high-dpi/) seems [portable to X11/android](http://blog.qt.io/blog/2016/01/26/high-dpi-support-in-qt-5-6/).  a little different for [osx](http://blog.qt.io/blog/2013/04/25/retina-display-support-for-mac-os-ios-and-x11/)

please leave this issue open as an enhancement,
thanks!

skftn:
+enhancement

pazos:
@Warchant : I think I found a way to fix that in Windows. Since I don't have any hiDPI screens I cannot test. Would you want to test that for me? In that case I will submit a PR :taco:

valiant1x:
@pazos we would also appreciate your test [here](https://github.com/valiant1x/intensecoinwallet) - feel free to submit a PR :)

pazos:
Sorry, I have no incentive to do stuff for your forked coin. But if you have a windows machine and a hidpi screen and want to test stuff please ping me :slightly_smiling_face:

valiant1x:
We contribute upstream as well when able. Open source doesn't have to be a one way street! In any case, I am on a 1440p monitor and am happy to test.

QuickBASIC:
I just got a high density screen and there's a way we could work around this on Windows without mucking with QT DPI scaling.

@medusadigital is on the right track, but it might be confusing to new users to have to do this on first run. (4k screens are becoming more and more common, my cheap $1300 laptop has one, so it's not just enthusiasts, programmers, and artists that might end up having one.)
 
We could have the GUI provide an option to write the compatibility options to the registry directly w/ a command line switch and/or a checkbox in the GUI.

For an individual user: 
[...]

PR #2277, opened 2019-07-09
main: add high DPI for Linux and Windows

selsta:


erciccione:
I guess this will solve #1309, but will it help for #1273?

selsta:
This need further work by someone who has a high DPI screen.

PR #2280, opened 2019-07-15
Refactor mode selection text

### LINKED ISSUE #2208: "Mode selection"-window: Wording overhaul

xmrdsc:
Closes #2208

Changes mode selection text a bit.

### DIFF
--- a/wizard/WizardModeSelection.qml
+++ b/wizard/WizardModeSelection.qml
@@ -66,7 +66,7 @@ Rectangle {
                 headerText: qsTr("Simple mode") + translationManager.emptyString
                 bodyText: {
                     if(appWindow.persistentSettings.nettype == 0){
-                        return qsTr("Easy access to sending, receiving and basic functionality.") + translationManager.emptyString;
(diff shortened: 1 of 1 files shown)


xmrdsc:
People on IRC `#monero-gui` do not seem to like this change. Specifically; dEBRYUNE/sgp/ErCiccione. Closing this and #2208

Issue #2290, opened 2019-07-16
Improve password dialog button wordings


selsta:
Not 100% sure if this belongs here (please tell me):
![image](https://user-images.githubusercontent.com/46682965/59261898-87758e00-8c3e-11e9-8999-cb995a412fde.png)

Reasons:

A) Clicking on "Change wallet" doesn't change your wallet, it just takes you back to the window before, where you can do many things, FOR EXAMPLE changing your wallet.

B) The wording can be misinterpreted like this: When a user actually closes his wallet to open another wallet, he may think "Do I have to click 'Change wallet' to confirm that I actually want to change my wallet now, or do I have to click 'Continue' to confirm that I want to continue to use my previous wallet?"

_Originally posted by @Realchacal in https://github.com/monero-project/monero-gui/issues/2204#issuecomment-500773387_

PR #2297, opened 2019-07-18
[Dialog] button mod from Change wallet/Continue to Cancel/Ok

notmike-5:
see #2290 #2204

### DIFF
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -390,7 +390,7 @@ Item {
                 MoneroComponents.StandardButton {
                     id: cancelButton
                     small: true
-                    text: root.walletName.length > 0 ? qsTr("Change wallet") + translationManager.emptyString : qsTr("Cancel") + translationManager.emptyString
+                    text: qsTr("Cancel") + translationManager.emptyString
                     KeyNavigation.tab: passwordInput1
                     onClicked: {
                         root.close()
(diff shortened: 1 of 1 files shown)


notmike-5:
> @notmike-5 could you plz explain the purpose of the conditional statement in line 393 (and why we don't need that anymore)?

Sure, I'll give it a shot. This PR addresses #2290 ref from #2204 .  Of the buttons, you suggested in those issues, "It should say 'Cancel' and 'Ok.' notmike has made for you.  

_**components/PasswordDialog.qml**_
Line 393 is a ternary operator. Prior to this change, the Monero component `cancelButton` checks if the root component property `walletName` has a length greater than zero. If this evaluates true, We assume there is a wallet that the app will be trying to open, and the button text displays 'Change wallet.' If false, We assume there is not a wallet, a wallet has not been found, or perhaps a wallet failed to load properly, and in that case the button text reads 'Cancel.' 

After the change this button simply reads 'Cancel,' and the text on the other button is changed from 'Continue' to 'Ok.' 

Since the point of a wallet app is to generate and peruse wallets, which themselves generate and store keys, it makes sense for the button to read 'Change wallet' as that is the  telos of clicking that button, namely to access some other wallet. To say that we are canceling the operation of opening one wallet in order to open another is tautological. Others may feel the modified language is simpler to understand or that unessential or potentially confusing information is thus abstracted away. 

'Change wallet'/'Continue,' 'Cancel'/'Ok,' this is really a matter of preference at the end of the day.

PR #2302, opened 2019-07-20
Account: fix theme switch label color

### LINKED ISSUE #2301: Light theme bug: Some white text on white background on `Account` page
- Go to the `Account` page.
- Move the mouse over the numbers for "Total balance" and "Total unlocked balance" so that the numbers turn orange.
- Switch to the light theme

Now the two numbers are unreadable (white text on white background).

The same bug applies vice versa when switching to the dark theme.

selsta:
Fixes #2301


MaxXor:
I guess this happens also with all other labels that use `Style.dimmedFontColor`?

selsta:
`themeTransition: false` was the problem.

PR #2307, opened 2019-07-21
QML: fix various warnings

selsta:


xmrdsc:
Should probably be rewritten to move away from anchors to Layout. Approving this for now.

PR #2329, opened 2019-07-31
SettingsWallet: set trusted daemon before rescan

selsta:
When opening the GUI, the trusted daemon flag doesn’t get set properly. Easy fix: set it manually before rescanning.

Reported here: https://monero.stackexchange.com/questions/11512/rescan-ledger-wallet-from-0-in-the-gui


selsta:
Also ping @xiphon, maybe you can fix this properly because this is more of a hack :)

xiphon:
@selsta 

Thanks, see #2336

PR #2336, opened 2019-08-06
Wallet: explicitly set trusted daemon flag

xiphon:
> @selsta
> 
> When opening the GUI, the trusted daemon flag doesn’t get set properly. Easy fix: set it manually before rescanning.
>
> Reported here: https://monero.stackexchange.com/questions/11512/rescan-ledger-wallet-from-0-in-the-gui


xiphon:
Minor update, added `isTrustedDaemon` function

PR #2337, opened 2019-08-06
build: set submodule to master

selsta:


luigi1111:
I think this PR is unnecessary.

selsta:
You can also revert an older commit.

Issue #2344, opened 2019-08-09
Improvements when denying "Export View Key" on hardware wallet


Realchacal:
[Updated]
![image](https://user-images.githubusercontent.com/46682965/64254398-9a3aee80-cf1f-11e9-99c0-b64f049a1c1f.png)
When you deny "Export View Key"...
- the user should be informed _"You have denied the export of your public view key, therefore calculations will have to be done on your hardware wallet. This may take a very long time."_
- `Wallet blocks remaining: 12345 blocks` should be updated every block - not every 999 blocks. (1 block takes about 1 second with Ledger Nano S and subaddress lookahead 1:1)
- clicking on the "close wallet" icon in the balance card shouldn't close the whole GUI.

selsta:
When you deny it, the calculations are all done on your Ledger and scanning gets really slow. Last time I tried it, it was unusably slow.

Realchacal:
Experiment:
- Fully synced blockchain
- Restoring from Ledger 
- Subaddress lookahead 1:1
- Restore height = `current block height - 100`

Results:
- In the beginning it said `Wallet blocks remaining: 1092 blocks` (I expected 100)
- `Wallet blocks remaining: 1092 blocks` didn't change for ~18 minutes
- After a total time of ~20 minutes the wallet was synced
- Clicking the door symbol still closes the whole GUI instead of just the wallet

Conclusions:

This feature only seems usable for new wallets/ super young wallets and with subaddress lookahead 1:1, otherwise it could easily take multiple years to restore a wallet. If we want to keep the future, we should:
- inform the user: `You have denied the export of your public view key, therefore calculations have to be done on your hardware wallet. This may take an extremely long time.`
- update the `Wallet blocks remaining: 1092 blocks` more frequently (every block)
- we should fix the door button

Realchacal:
Thx! I updated the issue.

Realchacal:
I guess nobody needs this anyway, right? Close?

selsta:
AFAIK the GUI doesn’t know if the view key gets exported or not. Changing this behavior isn’t trivial, I think it’s unlikely that this will get changed soon. People export their view key anyway.

Realchacal:
I see, and I wouldn't recommend putting any work in it. So close, right?

Realchacal:
Closed because not really important. Let's hope noobs just will click "export".

Issue #2324, opened 2019-07-27
Wording overhaul "Welcome to Monero" screen


Realchacal:
![image](https://user-images.githubusercontent.com/46682965/61990838-089ab000-b048-11e9-96f6-210075c91c74.png)

Reasons:

- What is "from hardware"?! Maybe from my computer? Or maybe Monero is special here and I'm missing something?
- The small text should be used CONSISTENTLY either to explain OR to give commands. Not a mix of both. We should use it to explain, like in row 1.
- The wording must aim at NEW/ UNEXPERIENCED users. Pros will understand it anyways. So for example let's not talk of "Importing *.keys files". For a noob that sounds like some special action ("Importing") where he has to have some special file (a "keys file") and he must know the file path - while actually this is just the standard option if you have used Monero before. (Furthermore, it's enough to tell in the next window that these files are *.keys files.)

selsta:
4 times “Choose this option”, is this an improvement?

Realchacal:
Thank's for your feedback!! Updated!

selsta:
I’m not happy with the 3rd one.

“Choose this option if you have used Monero before”, can be confusing. You can have used Monero on a different computer before.

What about: “Choose this option to open an existing Monero wallet from your computer.” ?

Realchacal:
Your argument is good! What about leaving "from your computer", though?
--> "Choose this option to open an existing Monero wallet." ?

selsta:
Ok. I’ll wait on other opinions before I implement it though.

Realchacal:
Sure! Updated.

rating89us:
Some considerations:
- "Open a wallet from file" can be used both for users that have and don't have hardware wallets
- "Restore wallet from keys or mnemonic seed" can only be used for Monero wallets that don't support hardware. Ledger/Trezor users can't type device seed in this page. Also, hardware wallets don't inform their private keys to users.

So, I believe there should be a clear division between "normal" and "hardware wallets" users:

**I don't have a hardware wallet:**
- **Create a new wallet** (_for users that just want to create a new Monero wallet that doesn't support hardware wallet_)
- **Open a wallet from file** (_for users that have a wallet file that doesn't support hardware wallet_)
- **Restore wallet from keys or mnemonic seed** (_for users that want to restore from their keys/mnemonic seed_)

**I have a hardware wallet:**
[...]

PR #2353, opened 2019-08-16
Ledger: improve hw wallet splash screen

selsta:
Screenshot 2019-08-16 at 21 12 49

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -859,7 +859,9 @@ ApplicationWindow {
                     ", priority: ", priority,
                     ", description: ", description);
 
-        showProcessingSplash("Creating transaction");
+        var splashMsg = qsTr("Creating transaction...");
+        splashMsg += appWindow.currentWallet.isLedger() ? qsTr("\n\nPlease check your hardware wallet –\nyour input may be required.") : "";
+        showProcessingSplash(splashMsg);
 
         transactionDescription = description;
 
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -258,6 +258,11 @@ bool Wallet::isHwBacked() const
     return m_walletImpl->getDeviceType() != Monero::Wallet::Device_Software;
 }
 
+bool Wallet::isLedger() const
+{
+    return m_walletImpl->getDeviceType() == Monero::Wallet::Device_Ledger;
+}
+
 //! create a view only wallet
 bool Wallet::createViewOnly(const QString &path, const QString &password) const
 {
--- a/src/libwalletqt/Wallet.h
+++ b/src/libwalletqt/Wallet.h
@@ -180,6 +180,7 @@ public:
 
     //! hw-device backed wallets
     Q_INVOKABLE bool isHwBacked() const;
+    Q_INVOKABLE bool isLedger() const;
 
     //! returns if view only wallet
     Q_INVOKABLE bool viewOnly() const;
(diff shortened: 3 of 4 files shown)


Realchacal:
Nice!!!!!!!!

selsta:
Updated PR, it now also shows while transacting. Only wallet opening is missing now.

Realchacal:
Nice!!! And hard to understand why the opening screen is any more difficult to change? (I'm not complaining in any way, I really just wonder)

selsta:
To find out if a wallet is a Ledger wallet is has to get opened first, the splash screen gets displayed before a wallet is opened.

Realchacal:
Oh, I see! Interesting, thanks!

selsta:
> This could be solved if we went with the alternative proposal shown in #2324.

This wouldn’t work as I need to differentiate between Trezor and Ledger. Trezor already shows a splash screen if user input is required. So I’d need to know if it’s a Ledger wallet specifically.

Realchacal:
I see, thanks! (By the way that's why I suggested "...your input **may** be required..." - so that it will work with different hardware wallets in the future and also if a hardware wallet has an option like "export public view key without confirmation")

main.qml
```diff
@@ -859,7 +859,9 @@ ApplicationWindow {
                     ", priority: ", priority,
                     ", description: ", description);
 
-        showProcessingSplash("Creating transaction");
+        var splashMsg = qsTr("Creating transaction...");
+        splashMsg += appWindow.currentWallet.isLedger() ? qsTr("\n\nPlease check your hardware wallet –\nyour input may be required.") : "";
```
xiphon: Do we need to end the sentence with a period here?

Realchacal: I'd be fine with or without period.

Realchacal:
I wrote the new sentence under the assumption that it must fit to all cases (Ledger/ Trezor/ other future hardware wallets, which have an option like: `[x] export public view key without confirmation.` ) If that is **not** the case (as it does look now), the sentence isn't that good anymore. Better:

- Please confirm the export of your view key on your hardware wallet.
(If you deny it, computations will have to be done on your hardware wallet, which may take an extreme long time.)

And if we wanted it perfect, we don't need "Creating wallet from device..." at this point IMO.

Realchacal:
I phrased the added sentence under the assumption that it must fit in all cases, even if no user action is required. But if my assumption is wrong (as it seems), we'd better go like this:

- **Please confirm the export of your view key on your hardware wallet.**
(If you deny it, computations will have to be done on your hardware wallet, which may take a very long time.)

Not sure if we should keep the part in brackets. Because denying export view key is clearly a feature for pro users who know what they are doing since this may easily take weeks (see #2344).

For the transaction splash screen:
- **Please confirm the transaction details on your hardware wallet.**
(The change amount that is shown to you will not be spendable for ~20 minutes.)

Not sure if we should keep the part in brackets. But since the change amount may be a huge number this confirmation could easily cause anxiety for an average user who does not understand what this means.

main.qml
selsta: I think it looks better with a period, but I don’t care really.

selsta:
@Realchacal I prefer it being like it is in this PR.

> Please confirm the transaction details on your hardware wallet.

This can confuse the user. It takes up to 10 seconds until the dialog shows up on the Ledger.

Realchacal:
> It takes up to 10 seconds until the dialog shows up on the Ledger.

Strong argument, I wasn't aware! Thanks.

Issue #2339, opened 2019-08-06
Improve title bar icons


Realchacal:
[Updated!]

- Currently, the "show/hide menu" icon doesn't give away what it's about.
- Currently, the "close wallet" icon is misplaced in the balance card because it doesn't close the "balance" or the account.

![image](https://user-images.githubusercontent.com/46682965/64287649-895e9d00-cf60-11e9-80c3-2e9cb48f6981.png)

xiphon:
https://github.com/monero-project/monero-gui/pull/2358

Realchacal:
Updated.

Realchacal:
Do I close this since we have the moon now? Or does anybody see any merit in the rest?

selsta:
I think the current globe icon is fine. A Google search for “language icon” shows both a globe and Chinese characters.

Regarding the flag as an alternative, see https://github.com/monero-project/monero-gui/issues/1092.

Realchacal:
> I think the current globe icon is fine. A Google search for “language icon” shows both a globe and Chinese characters.

Damn you're right!! Convinced with regard to the globe.

PR #2358, opened 2019-08-24
TitleBar: change Light Theme icon from 'light bulb' to 'moon'

xiphon:
Requires and to be merged after #2372

light-theme-icon

### DIFF
--- a/components/TitleBar.qml
+++ b/components/TitleBar.qml
@@ -156,8 +156,8 @@ Rectangle {
             Layout.preferredHeight: parent.height
 
             Text {
-                text: MoneroComponents.Style.blackTheme ? FontAwesome.lightbulbO : FontAwesome.moonO
-                font.family: FontAwesome.fontFamily
+                text: FontAwesome.moonO
+                font.family: MoneroComponents.Style.blackTheme ? FontAwesome.fontFamilySolid : FontAwesome.fontFamily
                 font.pixelSize: 16
                 color: MoneroComponents.Style.defaultFontColor
                 anchors.verticalCenter: parent.verticalCenter


selsta:
Two similar looking circles next to each other don’t look good in my opinion :P

xiphon:
> Two similar looking circles next to each other don’t look good in my opinion :P

Yep. Any idea on better replacement for `light bulb` (or `world`) icon?

Realchacal:
> Yep. Any idea on better replacement for `light bulb` (or `world`) icon?

- **I like the new symbol.** That said, the more intuitive symbol would be a sun/moon. Google image search: "brightness symbol", "day night mode symbol" - there's no doubt about what people are used to.
- Language symbol (less complex than #2339):
image

xiphon:
@Realchacal

Regarding available icons - you can browse [Font Awesome](https://fontawesome.com/icons) icons (free only). That is what we are using.

Realchacal:
> Regarding available icons - you can browse [Font Awesome](https://fontawesome.com/icons) icons (free only). That is what we are using.

Thanks man! Is that our exclusive source?

xiphon:
> Thanks man! Is that our exclusive source?

I would say it is preferable, although we can introduce some handmade icon, but would be better not to do that and use something from the icons pack.

> Can we just use the right part?

Doesn't seem to be possible.

Realchacal:
Ok then the last possibility I see (given the limited source):
image
IMO by far not as good as the Chinese character, but better than the world icon. (I barely can recognize the world as a world, and to me it means "Internet" anyway)

Realchacal:
I propose going with these 2 for nightmode on/off:
image

Rationale:
- It's more correct than the adjust icon, which is commonly used to **adjust** the brightness (smoothly).
- I think it's intuitive because when people look for this feature they have "**night** mode" in mind (even when they want to turn the night mode off).
- It's easier if you always have to look for the same icon.
- The moon icon is used by many other sites (blockstream.info, poloniex, binance).

For the language **MAYBE** we could go like this:
image
Don't know if I like it, though. The Chinese character would be so much better. But the world is not good IMO. (Hard to recognize, and it also stands for "Internet/WWW")

Realchacal:
I propose this for nightmode on/off:
image

Rationale:
- It's more correct than the adjust icon, which is commonly used to **adjust** the brightness (smoothly).
- I think it's intuitive because when people look for this feature they have "**night** mode" in mind (even when they want to turn it off).
- It's easier if you always have to look for the same/similar icon.
- Other websites have that, too, for example https://blockstream.info/
- We can keep the exact icon that we are currently using for the light theme.

For the language maybe we could use this:
image
Unfortunately, not as intuitive as the Chinese character, but definitely better than the world icon (The world icon is hard to recognize, and it also stands for "Internet/WWW".)

This is how it could look like:
image
(Ignore the "X". That's just where I'd put the "Close Wallet" button because it closes the whole wallet, not just the balance/account.)

Realchacal:
Latest proposals from #2339:
image

image

Realchacal:
Latest proposal in accordance with #2339 (updated):
image

xiphon:
Updated. Using `moon` icon instead.

Realchacal:
> Updated. Using `moon` icon instead.

Very nice, but it would be nicer if the moon in light theme wasn't filled!

xiphon:
Just noticed that for some reason we are using outdated [2017 Font Awesome version](https://fontawesome.com/v4.7.0/). It doesn't have solid moon icon. 

Updated once again.

selsta:
Would be nice if we could update to the latest Font Awesome version.

xiphon:
> Would be nice if we could update to the latest Font Awesome version.

On it. Will send a separate PR in a while and then we will be able to use both moon versions here.

Realchacal:
> On it. Will send a separate PR in a while and then we will be able to use both moon versions here.

Awesome!! It's gonna feel like "Ah, it's shining bright tonight!" every time when switching to dark mode!

xiphon:
Updated, based on the recent Font Awesome 5.10.2

Realchacal:
Absolutely perfect. ABSOLUTELY perfect!!

PR #2362, opened 2019-08-28
fix invalid include paths, compilation warnings, update README.md

xiphon:


src/qt/utils.cpp
```diff
@@ -29,7 +29,7 @@
 #include <QtCore>
 #include <QApplication>
 
-#include "src/qt/TailsOS.h"
+#include "qt/TailsOS.h"
```
MaxXor: shouldn't this be just "TailsOS.h" like the "utils.h" below?

xiphon: Yep, thanks, fixed.

xiphon:
> Weird thing with the invalid include paths. Still compiles here, but the new paths don't look more valid. Just check out https://github.com/monero-project/monero-gui/blob/master/main.cpp#L63-L70 it looks like this complete mixture is all okay. smile

From the compiler's point of view it is all okay. Though looks weird for the human being like we are since it doesn't maintain include declaration consistency across the same `main.cpp` file.

I didn't touch places like this one cause it doesn't break the compilation, more like a styling issue.

PR #2375, opened 2019-09-04
fontawesome: fix regular/solid font style

selsta:


selsta:
Updated.

fonts/FontAwesome/FontAwesome.qml
```diff
@@ -17,9 +18,8 @@ Object {
         source: "./fa-solid-900.ttf"
     }
 
-    property string fontFamily: "FontAwesome"
+    property string fontFamily: regular.name // regular and solid have the same name
```
xiphon: Regular and Solid font names differ on Windows, for example.

selsta:
Updated

Issue #2298, opened 2019-07-18
Improve balance readability


Realchacal:
![monero2](https://user-images.githubusercontent.com/46682965/61470336-28174600-a981-11e9-85a6-3f858e653f41.jpg)

People LOVE to see their balance, so let's make it look beautiful! And helpful! It's SO easy to get confused, since half of the world uses decimal point, other half decimal comma. And "1.000" could also be interpreted as 1000 in many places. And with the lack of a unit it may get really confusing. Not having a unit is just wrong!

(Details: Main number is bold. Decimal places are not bold and 14% smaller in font size. The unit has the same font size as the main number, but is not bold.)

GBKS:
I really like how you're simplifying things. What do you think of using Roboto Mono for the numbers and replacing the word "Monero" with the account name (something I also suggested [here](https://www.reddit.com/r/Monero/comments/c23zxn/feedback_request_ideas_for_improving_how_multiple/))? Something looks a little off still about "XMR" being bigger next to the smaller numbers...

![monero-balance-card-readability-190719-gbks](https://user-images.githubusercontent.com/695901/61519666-bab0f700-aa0c-11e9-8f75-b3c36c5eb277.png)

Realchacal:
@GBKS You are so right! My "XMR" was too big, and we should simplify even more! Thanks for your great input!

Plz check out my proposal 3 (yours is 2, my first one is 1):
![monero 5](https://user-images.githubusercontent.com/46682965/61524248-627ef280-aa16-11e9-8164-349773133555.jpg)

GBKS:
My concern would be that longer account names don't fit in. Somebody at Google told me a while back that German is 60% longer than English on average, so they always assume that that much more text has to fit. Here's an example of putting the account number and name in separate lines, with some rulers for the layout to keep things visually structured.

![monero-balance-card-readability-2-190719-gbks](https://user-images.githubusercontent.com/695901/61525473-c86c7980-aa18-11e9-8c1c-f810c4b83dfe.png)

[...]

PR #2383, opened 2019-09-08
LeftPanel: new balance card

selsta:
To do in future PRs:
- Explore showing the fiat balance all the time instead of toggle
- Close wallet button in titlebar
- Better unlocked balance explanation.

Screenshots:
-  Light mode: https://user-images.githubusercontent.com/7697454/64489922-801f4a00-d258-11e9-820d-d897c99700b3.png
- Dark mode: https://user-images.githubusercontent.com/7697454/64489923-801f4a00-d258-11e9-9155-2eb37200e2cb.png
- During sync: https://user-images.githubusercontent.com/7697454/64490079-d9d44400-d259-11e9-9934-cd5b927462f4.png
- Fiat conversion: https://user-images.githubusercontent.com/7697454/64490087-e22c7f00-d259-11e9-82fd-a51d536c1faa.png
- Unlocked balance: https://user-images.githubusercontent.com/7697454/64490094-ff614d80-d259-11e9-8716-9f44350e9849.png

Referencing: #2298


Realchacal:
Nice one!

1. It's not me, but the balance card "color swap" is going to be attacked from all sides... that's just what's gonna happen.

2. As @GBKS [teached me](https://github.com/monero-project/monero-gui/issues/2298#issuecomment-513164364): 
image
So your `EUR` is perfect while your `XMR` needs a little shrinking.

xiphon:
Gray card (https://github.com/monero-project/monero-gui/issues/2298#issue-469853014) looks better to me. Or at least the current brown one.

GBKS:
@Realchacal just a really small comment that the brown is a result of mixing the Monero orange and grey/black together to avoid a monotonous grey/black look. Probably just not obvious at all.

Realchacal:
@GBKS Ah ok I get the idea, but the resulting brown just doesn't remind of the orange. Any chance that you can come up with something grayish that doesn't look disabled?

selsta:
Updated and removed the reverted color cards. Screenshot includes #2298.

https://user-images.githubusercontent.com/7697454/65189573-a8395500-da71-11e9-80e7-c14b26da15c0.png
https://user-images.githubusercontent.com/7697454/65189574-a8d1eb80-da71-11e9-8f8e-e8a8586b1326.png

Feedback? I think it would be nice if we changed the white card to more of a grey. Ping @GBKS, what is your opinion on the white card?

Realchacal:
VERY NICE!!! 

This font size could be reduced a bit:
image

Changing the white card to grey sounds interesting. (Changing the brown card to grey seems more important to me, though.) But since that's a matter of taste maybe we could leave it as it is right now and make a poll some time later.

selsta:
> This font size could be reduced a bit:

The problem is I’m using Qt/QML and not something like InDesign. Getting everything pixel perfect isn’t trivial :/ I can reduce the font size but then `EUR` or `USD` wouldn’t fit perfectly. Don’t know how to solve this yet.

Yes, I’m okay with merging this for now and discussing other card colors later.

Realchacal:
> I can reduce the font size but then `EUR` or `USD` wouldn’t fit perfectly.

Oh I didn't think of that! We could prioritize `XMR` over `EUR`/`USD` given that we'll maybe switch to a 2-balance solution anyway, as [wished by users](https://www.reddit.com/r/Monero/comments/cynicf/poll_how_would_you_prefer_your_fiat_balance_to_be/).
image

Realchacal:
> I can reduce the font size but then `EUR` or `USD` wouldn’t fit perfectly.

I'd prioritize `XMR` over `EUR`/`USD` since [users want](https://www.reddit.com/r/Monero/comments/cynicf/poll_how_would_you_prefer_your_fiat_balance_to_be/) an opt-in 2-balance solution anyway:
image
Even if that doesn't happen, we should still prioritize `XMR` over `EUR`/`USD`.

Also, using a monospaced font could solve this problem.

selsta:
> I'd prioritize XMR over EUR/USD since users want an opt-in 2-balance solution anyway:

Almost all text elements in the balance card are font size 16, changing the “XMR” to 15 would look inconsistent. I think getting this merged first is more important than making this alignment pixel perfect :P

> Another thing:

Thanks, fixed.

Realchacal:
> changing the “XMR” to 15 would look inconsistent

valid argument, but not really, see:
image

> I think getting this merged first is more important

Absolutely, so I'd be fine with letting it as it is!

PR #2391, opened 2019-09-16
TitleBar: add close wallet button

selsta:
Screenshot 2019-09-16 at 23 57 15
To be used in combination with: #2383

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -171,44 +171,6 @@ Rectangle {
                     color: "#ff9323"
                     themeTransition: false
                 }
-
-                Rectangle {
-                    height: (logoutImage.height + 8)
-                    width: (logoutImage.width + 8)
-                    color: "transparent"
-                    anchors.right: parent.right
-                    anchors.rightMargin: 8
-                    anchors.top: parent.top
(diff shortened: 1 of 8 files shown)


selsta:
also ping @Realchacal

Realchacal:
Extremely nice!

Just to understand it: You don't like the `x` because we already have a `x` in the titlebar, right?
image
That's a valid argument. Not sure if it's enough to outweigh the reduced intuitiveness, though.

selsta:
Yes, 2 x icons would be bad IMO. Sign out is similar to the one we already have and was the best I found on fontawesome.

Realchacal:
Ok! One last thing: IMO the moon is a **LITTLE** bit too big. (I know it's exactly the same height as the globe, but to my eye it looks too big, though.)

selsta:
I reduced the moon font size by 1 point.

Realchacal:
I already had noticed it in your other screenshots ;) LOOKS SO MUCH BETTER!!!

Realchacal:
Fixes #2339

luigi1111:
@selsta please rebase.

selsta:
@luigi1111 done

PR #2393, opened 2019-09-19
windeploy: bump icu version to 64, add libdouble-conversion.dll

stoffu:
Dependency on libcouble-conversion.dll seems to have been introduced here: https://github.com/msys2/MINGW-packages/commit/ec9c7ac1c80476bf386d165c63fec88cba7415d5

### DIFF
--- a/windeploy_helper.sh
+++ b/windeploy_helper.sh
@@ -17,13 +17,13 @@ fi
 
 if [[ "$BUILD_TYPE" == "Release" ]]; then
 	echo "Release build"
-	ICU_FILES=(libicudt62.dll libicuin62.dll libicuio62.dll libicutu62.dll libicuuc62.dll)
+	ICU_FILES=(libicudt64.dll libicuin64.dll libicuio64.dll libicutu64.dll libicuuc64.dll)
 else
 	echo "Debug build"
-	ICU_FILES=(libicudtd62.dll libicuind62.dll libicuiod62.dll libicutud62.dll libicuucd62.dll)
+	ICU_FILES=(libicudtd64.dll libicuind64.dll libicuiod64.dll libicutud64.dll libicuucd64.dll)
 fi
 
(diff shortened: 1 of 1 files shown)


selsta:
Do you know if this also works with Qt 5.9?

stoffu:
@selsta 
I’m not sure what you mean, my intent here is to make it work with the latest Qt shipped with MINGW which is 5.12. Why do you care about 5.9 for windows when it’s supposed to be not used?

selsta:
> Why do you care about 5.9 for windows when it’s supposed to be not used?

We still support 5.9 and also use it for release binaries, that’s why I asked.

stoffu:
> We still support 5.9 and also use it for release binaries, that’s why I asked.

The Windows build instruction doesn't specify which version of Qt to use, and the user is instructed to keep the MINGW system updated via `pacman -Syuu`. I assume that the release binaries are built the same way as in the instruction, using the latest 5.12. Thus I don't think your question applies here, and I wonder when you said that release binaries are made with 5.9; shouldn't 5.12 be used if the build instructions were to be followed?

selsta:
The build instructions have to get updated then.

stoffu:
Why, and in what ways?

selsta:
> Why? And in what ways should the build instruction be updated?

See #2396.

> I assume that the release binaries are built the same way as in the instruction, using the latest 5.12. 

No, the release binaries use Qt 5.9.7. The README instructions show the easiest way to get the GUI compiled.

stoffu:
> No, the release binaries use Qt 5.9.7. The README instructions show the easiest way to get the GUI compiled, not necessarily the steps we use for release binaries.

Oh, this is a news to me, and I wonder why. Care to elaborate?

Exactly what’s the way to build the release binary for Windows using 5.9.7? Still using MINGW, and using the same deploy script?

I’d like someone familiar with Windows build to weigh in.

stoffu:
@xiphon 
Thanks for clarifying.

PR #2398, opened 2019-09-27
DaemonManager: implement async sendCommand

xiphon:

PR #2399, opened 2019-09-27
SimpleMode: implement "switch to another public node" button

xiphon:
Requires https://github.com/monero-project/monero-gui/pull/2398 https://github.com/monero-project/monero-gui/pull/2400

switch-public-node

### DIFF
--- a/components/NetworkStatusItem.qml
+++ b/components/NetworkStatusItem.qml
@@ -29,6 +29,7 @@
 import QtQuick 2.9
 import QtQuick.Layouts 1.1
 
+import FontAwesome 1.0
 import moneroComponents.Wallet 1.0
 import "../components" as MoneroComponents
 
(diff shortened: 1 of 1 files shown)


components/NetworkStatusItem.qml
```diff
@@ -146,6 +147,46 @@ Rectangle {
...
+                anchors.left: statusTextVal.right
+                anchors.leftMargin: 16
+                anchors.verticalCenter: parent.verticalCenter
+                color: refreshMouseArea.containsMouse ?  MoneroComponents.Style.dimmedFontColor : MoneroComponents.Style.defaultFontColor
+                font.family: FontAwesome.fontFamilySolid
+                font.pixelSize: 24
+                font.styleName: "Solid"
+                opacity: iconItem.opacity * (refreshMouseArea.visible ? 1 : 0.5)
+                text: FontAwesome.random
+                visible: item.connected && appWindow.walletMode < 2
```
selsta: What about showing this button when booststrap-daemon is set to auto? This would also be useful for advanced mode users.

xiphon: Yep

components/NetworkStatusItem.qml
```diff
@@ -146,6 +147,46 @@ Rectangle {
...
+                anchors.left: statusTextVal.right
+                anchors.leftMargin: 16
+                anchors.verticalCenter: parent.verticalCenter
+                color: refreshMouseArea.containsMouse ?  MoneroComponents.Style.dimmedFontColor : MoneroComponents.Style.defaultFontColor
+                font.family: FontAwesome.fontFamilySolid
+                font.pixelSize: 24
+                font.styleName: "Solid"
+                opacity: iconItem.opacity * (refreshMouseArea.visible ? 1 : 0.5)
+                text: FontAwesome.random
+                visible: item.connected && !persistentSettings.useRemoteNode && persistentSettings.bootstrapNodeAddress == "auto"
```
selsta: Can you explain why `item.connected` ?

xiphon: Yep, should be `item.connected != ConnectionStatus_Disconnected`, fixed

rating89us:
What about a tooltip to inform the function of this button?

xiphon:
> What about a tooltip to inform the function of this button?

Yep, it obviously will have one, but have to implement tooltips in GUI first.

PR #2401, opened 2019-09-28
Transfer: warn on long Payment ID filled into description field

xiphon:


xmrdsc:
Maybe an idea to make `walletManager.paymentIdIsLong()` since we already have `walletManager.paymentIdValid()`.

xiphon:
I plan to drop the warning in the next Monero GUI version and don't want to introduce `paymentIdIsLong()` that won't be ever used somewhere else.

PR #2402, opened 2019-09-28
Transfer: forbid long Payment IDs

xiphon:


pages/Transfer.qml
```diff
@@ -394,8 +394,10 @@ Rectangle {
       MoneroComponents.WarningBox {
           // @TODO: remove after pid removal hardfork
           id: paymentIdWarningBox
-          text: qsTr("You can enable transfers with payment ID on the settings page.") + translationManager.emptyString;
-          visible: !persistentSettings.showPid && (warningLongPidTransfer || warningLongPidDescription)
+          text: qsTr("Long payment IDs are obsolete. \
+          Long payment IDs were not encrypted on the blockchain and would harm your privacy. \
```
xmrdsc: *could*

xiphon: The message comes from https://github.com/monero-project/monero/pull/5834

pages/Transfer.qml
```diff
@@ -360,7 +359,7 @@ Rectangle {
           }
 
           ColumnLayout {
-              visible: appWindow.persistentSettings.showPid || paymentIdCheckbox.checked
+              visible: paymentIdCheckbox.checked
```
selsta: How is this supposed to be visible?

xiphon: To support Monero Payment URI decoding

selsta: Ah makes sense

PR #2404, opened 2019-09-29
Cmake

TheCharlatan:
This is a work in progress pull request. Any help to get it merge ready is appreciated. I do not have much experience with the gui development workflows, so any suggestions for tooling and workflow changes will also be much appreciated.

The current linking is done for latest monero master, so currently please test this with: `make devmode` and be sure to fetch the latest tip of the submodule.

Currently this only works on Linux and I have not tested building the translations yet. Mac and Windows support will require some smaller patches, like including monero's EXTRA_LIBRARIES and adding some search paths, but it should not be complicated to get it working. If you'd like to add your changes, please either pass me your git diff, or open a pull request against this branch on my repository.

ZBAR and ZXCVBN support are also not fully tested yet, since I had some trouble finding and linking the correct libraries.

The content in the big commit is not split in order to preserve working compilation across commits. Once its content is added to master, the old build script will no longer work and all existing build toolings will require changes.

Monero's cmake directory's files need to be copied to this project's cmake directory in order for the linking and function definitions to work correctly. Monero-gui gets its own version check and generate file in order to not conflict with monero's destination version files.

Most of the source files that are currently in monero-gui's root directory are now moved to subdirectories. This is done to most importantly preserve compilation order, but also to give some content structure.

The current build script always checks out latest master of the monero submodule. The submodule rules in the CMakeLists.txt file do not enforce latest master, but rather the submodule's latest committed hash. An override to compile master nevertheless can still be given
with `-D DEV_MODE` as an argument to cmake.

To enable the linux X11 xcb linking the libraries had to be hardcoded. There does not seem to be good support for this in pkgconfig, or in existing cmake checks.


TheCharlatan:
Added support for windows and mac, but I need some input from the people that build releases and those that just test the software, before I can continue.

selsta:
I’m really not familiar with cmake but I did `cmake .` and now I’m getting

> Could NOT find X11 (missing: X11_X11_INCLUDE_PATH X11_X11_LIB)

Before that it said:

> git submodule update --init --force

and then

> git submodule update --init --force external/miniupnp

Maybe it’s better to recommend `git submodule update --init --force --recursive` in the first place?

xiphon:
Finally successfully built static Windows binaries, sent a PR with MinGW-related changes.

xmrdsc:
I believe [modern CMake](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/) follows the "target" idiom, example:

Instead of `include_directories()` use `target_include_directories()`

Issue #2410, opened 2019-10-08
Extend warning when using a view key that the balance might not be correct


peli-pro:
If you are using a view key outgoing transactions are not reflected in the balance. This is counterintuitive for the not-so-experienced user and different from what people are used to from other cryptocurrencies. It leads to quite some questions in forums on why monero shows a wrong balance. But we have no warnings in place in the gui. This will lead to confusion for many users and it has been raised here too: https://github.com/monero-project/monero-gui/issues/1260. It can be corrected by importing keyimages.
We already have a warning in place in the gui when using a view wallet in file pages/Transfer.qml in line 751:
```
if(appWindow.viewOnly){
            root.sendButtonWarning = qsTr("Wallet is view-only and sends are not possible.") + translationManager.emptyString;
            return false;
        }
```
We might just extend this warning a bit to make clear that the balance might not be correct. Maybe something like:
```
Wallet is view-only and sends are not possible. 
Balance reflects only incoming but not outgoing transactions. 
The balance shown might be wrong.
```
or
```
Wallet is view-only and sends are not possible. 
Balance reflects only incoming but not outgoing transactions. 
The balance shown might be wrong unless you have imported key images.
```

peli-pro:
![m](https://user-images.githubusercontent.com/55572025/66415110-0e732100-e9fb-11e9-8ca1-04a3af960af0.png)

rating89us:
I prefer 2nd option + adding a "Read more..." link to a getmonero.org guide explaining further this issue and how to import key images

SamsungGalaxyPlayer:
I would rather not include a link unless absolutely necessary. I propose the following:

```
Wallet is view-only and sends are not possible. 
Unless key images are imported, the balance reflects only incoming but not outgoing transactions.
```

We can perhaps add some help text on key images. But this is already an advanced feature.

peli-pro:
I think we do not need a link. If the user is aware that the balance might be incorrect and that the solution has something to do with key images - that should be enough as a warning. 
Maybe we could add that the balance reflects incoming and change transactions.
```
Wallet is view-only and sends are not possible. 
Unless key images are imported, the balance reflects only incoming and change but not outgoing transactions.
```

SamsungGalaxyPlayer:
@peli-pro "incoming" includes change, so it may be redundant.

peli-pro:
[...]

PR #2417, opened 2019-10-17
View only wallet -> warning that balance reflects only incoming transactions

peli-pro:
See discussion here:   #2410 

For view-only wallets change warning from:
`Wallet is view-only and sends are not possible.`
to:
`Wallet is view-only and sends are not possible. Unless key images are imported,
the balance reflects only incoming but not outgoing  transactions.`  

Updated Warning

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -756,7 +756,8 @@ Rectangle {
 
         // Currently opened wallet is not view-only
         if(appWindow.viewOnly){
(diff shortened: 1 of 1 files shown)


erciccione:
Be aware that if included in 0.15, this string will result untranslated for all languages, which can be disappointing for a translator who spent time and efforts to see the language they worked on 100% complete (beside being suboptimal UX in general).

It's better to not merge string changes right before a release and leave only bug fixes during the code freeze.

PR #2423, opened 2019-11-01
Merchant: fix leaving the page

xiphon:


xiphon:
> left panel should be hidden in merchant mode.

Thanks, updated.

Issue #2405, opened 2019-10-01
Send page


Realchacal:
Little cleanup going for simplicity and intuitiveness ([which may have been lacking](https://www.reddit.com/r/Monero/comments/dbnid3/help_with_sending_monero_from_gui_wallet_and_fees/)):

![image](https://user-images.githubusercontent.com/46682965/65975039-4d8afa80-e46e-11e9-9f9f-ca217ae5ac74.png)

- The drop down menu in the address field shows addresses from the address book
- "Advanced options" combines "Add description" and "Advanced options".
- A right click in the address or amount field brings up the context menu "paste". That's where people are looking for this feature. 
![image](https://user-images.githubusercontent.com/46682965/65979715-0e60a780-e476-11e9-8ba4-53f770d96166.png)

selsta:
@Realchacal what do you think of https://github.com/monero-project/monero-gui/issues/740#issuecomment-492886836

Your design doesn’t support sending to multiple recipients.

Realchacal:
> Your design doesn’t support sending to multiple recipients.

Thx man! I had no clue this is a thing.

> what do you think of [#740 (comment)](https://github.com/monero-project/monero-gui/issues/740#issuecomment-492886836)?

That's some excellent ideas! I copied all I like (see updated proposal) but couldn't justify adding any complexity to the default page for that feature.

Realchacal:
Updated

selsta:
What do you think about hiding fee settings under advanced? The average user should not change the fee settings.

Realchacal:
That's a tough one, took me some time to think about it.

- On the one hand, it simplifies the `Send` page that a noob sees, which is extremely nice!
- On the other hand, the concept of fees is an integral part of transactions known from almost **all** other wallets.

Given that IMO one day every better wallet will have something like this
![image](https://user-images.githubusercontent.com/46682965/66250969-4a338f80-e74a-11e9-81ca-1867c9621e3d.png)
(since everything else is basically useless from a user perspective) I would not move the fee settings to `Advanced` but I'm not opposing it either.

rating89us:
User should be able to enter the amount of transaction in XMR and USD.

Realchacal:
> * Address field could have a placeholder _Send to Monero address..._, like Exodus wallet has (see below), since some users could inadvertently enter a Bitcoin address.

Not a problem since the address is checked anyway. But a disadvantage, since it adds visual complexity.

[...]

PR #2429, opened 2019-11-08
ContextMenu: implement 'paste' context menu for all text fields

### LINKED ISSUE #2359: The "Receive" tab format gives an address that doesn't fit the "Spend" tab format
image

The button here gives an output that looks like this: `monero:49dpgSbHdA6auSv2gMPEsfj3uXVqqon4LEGWpuhdfXJsULVLKjFXkFQhpywAU8WddGbsN2ZDDpiVb8byZH6FfgL2R7ZwKZE`

If you paste this in the send tab, it looks like this:

image

And it's not a valid address.

xiphon:
Fixes #2359
Partially addresses #2405

context-menu-paste

### DIFF
--- a/components/ContextMenu.qml
+++ b/components/ContextMenu.qml
@@ -1,0 +1,43 @@
+import QtQuick.Controls 2.2
+import QtQuick 2.9
+
+import "../components" as MoneroComponents
+
+MouseArea {
+    signal paste()
+
+    id: root
+    acceptedButtons: Qt.RightButton
+    anchors.fill: parent
+    onClicked: {
+        if (mouse.button === Qt.RightButton)
+            contextMenu.open()
+    }
+
+    Menu {
+        id: contextMenu
+
+        background: Rectangle {
+            radius: 2
+            color: MoneroComponents.Style.buttonInlineBackgroundColor
+        }
+
+        font.family: MoneroComponents.Style.fontRegular.name
+        font.pixelSize: 14
+        width: 50
+        x: root.mouseX
+        y: root.mouseY
+
+        MenuItem {
+            id: pasteItem
+            background: Rectangle {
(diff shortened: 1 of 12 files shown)


selsta:
Can you disable the paste button on inputs that are read only? e.g. adding

```
diff --git a/components/InputMulti.qml b/components/InputMulti.qml
index 8fccb6a..c64acfe 100644
--- a/components/InputMulti.qml
+++ b/components/InputMulti.qml
@@ -70,6 +70,7 @@ TextArea {
     }
 
     MoneroComponents.ContextMenu {
+        visible: !parent.readOnly
         onPaste: {
             textArea.clear();
             textArea.paste();
```

to every input. Better solution would be to have a disable paste property as other context menu items could still exist.

xiphon:
> e.g. adding ... to every input. 

No

> Better solution would be to have a disable paste property

No

> Can you disable the paste button on inputs that are read only?

Yes. Updated

xiphon:
Updated. Now you can paste a seed during wallet restore.

PS: Will update remaining `TextArea`s later (in a separate PR). There's no hurry cause all of them are readOnly, so you can't use `paste` anyway.

Realchacal:
NICE!!!!!!!!!

Any chance to get...

- rounded edges?
- light text color? (which doesn't make the button look disabled)
- shadow? (which makes it look like an element with depth that can be clicked)
- a larger button? (which reminds of a [classical context menu](https://ourcodeworld.com/public-media/articles/articleocw-5c9a7439d33aa.png))

image

xiphon:
> Any chance to get...
> 
>     * rounded edges?

the edges are already rounded

Realchacal:
> the edges are already rounded

Oh you're right. Didn't see.

image

PR #2438, opened 2019-11-12
installer: automatically set app version information

xiphon:
Requires https://github.com/monero-project/monero-gui/pull/2437


rbrunner7:
Nice! So at least for point releases things get pretty much automated. And it was time that the exe files have version info embedded.

PR #2464, opened 2019-11-23
AppWindow: append wallet name to window title

### LINKED ISSUE #2459: Add wallet name to window title
Wallet name is not included in window title. 

Current window title in Windows is always Monero. I suggest renaming to "**Monero - _Wallet name_**".

This would be useful when custom decorations are disabled and when multiple instances are running.
image

xiphon:
Closes #2459

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -47,7 +47,7 @@ import "js/Windows.js" as Windows
 
 ApplicationWindow {
     id: appWindow
-    title: "Monero"
+    title: "Monero" + (walletName ? " - " + walletName : "")
 
     property var currentItem
     property bool hideBalanceForced: false


selsta:
Does a zero length string evaluate as false in JS?

jonathancross:
This can lead to strange results if `walletName` ever ends up as `null` or `false`:
```
var walletName = null
"Monero" + (walletName && " - " + walletName)
"Moneronull"

var walletName = false
"Monero" + (walletName && " - " + walletName)
"Monerofalse"
```

Maybe better to be explicit?
```
"Monero" + (walletName ? " - " + walletName : "")
```

xiphon:
> This can lead to strange results if `walletName` ever ends up as `null` or `false`:

You can't set `walletName` to `null` or `false`.

xiphon:
Anyway, i do agree, better to use some typical syntax for this. Updated.

PR #2484, opened 2019-11-24
build: fix on macOS

selsta:


selsta:
We could use the same command to get the brew openssl dir as monero: https://github.com/monero-project/monero/blob/v0.15.0.1/CMakeLists.txt#L422

Not sure how to handle fallback if there’s no brew installed.

xiphon:
https://github.com/monero-project/monero-gui/pull/2494

Issue #2488, opened 2019-11-25
Send page: "Add description" icon should change from + to - when expanded


rating89us:
Since the Add description field can be expanded and collapsed, the + icon should change to - when it is expanded. 

Proposed design: 
![image](https://user-images.githubusercontent.com/45968869/69506538-5c060780-0f2f-11ea-896a-d7e1283c8266.png)

Related to #2487

PR #2489, opened 2019-11-25
CheckBox: use Font Awesome plus/minus icons, remove unused images

xiphon:
Closes https://github.com/monero-project/monero-gui/issues/2488


rating89us:
I'm not sure checked / unchecked should be used for pages other than the send page.

The + icon is only used as a collapse/expand function in the send page. In other pages it is used only as a common button.

pages/Account.qml
```diff
@@ -349,8 +349,9 @@ Rectangle {
                 id: addNewAccountCheckbox 
                 visible: !selectAndSend
                 border: false
-                checkedIcon: "qrc:///images/plus-in-circle-medium-white.png" 
-                uncheckedIcon: "qrc:///images/plus-in-circle-medium-white.png" 
+                checkedIcon: FontAwesome.minusCircle
```
selsta: 2019-11-25 06:31:38.540	W qrc:/pages/Account.qml:353: ReferenceError: FontAwesome is not defined

pages/Transfer.qml
```diff
@@ -370,8 +371,9 @@ Rectangle {
               CheckBox {
                   id: paymentIdCheckbox
                   border: false
-                  checkedIcon: "qrc:///images/plus-in-circle-medium-white.png"
-                  uncheckedIcon: "qrc:///images/plus-in-circle-medium-white.png"
+                    checkedIcon: FontAwesome.minusCircle
```
selsta: not indented

xiphon: Actually is the correct indentation. The lines above and below are not indented properly.

selsta: yep, half of the Transfer page is all wrong with indentation. If you prefer to use right indentation with new code also ok

xiphon: > If you prefer to use right indentation with new code also ok

Yeah, kinda tired of manually adjusting the spaces

xiphon:
> I'm not sure checked / unchecked should be used for the + icons in pages other than the send page.

Yep, fixed

rating89us:
Can we reduce the space between + icon and Text? see below
image

luigi1111:
@rating89us PR welcome or consider making a new issue regarding this. :)

PR #2499, opened 2019-11-26
wizard: remove wizard specific languages variables

### LINKED ISSUE #2478: Can't change language using main menu
Monero GUI v0.15.0.1

After opening Monero GUI for the second time, there are two ways to change the Language:
- Inside the wallet: globe icon in upper left side
- Main menu: using `Change language` button in Main menu or in "Change wallet mode" page 

The language is not changing in the second case, when the user is selecting a language for the second time in main menu, with the wallet closed. 

After choosing a new language, the main menu changes to the new language, but when the wallet is reopened, the old language is used.

Steps to reproduce:
1. Open a wallet file
2. Close wallet and go to Main Menu
3. Click "Change wallet mode" > "Change Language" > "Language"
4. Select a new language to the Monero GUI and click "Continue"
5. Select Wallet mode
6. Open a wallet file
7. Monero GUI will open the wallet in the old language

selsta:
Fixes #2478

The first time I ran this patch my persistent storage got wiped so it’s most likely not correct.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -212,12 +212,6 @@ ApplicationWindow {
         else
           walletManager.setLogLevel(persistentSettings.logLevel)
 
-        // setup language
-        var locale = persistentSettings.locale
-        if (locale !== "") {
-            translationManager.setLanguage(locale.split("_")[0]);
-        }
-
         // Reload transfer page with translations enabled
         middlePanel.transferView.onPageCompleted();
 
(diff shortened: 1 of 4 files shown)


main.qml
```diff
@@ -1363,8 +1363,9 @@ ApplicationWindow {
             return "";
         }
 
-        property string language
-        property string locale
+        property string language : 'English (US)'
```
xiphon: Maybe drop the space preceding the colon

xiphon: Ah, i see, there are other properties declared this way. But still think we usually don't have the spacing here.

Issue #2430, opened 2019-11-08
GUI should default to light theme


SomaticFanatic:
Monero has a tough balance to make. On the one hand it is a coin defined by the awesome cutting-edge privacy it offers people. On the other hand it already has associations with enabling crime and seedy/dodgy behavior. 

With this balance in mind, the GUI, Monero’s official wallet software, should use the light theme by default. There’s no nice way to say this: the dark theme makes the wallet look like it is used for the dark web. The black background and electric orange buttons feel almost exactly like the kind of wallet that is used to buy illicit goods online. I know the dark theme also feels more “private”, which is the identity of Monero, but in my view the technology behind Monero is sufficient for the user to feel like their transactions are private.

I say this as a user of the GUI with dark-mode enabled. It is beautifully done. Kudos to the devs who designed it. But if we want to make Monero feel more about privacy and less about the dark web then it should be an option to enable, not on by default.

xmrdsc:
Disagree, white theme hurts my eyes. :|

PR #2539, opened 2019-12-01
main.qml: Light theme by default

SomaticFanatic:
#2430

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1399,7 +1399,7 @@ ApplicationWindow {
         property bool lockOnUserInActivity: true
         property int walletMode: 2
         property int lockOnUserInActivityInterval: 10  // minutes
-        property bool blackTheme: true
+        property bool blackTheme: false
 
         property bool fiatPriceEnabled: false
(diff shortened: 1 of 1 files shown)


selsta:
I think this should wait until we improve the contrast on the white theme a bit, IMO it’s not perfect to read at the moment.

SomaticFanatic:
What part of the lighter theme do you feel needs a better contrast? The font?

erciccione:
> Edit: We should just merge this and work on the light themes design in a future PR

I disagree. Using the white theme as default was discussed (and rejected) at the time the black theme was introduced. If we want to switch back to using the white theme as default, the design should be optimized before, not after.

elibroftw:
Can we close this and the related issue? If anything the default theme should be the system's theme and then dark theme as it is more readable. Light theme by default only exists because white is the default background. If #121212 was the default background color in the HTML spec when the web first launched, then this discussion wouldn't even be a thing 😂. 

It's been 3 years now.

Issue #2535, opened 2019-11-30
Transactions page: missing translations


rating89us:
Some texts aren't being translated, even when translation is available

![image](https://user-images.githubusercontent.com/45968869/69899824-0df86600-136c-11ea-878e-0ff1a37ec0b4.png)

selsta:
All 3 mentioned issues here seem resolved.

PR #2540, opened 2019-12-01
Translation: fix untranslated strings

### LINKED ISSUE #2533: Merchant page: some parts not being translated
Some parts of the merchant page are not being translated, even when translation is available:

image

selsta:
Fixes #2533
Partly addresses #2535

### DIFF
--- a/pages/merchant/Merchant.qml
+++ b/pages/merchant/Merchant.qml
@@ -158,7 +158,7 @@ Item {
                                         "<p>It's up to you whether to accept unconfirmed transactions or not. It is likely they'll be " +
                                         "confirmed in short order, but there is still a possibility they might not, so for larger " +
                                         "values you may want to wait for one or more confirmation(s).</p>"
-                                    );
+                                    ) + translationManager.emptyString;
                             } else if(root.trackingError !== ""){
                                 return root.trackingError;
                             } else if(trackingModel.count < 1){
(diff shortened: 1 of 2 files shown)


pages/merchant/Merchant.qml
```diff
@@ -265,7 +265,7 @@ Item {
                     font.pixelSize: 12
                     font.bold: false
                     color: "white"
-                    text: "<style type='text/css'>a {text-decoration: none; color: #FF6C3C; font-size: 12px;}</style>Currently selected address: " + addressLabel + " <a href='#'>(Change)</a>"
+                    text: qsTr("<style type='text/css'>a {text-decoration: none; color: #FF6C3C; font-size: 12px;}</style>Currently selected address: ") + addressLabel + qsTr(" <a href='#'>(Change)</a>") + translationManager.emptyString
```
jonathancross: In general, we should avoid including code (HTML, css, JavaScript) in translated strings. It makes translations messy & code changes more difficult / error-prone.

We can fix later to avoid breaking existing translations.

xiphon: Already pointed to this issue long time ago. Though the decision was "let's keep it that way for now".

Yes, i agree and i'm sure we should strip all the **garbage** from translated strings.

jonathancross: Yeah, I've done a lot of cleanup like this in the past... One can massage the translations and qml such that existing translated strings still match up.

PR #2576, opened 2019-12-08
InputDialog: pre-populate input text

### LINKED ISSUE #2567: No pre-populated text when editing pre-existing description, account label or address label
Current text is not pre-populated when you are editing a pre-existing description (Transactions page), an account label (Account page) or an address label (Receive page).

descrip

selsta:
Fixes #2567

### DIFF
--- a/pages/Account.qml
+++ b/pages/Account.qml
@@ -53,12 +53,11 @@ Rectangle {
 
     function renameSubaddressAccountLabel(_index){
         inputDialog.labelText = qsTr("Set the label of the selected account:") + translationManager.emptyString;
-        inputDialog.inputText = appWindow.currentWallet.getSubaddressLabel(_index, 0);
         inputDialog.onAcceptedCallback = function() {
             appWindow.currentWallet.subaddressAccount.setLabel(_index, inputDialog.inputText)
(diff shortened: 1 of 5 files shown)


selsta:
Will update PR later, current patch is wrong.

selsta:
That’s what I had in the beginning, I used this variant because the diff is smaller. Both is fine for me.

xiphon:
> That’s what I had in the beginning, I used this variant because the diff is smaller. Both is fine for me.

Great, would be nice to use the optional argument.

selsta:
updated

pages/Account.qml
```diff
@@ -354,7 +353,6 @@ Rectangle {
                 text: qsTr("Create new account") + translationManager.emptyString; 
                 onClicked: { 
                     inputDialog.labelText = qsTr("Set the label of the new account:") + translationManager.emptyString
-                    inputDialog.inputText = qsTr("(Untitled)") + translationManager.emptyString
```
xiphon: Did you forget this one?

xiphon: Ah, seems it is intentional

PR #2578, opened 2019-12-09
components: handle enter and return

### LINKED ISSUE #2562: Text input dialog: Enter key on the numeric keyboard doesn't work
image

image

On Windows, nothing happens when you type your password/description and press Enter key on the numeric keyboard.

selsta:
Fixes #2562

### DIFF
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -254,6 +254,7 @@ Item {
                 }
 
                 Keys.enabled: root.visible
+                Keys.onEnterPressed: Keys.onReturnPressed(event)
                 Keys.onReturnPressed: {
                     root.close()
                     if (passwordDialogMode) {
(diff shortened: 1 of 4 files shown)


components/PasswordDialog.qml
```diff
@@ -352,7 +353,8 @@ Item {
                     }
                 }
 
-                Keys.onReturnPressed: {
+                onAccepted: {
+                    if (!root.visible) return
```
xiphon: Why do you need this?

selsta: Wanted to avoid something like #1512

xiphon: Would prefer not to do this

selsta: Everywhere or just here?

xiphon: Everywhere. Let me know if this will lead to a bug, would like to reproduce it.

selsta: Yes, it leads to a bug. You can try yourself: Open GUI, go to Transactions -> Payment proof -> Press enter.

Issue #2474, opened 2019-11-24
Settings > Info page: add wallet name and some links


rating89us:
![image](https://user-images.githubusercontent.com/45968869/69487908-bb4a1600-0e62-11ea-96c8-43480bd1ce1a.png)

The monero wallet file is strange because it doesn't have any extension. I propose the following changes to this page:

- Add a **Wallet file name:**.
- Rename **Wallet path** to **Wallet location**. In this case it would be .../wallets/test
- Make the wallet location text clickable, and open the wallet directory in the OS file explorer.
- Rename **Wallet log path** to **Wallet log file**.
- Make the wallet log file clickable. Open the .log file when clicked.
- Wallet mode: add a link "**(change)**", that sends the user to wallet mode selection page. Currently, in order to access this page, the user has to close the wallet and then click on the button "_change wallet mode_".

PR #2580, opened 2019-12-10
SettingsInfo: open containing folder on wallet log path click

xiphon:
Partially addresses https://github.com/monero-project/monero-gui/issues/2474

### DIFF
--- a/pages/settings/SettingsInfo.qml
+++ b/pages/settings/SettingsInfo.qml
@@ -132,14 +132,21 @@ Rectangle {
 
             MoneroComponents.TextBlock {
                 Layout.fillWidth: true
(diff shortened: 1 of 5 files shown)


selsta:
#2474 asks for opening the wallet folder for the wallet and the log file directly for the log. I would also prefer it this way. Code looks good otherwise.

xiphon:
> opening the wallet folder for the wallet 

yep, implemented this as well and updated the PR

> the log file directly for the log

We should open the folder, not the file. There is a plenty of other things (except viewing the log) a typical user would like to do: copy the file, remove it, wipe old logs, etc.

PR #2584, opened 2019-12-10
installer: change blockchain folder string

### LINKED ISSUE #2583: Start Menu: check blockchain folder link
After installing Monero GUI in Windows with installer, some Start Menu links are created. Since the blockchain folder can be changed in Monero GUI, the `Check Blockchain Folder` link should be renamed to `Check Default Blockchain Folder`. This is already the case for the `Check Default Wallet Folder` link.

image

selsta:
ping @rbrunner7, is this going to work when upgrading the next time? What will happen to the old shortcut?

Fixes #2583

### DIFF
--- a/installers/windows/Monero.iss
+++ b/installers/windows/Monero.iss
@@ -166,6 +166,7 @@ Type: files; Name: "{app}\libssp-0.dll"
 Type: files; Name: "{app}\libhidapi-0.dll"
 Type: files; Name: "{app}\libeay32.dll"
 Type: files; Name: "{app}\ssleay32.dll"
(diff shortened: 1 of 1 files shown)


rbrunner7:
I guess the update / replacement of that shortcut will work, but I am not sure. [This StackOverflow question and answer](https://stackoverflow.com/questions/46220922/inno-setup-rename-existing-shortcut) seem to suggest otherwise.

In such cases I tend to check with a test run of the installer to find out what's the exact result.

selsta:
@rbrunner7 Can you test / review this? :)

Issue #2585, opened 2019-12-11
bug creating and saving GUI wallet and key files (solved)


criptonoia:
hello,
as i was creating a monero wallet, the app "GUI wallet" failed to create the files and key for the wallet. it was my first time using monero and no error message popped up for me. after using the wallet and sending 1.69926200 monero, the app closed my wallet and there is no way to recover that. 
i still have the log, i will paste a part of it here where you can see the bug
"2019-12-11 16:54:44.892  16632  ERROR  wallet.wallet2  src/wallet/wallet2.cpp:3849  failed to generate wallet keys file C:/Users/x/Documents/Monero/wallets/monero /monero .keys.new
2019-12-11 16:54:44.893  16632  ERROR  wallet.wallet2  src/wallet/wallet2.cpp:5691  !r. THROW EXCEPTION: error::file_save_error
2019-12-11 16:54:44.893  16632  ERROR  WalletAPI  src/wallet/api/wallet.cpp:917  Error saving wallet: failed to save file "C:/Users/x/Documents/Monero/wallets/monero /monero .keys"
2019-12-11 16:54:44.909  16632  ERROR  wallet.wallet2  src/wallet/wallet2.cpp:5256  !boost::filesystem::exists(m_keys_file, ignored_ec). THROW EXCEPTION: error::file_not_found
2019-12-11 16:54:44.912  16632  INFO  logging  contrib/epee/src/mlog.cpp:273  New log categories: *:WARNING,net:FATAL,net.http:FATAL,net.ssl:FATAL,net.p2p:FATAL,net.cn:FATAL,global:INFO,verify:FATAL,serialization:FATAL,daemon.rpc.payment:ERROR,stacktrace:INFO,logging:INFO,msgwriter:INFO
2019-12-11 16:54:44.937  9260  INFO  global  contrib/epee/src/net_ssl.cpp:127  Generating SSL certificate
2019-12-11 16:54:44.944  2188  INFO  global  contrib/epee/src/net_ssl.cpp:127  Generating SSL certificate
2019-12-11 16:54:44.964  16632  ERROR  wallet.wallet2  src/wallet/wallet2.cpp:5256  !boost::filesystem::exists(m_keys_file, ignored_ec). THROW EXCEPTION: error::file_not_found
2019-12-11 16:54:52.634  10076  ERROR  WalletAPI  src/wallet/api/wallet.cpp:1031  daemonBlockChainHeight: Failed to connect to daemon
[...]

PR #2590, opened 2019-12-13
WizardController: always close and re-open newly created wallet

xiphon:
Fixes https://github.com/monero-project/monero-gui/issues/2585

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -57,7 +57,6 @@ ApplicationWindow {
     property var transaction;
     property var transactionDescription;
     property var walletPassword
-    property bool isNewWallet: false
     property int restoreHeight:0
     property bool daemonSynced: false
     property bool walletSynced: false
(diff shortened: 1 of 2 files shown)


selsta:
Clicking on `Cancel` on the `Open wallet from file` page now redirects back to wizard home. I preferred previous behaviour. Patch looks good otherwise.

xiphon:
Updated

selsta:
I would also suggest something like this so behaviour is like it was before this PR.
```
diff --git a/main.qml b/main.qml
index ad22942..6ceba15 100644
--- a/main.qml
+++ b/main.qml
@@ -202,7 +202,8 @@ ApplicationWindow {
 
     function openWallet() {
         const prevState = appWindow.viewState;
-        appWindow.viewState = "normal";
+        if (appWindow.viewState !== "wizard" || wizard.wizardState !== "wizardOpenWallet1")
+            appWindow.viewState = "normal";
         passwordDialog.onAcceptedCallback = function() {
             walletPassword = passwordDialog.password;
             initialize();
```

PR #2592, opened 2019-12-14
WizardOpenWallet1: fix warning

selsta:
```
2019-12-14 02:07:08.800	W qrc:/wizard/WizardOpenWallet1.qml:72: ReferenceError: WalletKeysFilesModel is not defined
2019-12-14 02:07:08.800	W qrc:/wizard/WizardOpenWallet1.qml:48: ReferenceError: WalletKeysFilesModel is not defined
```


xiphon:
Actually the warning should never appear, we are setting `walletKeysFilesModel` before we load any QML document. Not sure whether it is a Qt bug or we are doing it wrong.

selsta:
@xiphon It shows up when closing the GUI.

PR #2601, opened 2019-12-15
remove colored dots

### LINKED ISSUE #2573: Menu: colored dots aren't consistent with wallet's design
After removing the color stripe at the top (#2527), we should also remove colored dots in Menu.

Current:
image

After:
image

tobtoht:
closes #2573

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -364,7 +364,7 @@ Rectangle {
             MoneroComponents.MenuButtonDivider {
                 anchors.left: parent.left
                 anchors.right: parent.right
-                anchors.leftMargin: 16
+                anchors.leftMargin: 20
             }
 
             // ------------- Account tab ---------------
@@ -374,7 +374,6 @@ Rectangle {
                 anchors.right: parent.right
                 text: qsTr("Account") + translationManager.emptyString
                 symbol: qsTr("T") + translationManager.emptyString
-                dotColor: "#44AAFF"
                 onClicked: {
                     parent.previousButton.checked = false
                     parent.previousButton = accountButton
@@ -386,7 +385,7 @@ Rectangle {
                 visible: accountButton.present
                 anchors.left: parent.left
                 anchors.right: parent.right
-                anchors.leftMargin: 16
+                anchors.leftMargin: 20
             }
 
             // ------------- Transfer tab ---------------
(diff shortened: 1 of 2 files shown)


tobtoht:
Comparison:

gui_dots_before
gui_dots_after

IMO dots removed gives a cleaner look that is more in line with the rest of the UI.

selsta:
> IMO dots removed gives a cleaner look.

Agree, it does look slightly cleaner but I think it worsens the UX. The colored dots allow you to quickly see which menu you have selected and they can also help find the menu item you are looking for. Just text is IMO a bad menu.

What do you think about replacing the dots with minimalistic icons? See Github as an example:

Screenshot 2019-12-17 at 00 38 55

Also ping @rating89us

tobtoht:
There are currently 3 indicators for the selected item: the dot with the arrow, the highlighted '>' and the color gradient. I find that with the dots removed it is not any harder to see which item is selected.

I agree with your point on quickly being able to find a menu item, although I don't think the colored dots aid in that regard as it is now. I'm not opposed to replacing the dots with icons if they represent the tab in a meaningful way.

selsta:
If you want to update this PR for now can you change the left margin between the app border and menu text to be the same as the balance card? (Setting it to the half of the current value).

tobtoht:
Slightly more flush to the eye with 22 pixels instead of 20.

rating89us:
Maybe increase the gray in the black-gray gradient to better highlight the selected item

image

tobtoht:
gui_dots_line_1

vs.

gui_dots_line_2

Which one do you think is better?

rating89us:
I like the 2nd one.

And what if we also select Send when Address book is selected?
image

tobtoht:
gui_dots_removed_1

How's this?

rating89us:
Looks good to me

rating89us:
You forgot the orange bar

tobtoht:
It's the Rectangle in MenuButton.qml

rating89us:
Ok, sorry

selsta:
Orange bar is an improvement. Still curious what others think, I’m ok with it.

selsta:
dEBRUYNE prefers the dots but is ok with this change. What is your opinion @xiphon ?

PR #2604, opened 2019-12-15
do not ask to start daemon if in remote node mode

### LINKED ISSUE #2456: Do not automatically start a local node when connection to remote node failed.
Especially on Tails, as running a local node will not work without additional setup. 

User choice to use a remote node should be respected. The 10 second countdown is anxiety inducing. Instead, let the user know that the connection to the remote node failed.

tobtoht:
closes #2456

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -467,7 +467,7 @@ ApplicationWindow {
 
         // update local daemon status.
         const isDisconnected = status === Wallet.ConnectionStatus_Disconnected;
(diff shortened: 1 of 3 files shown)


xiphon:
Will be nice to completely replace all `isDaemonLocal` calls with `!persistentSettings.useRemoteNode`.
Would you like to do this and update the PR?

PS: It is okay if you don't have time to do it, just let me know

PR #2607, opened 2019-12-15
hide advanced options in tx menu when there are no txs

### LINKED ISSUE #2603: Transactions page: advanced options are visible when there aren't any transactions yet
image

tobtoht:
closes #2603

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -1288,6 +1288,7 @@ Rectangle {
         }
 
         MoneroComponents.CheckBox2 {
+            visible: root.txCount > 0
             id: showAdvancedCheckbox
             Layout.topMargin: 30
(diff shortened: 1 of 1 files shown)


xiphon:
Hiding the configuration option doesn't look good to me. 

Shouldn't hide any functionality from a user that just set up a wallet. A newbie observing GUI wallet for the first time should learn that such a functionality exists and he can use it in the future.

rating89us:
> Hiding the configuration option doesn't look good to me.

But all other configuration (sorting, search, etc.) are hidden when there are no transactions yet. We should either hide or show everything.

selsta:
I think you can also close this as we went with the other way. (always show options)

PR #2616, opened 2019-12-16
History: don't hide 'Export all history' if there ain't any txes

xiphon:
To be merged after #2614

Shouldn't hide any functionality from a user that just set up a wallet.

A newbie observing Monero GUI wallet for the first time should learn that such a functionality exists and could be used in the future.

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -1315,7 +1315,7 @@ Rectangle {
             }
 
             MoneroComponents.StandardButton {
-                visible: !isIOS && root.txCount > 0
+                visible: !isIOS
                 small: true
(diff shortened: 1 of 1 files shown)


xiphon:
> Exporting empty tx history results

to figure out the CSV format, i.e. available fields and their respective order, for example.

selsta:
> to figure out the CSV format, i.e. available fields and their respective order, for example.

RIght but that does not work right now.

xiphon:
> RIght but that does not work right now.

Once https://github.com/monero-project/monero-gui/pull/2614 gets merged, GUI will always export csv header.

PR #2622, opened 2019-12-16
fix clunky scrolling behavior on flickables

tobtoht:


selsta:
I guess `Flickable.DragAndOvershootBounds` is more of a macOS thing. I wouldn’t call it clunky (macOS has it everywhere) but I can understand that it probably is unusual on Linux/Windows. What do you prefer @xiphon?

tobtoht:
Comparison (Tor required):

http://xmrguide42y34onq.onion/static/overshoot.mp4
http://xmrguide42y34onq.onion/static/stop_at_bounds.mp4

rating89us:
I'm not familiar with MacOS, but if it's the default scrolling behavior, we should keep it in MacOS and remove only for Windows and Linux.

selsta:
Yes, most native macOS apps use this rubber band scrolling. We could do:

`boundsBehavior: isMac ? Flickable.DragAndOvershootBounds : Flickable.StopAtBounds`

PR #2625, opened 2019-12-17
fix dropdown closure behavior

tobtoht:
Clicking on parent will now properly close dropdown.


selsta:
Nice catch, please also add this here https://github.com/monero-project/monero-gui/blob/master/components/DatePicker.qml#L251

PR #2631, opened 2019-12-19
build: support DEV_MODE env var to override default value

xiphon:
|Target|Default DEV_MODE|
|-|-
|default|OFF|
|debug|ON|
|devmode|ON|
|scanner|ON|
|debug-static-win64|ON|
|debug-static-mac64|ON|
|release-static-win64|OFF|

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,23 +18,23 @@ endif
 
 
 default:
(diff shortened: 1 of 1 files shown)


xiphon:
@TheCharlatan

PR #2626, opened 2019-12-17
Improve transfer warning flow

tobtoht:
closes #2620

PR #2632, opened 2019-12-19
Transfer: cleanup payment id code

tobtoht:
Test with #2626

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -54,8 +54,6 @@ Rectangle {
     property string sendButtonWarning: ""
     property string startLinkText: qsTr("<style type='text/css'>a {text-decoration: none; color: #FF6C3C; font-size: 14px;}</style><font size='2'> (</font><a href='#'>Start daemon</a><font size='2'>)</font>") + translationManager.emptyString
     property bool showAdvanced: false
-    // @TODO: remove after pid removal hardfork
-    property bool warningLongPidTransfer: false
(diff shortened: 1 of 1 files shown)


selsta:
The long PID code is still used for when pasting e.g.

```
monero:46BeWrHpwXmHDpDEUmZBWZfoQpdc6HaERCNmx1pEYL2rAcuwufPN9rXHHtyUA4QVy66qeFQkn6sfK8aHYjA3jk3o1Bv16em?tx_amount=239.39014&tx_description=donation&tx_payment_id=0f33ad024c7788e623662a4b1461699225848ea90948a783a37e261f2d55c087
```

tobtoht:
Thanks, will fix.

tobtoht:
Hmm, what should the expected behavior be here, keep as is?

In that case the "remove after pid removal hardfork" comments should be removed.

selsta:
> Hmm, what should the expected behavior be here, keep as is?

yes

> In that case the "remove after pid removal hardfork" comments should be removed.

yea they probably should get removed

pages/Transfer.qml
```diff
@@ -403,17 +387,17 @@ Rectangle {
...
+                  error: paymentIdCheckbox.checked
               }
           }
       }
 
       MoneroComponents.WarningBox {
-          // @TODO: remove after pid removal hardfork
           id: paymentIdWarningBox
-          text: qsTr("Long payment IDs are obsolete. \
+          text: qsTr("Long payment IDs are no longer supported. \
```
selsta: Please change it back to obsolete, we lose the translation otherwise.

PR #2634, opened 2019-12-19
Merchant: add option to enable in settings

tobtoht:
Currently it will only show up in advanced mode.

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -461,7 +461,7 @@ Rectangle {
 
             MoneroComponents.MenuButton {
                 id: merchantButton
-                visible: appWindow.walletMode >= 2
+                visible: persistentSettings.enableMerchantPage
                 anchors.left: parent.left
                 anchors.right: parent.right
                 text: qsTr("Merchant") + translationManager.emptyString
@@ -476,7 +476,7 @@ Rectangle {
             }
 
             MoneroComponents.MenuButtonDivider {
-                visible: merchantButton.present && appWindow.walletMode >= 2
+                visible: merchantButton.present && persistentSettings.enableMerchantPage
                 anchors.left: parent.left
                 anchors.right: parent.right
                 anchors.leftMargin: 16
--- a/main.qml
+++ b/main.qml
@@ -1397,7 +1397,7 @@ ApplicationWindow {
         property int walletMode: 2
         property int lockOnUserInActivityInterval: 10  // minutes
         property bool blackTheme: true
-
+        property bool enableMerchantPage: false
         property bool fiatPriceEnabled: false
         property bool fiatPriceToggle: false
         property string fiatPriceProvider: "kraken"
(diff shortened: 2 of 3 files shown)


selsta:
Don’t think we need a settings entry for it.

tobtoht:
I dislike having it listed by default. It is not relevant to most users and accidentally clicking on it takes one out of the UI flow to a page that is not immediately apparent how to exit. With a setting it is easily discoverable by anyone that wants to use it.

selsta:
It’s a Receive page submenu so it’s rather hidden anyway. If someone accidentally clicks on it then they can easily exit it. I’m really against a settings menu for it.

Ok with making it available in simple mode.

rating89us:
I agree with tobtoht:
* Most users don't use Merchant page
* Its design is completely different from the wallet.
* Its not something fundamental in a wallet for 99% of users. It could easily be a different software.
* If adding to the Settings page, it should be disable by default.

selsta:
> If added to the Settings page, it should be disabled by default in simple mode and enabled in advanced mode.

If we add a settings entry, this should be the default values. But I still think it adds unnecessary complexity. How many users are thinking “I would like to disable this submenu item.” ?

tobtoht:
From a user's perspective (with the merchant page enabled by default on all modes):

* The merchant page is completely irrelevant.
* People will click it on it if they're trying to find a particular menu but can't remember where it was. (I do this all the time)
* People will click on it out of curiosity.
* People will click on it because they misclick.
* If you see the page for the first time you have to parse the page before you know where the exit button is. -> "Oh it's this page again, how do I exit it again?" "Why would they put this here?"
* If your window is too small you can't exit it. (this should be fixed regardless)
* Taking a user out of the expected user interface flow is bad UX. 
* In general, do not force users to be cognizant of a feature that is irrelevant to them.

Considering this, I'm opposed to listing it by default in simple mode.

From a merchant's perspective looking to accept Monero (with the merchant page listed as an option in the settings):

* They will discover it by themselves through the settings because they learned the GUI supports such a feature
* They will ask for a way to accept Monero payments and be easily pointed to the setting.

Think about the [support query](https://old.reddit.com/r/Monero/comments/ecrlc2/merchant_page/fbdahj9/) I answered yesterday. It would have been much easier if I could have pointed to the setting instead of having to explain how to switch wallet modes.

If listed as an option in the settings, perhaps it makes more sense to list it as a top level menu entry instead of as a submenu item of the receive page.

selsta:
What do you think about adding the open merchant page button to the wallet settings page here? #2655

I don‘t think that the current submenu item is a good place for it. But leaving it there but adding an additional settings entry makes it worse IMO.

tobtoht:
@selsta I think that is a decent solution.

image

xiphon:
Yeah, support removing Merchant submenu item.

Just maybe use "Merchant mode" instead of "Open the merchant page" or something like that? Opinions?

tobtoht:
To be consistent with the other items there should be a verb at the beginning. 

What do you think of: "Enter merchant mode", or "Open merchant page"

xiphon:
In my understanding Merchant functionality is more like a mode, not just a page.

tobtoht:
I agree, changed.

image

selsta:
Looks good.

PR #2639, opened 2019-12-19
Save wallet after generating accounts and subaddress and after changing labels

tobtoht:
Losing subaddresses and descriptions due to wallet crashes is bad UX

### DIFF
--- a/pages/Account.qml
+++ b/pages/Account.qml
@@ -56,6 +56,7 @@ Rectangle {
         inputDialog.inputText = appWindow.currentWallet.getSubaddressLabel(_index, 0);
         inputDialog.onAcceptedCallback = function() {
             appWindow.currentWallet.subaddressAccount.setLabel(_index, inputDialog.inputText)
+            appWindow.currentWallet.store()
         }
         inputDialog.onRejectedCallback = null;
         inputDialog.open()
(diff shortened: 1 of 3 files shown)


rating89us:
Are descriptions of sent transactions also lost in case of a wallet crash?

tobtoht:
@rating89us No, it saves the wallet after sending a TX. If the description is updated afterwards it will be lost in case of a crash.

rating89us:
And account labels (names)?

tobtoht:
Currently the only instances where the wallet is saved are:
* Clean exit of the GUI
* After sending a TX
* After the first initial sync

Created accounts and descriptions will be lost on a wallet crash. I will update the PR to include saving after account creation. I can also have it update the wallet after every description update, but that may increase the risk of corrupting the wallet files if the wallet crashes during a save, I'm not sure.

tobtoht:
Never mind, better to have the behavior be consistent.

tobtoht:
Saving the wallet on my machine appears to take a fraction of a second. I closed it forcefully multiple times immediately after adding a new subaddress and so far the wallet has successfully saved the new subaddress.

selsta:
I wonder if this could be done async.

xiphon:
I'm against such a change. Should fix wallet crash instead (i'm working on it).

tobtoht:
This isn't just relevant for wallet crashes. Also Linux users that ^C their terminal, system freezes (common on live operating systems such as Tails), power failure, etc. I have personally answered many support queries regarding disappearing subaddresses, and while it never results in a loss of funds it does contribute to the shroud of unreliability surrounding the GUI. Moreover, losing subaddresses can be detrimental to privacy if they are relied upon to guard against off-chain linking. At the minimum it should be ensured that subaddresses and accounts are saved upon creation.

PR #2640, opened 2019-12-20
main: fix wallet lock is unlocked without password

### LINKED ISSUE #2563: Wallet lock is unlocked without password
This happens when the wallet is automatically locked for inactivity while the `Set description` dialog is open in Transactions page. 

image

When the `Please enter wallet password` dialog appears, both dialogs get mixed, and the `Set description` dialog can be canceled, which removes the black screen in the layer below thus bypassing the wallet lock without entering a password. 

image

image

lock2

This allows the user to see all wallet's transactions, accounts and balances. Funds can't be sent because the password is asked again.

selsta:
Closes #2563

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -2044,6 +2044,7 @@ ApplicationWindow {
         if(rootItem.state !== "normal") return;
         if(!persistentSettings.lockOnUserInActivity) return;
         if(passwordDialog.visible) return;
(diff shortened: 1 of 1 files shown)


xiphon:
What do you think about replacing this one with https://github.com/monero-project/monero-gui/pull/2681?

PR #2670, opened 2019-12-21
[Qt 5.14] main: enable high DPI scaling

selsta:
Tested on Ubuntu, someone has to test it on Windows. Requires Qt 5.14.


selsta:
@xiphon I guess we can merge this as Qt 5.13 or lower will simply ignore the env variable?

PR #2674, opened 2019-12-21
Scrollbar: don't autohide if enabled (except on Mac)

### LINKED ISSUE #2662: Do not autohide enabled scrollbar

tobtoht:
closes #2662

Unfortunately there seems to be no better way to do this.
`ScrollBar.AsNeeded` will autohide and `ScrollBar.AlwaysOn` will always display the scrollbar, even when not needed.

### DIFF
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -196,6 +196,7 @@ Rectangle {
                 anchors.topMargin: persistentSettings.customDecorations ? 60 : 10
                 anchors.bottom: parent.bottom
                 anchors.bottomMargin: persistentSettings.customDecorations ? 15 : 10
+                onActiveChanged: if (!active && !isMac) active = true
             }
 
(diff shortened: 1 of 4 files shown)


selsta:
Is this default behaviour on other platforms?

tobtoht:
@selta Yes, on Linux and Windows it's either always display or show when needed, most applications use the latter.

MiddlePanel.qml
```diff
@@ -196,6 +196,11 @@ Rectangle {
                 anchors.topMargin: persistentSettings.customDecorations ? 60 : 10
                 anchors.bottom: parent.bottom
                 anchors.bottomMargin: persistentSettings.customDecorations ? 15 : 10
+                onActiveChanged: {
```
selsta:
```suggestion
                onActiveChanged: if (!active && !isMac) active = true
```

PR #2681, opened 2019-12-23
InputDialog: visible override, hide when passwordDIalog is visible

xiphon:
Closes https://github.com/monero-project/monero-gui/issues/2563


selsta:
Does not work

Screenshot 2019-12-23 at 15 16 01

xiphon:
Ah, well, doesn't seem there is a clear fix, the whole dialogs design is broken

selsta:
> whole dialogs design is broken

yep

PR #2702, opened 2020-01-07
Rename "Open wallet" button to "Create wallet"

### LINKED ISSUE #2546: New wallet: "Open wallet" button is imprecise
In this page, the wallet is not created yet (if you click `Previous` button, and go back to Main menu, you won't see the wallet file). Therefore, instead of `Open wallet` the button should be `Create wallet` (it first creates and then it opens the new wallet).
image

rating89us:
Closes #2546


selsta:
See my comment in #2704

PR #2704, opened 2020-01-07
Transactions: add filter options in search field placeholder

### LINKED ISSUE #2628: Transactions page: search field placeholder doesn't inform what can be searched
Search field placeholder is `Search...`

But what can be searched in this this search field?

I suggest changing it to the search items that are supported. Example:
`Search by Transaction ID, Address, Description, Amount or Blockheight`

image

rating89us:
Closes #2628


selsta:
Thanks for contributing.

The `translations/monero-core.ts` gets changed automatically, it should be not changed manually. It would be nice if you can get familiar with the desktop client so that you can update PRs after opening them :) The GitHub web client is really basic and doesn’t allow this.

selsta:
The problem is now there are 3 commits for a single PR. We try to have a clean git history so they have to get squashed to one. See here: https://stackoverflow.com/a/5201642

Also please use a commit title prefix, e.g.: `History: add filter options in search field`.

PR #2706, opened 2020-01-08
Updated Card Background Images

TiagoDS7:
card-background-black
card-background-black@2x
card-background-white
card-background-white@2x


selsta:
+invalid

PR #2709, opened 2020-01-08
main: add blur to wizard password dialog

selsta:
@xiphon Was this left out intentionally?


xiphon:
Yep, there is no secret info to hide

selsta:
It looks weird now when opening a wallet from the wizard. This PR makes it more consistent.

selsta:
Before:
Screenshot 2020-01-08 at 22 11 29
After:
Screenshot 2020-01-08 at 22 09 14

PR #2710, opened 2020-01-09
Settings>Interface: add sections, rename labels, add language

### LINKED ISSUE #2651: Settings > Interface: redesign
* Create two sections: Security and Appearance
* Add icons to "Use light theme" and "Change language" (the same used in titlebar)
* When using light theme, change text to "Use dark theme"
* Change text to "Display balance in other currencies"

image

Current:
image

rating89us:
Closes #2651

* Add sections Security and Appearance
* Change "After * minutes" to "after * minutes", move text to the same line as "Lock wallet on inactivity"
* Slider is now always visible
* Add item Language
* Change language button now is always visible
* Add "Use" verb in "Light theme" item
* Add "Use" verb in "Custom decorations" item
* Change text to "Display balance in other currency"


selsta:
Nice. Here is some feedback:

- `Security` and `Appearance` don’t have the same bottom margin.
- The language text and button feel a bit out of place. Maybe this can be made visually nicer.
- Please use spaces for indentation. It appears that in some places you mix tabs and spaces. Also check if the indentation is correct everywhere.

PR #2718, opened 2020-01-15
Keys: add primary address to seed & keys page

### LINKED ISSUE #2434: No Primary address in "Seed and Keys" page
Normally when you want to know the seed and keys, you also want to know the main address, since you need the main address + private view-key + private spend-key in order to restore a wallet.

Therefore it would be nice to have all important information about the wallet in a single page.

rating89us:
Closes #2434

image

### DIFF
--- a/pages/Keys.qml
+++ b/pages/Keys.qml
@@ -131,7 +131,7 @@ Rectangle {
                 Layout.fillWidth: true
                 fontSize: 22
                 Layout.topMargin: 10
-                text: qsTr("Keys") + translationManager.emptyString
+                text: qsTr("Primary address & Keys") + translationManager.emptyString
             }
             Rectangle {
                 Layout.fillWidth: true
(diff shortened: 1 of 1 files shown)


xiphon:
Maybe place it right before the Keys section?

pages/Keys.qml
```diff
@@ -140,6 +140,15 @@ Rectangle {
...
+                Layout.fillWidth: true
+                id: primaryAddress
+                readOnly: true
+                copyButton: true
+                wrapMode: Text.Wrap
+                labelText: qsTr("Primary address") + translationManager.emptyString
+                fontSize: 16
+            }           
             MoneroComponents.LineEdit {
                 Layout.fillWidth: true
```
selsta:
```suggestion
                Layout.fillWidth: true
                Layout.topMargin: 25
```

pages/Keys.qml
```diff
@@ -300,4 +311,3 @@ Rectangle {
 
 
 
-
```
selsta: Please try to avoid unrelated whitespace changes, looks good otherwise.

PR #2731, opened 2020-01-17
TitleBar: add lock wallet button

### LINKED ISSUE #2005: Titlebar: create Lock wallet button
Currently you can't actively lock your wallet (put it in standby, waiting for a password). It is only done passively, when the wallet is inactive for a period of time.

A "Lock wallet" button should be implemented in title bar, together with "Close wallet" button.
When unlocked, icon should be:  image When locked, icon should be: image

image

See also issues:
#2656 (Add "Lock wallet" function in Settings > Wallet)
#2605 (Separate interface and wallet icons in titlebar)

### LINKED ISSUE #2658: Title bar: close wallet button should be visible when wallet is locked
Close wallet button (image) should be visible when wallet is locked and password input is open:

image

rating89us:
Closes #2005 and closes #2658

lockbutton2

### DIFF
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -73,7 +73,7 @@ Item {
         leftPanel.enabled = false
         middlePanel.enabled = false
         wizard.enabled = false
-        titleBar.state = "essentials"
+        titleBar.state = "default"
         root.visible = true;
         appWindow.hideBalanceForced = true;
         appWindow.updateBalance();
--- a/components/TitleBar.qml
+++ b/components/TitleBar.qml
@@ -62,12 +62,14 @@ Rectangle {
     states: [
         State {
             name: "default";
-            PropertyChanges { target: btnCloseWallet; visible: true}
-            PropertyChanges { target: btnLanguageToggle; visible: true}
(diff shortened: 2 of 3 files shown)


selsta:
Unsure on this one. I like the functionality but the titlebar is getting crowded, also 4 on the left and 3 on the right is uneven :P

@xiphon your opinion?

xiphon:
Like the functionality, don't like the particular patch

GBKS:
A simple solution to avoid clutter would be to move some of the icons into a dropdown (hamburger or three dots to indicate options). That way we can also have labels next to them, making them easier to understand.

rating89us:
I think we should remove the change language button from the toolbar while the wallet is open and display it only in wizard pages. It's probably not used by many users and the language can still be changed in Settings > Interface.

GBKS:
Here's a quick mock-up of a dropdown if we want to bundle these options ([Figma](https://www.figma.com/file/DplJ2DDQfIKiuRvolHX2hN/Monero-GUI?node-id=1575%3A5250)).

monero-top-bar-dropdown-gbks-200210

selsta:
I would put the `…` on the left side of the titlebar but like the idea otherwise.

selsta:
This got implemented in #3859 and #3869

PR #2735, opened 2020-01-19
Transfer: All button displays unlocked balance in amount field

### LINKED ISSUE #2466: Send page: All button should fill the amount with a number
image

When user clicks in All button, it should just fill the amount value in the amount field (in XMR), and not display "(all)". 

(all) is not specific. What is (all)? All balance in the wallet? In the account?

rating89us:
Closes #2466

The All button was added in commit https://github.com/monero-project/monero-gui/commit/c779b376fc06c1912b896549b5f39c4a6ef0ec5a#diff-3d442cd229041f412bc48acdcee456bc

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -843,37 +843,35 @@ ApplicationWindow {
         transactionDescription = description;
 
         // validate amount;
-        if (amount !== "(all)") {
-            var amountxmr = walletManager.amountFromString(amount);
-            console.log("integer amount: ", amountxmr);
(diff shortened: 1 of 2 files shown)


xiphon:
Recipient will receive less coins than displayed. You have to display the correct amount.

rating89us:
You're right. Amount field should display unlocked amount - fee. 

But fee isn't calculated yet.

PR #2739, opened 2020-01-22
Transfer: display estimated transaction fee, requires #6302

xiphon:
monero-gui-estimate-fee

### DIFF
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -133,3 +133,7 @@ function capitalize(s){
     if (typeof s !== 'string') return ''
     return s.charAt(0).toUpperCase() + s.slice(1)
 }
+
+function removeTrailingZeros(value) {
+    return (value + '').replace(/(\.\d*[1-9])0+$/, '$1');
+}
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -27,6 +27,7 @@
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import QtQuick 2.9
+import QtQuick.Controls 1.4
 import QtQuick.Layouts 1.1
 import QtQuick.Dialogs 1.2
 import moneroComponents.Clipboard 1.0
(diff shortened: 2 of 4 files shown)


SamsungGalaxyPlayer:
Should we also show the USD or desired fiat current conversion for the fee?

xiphon:
> Should we also show the USD or desired fiat current conversion for the fee?

Updated.

monero-gui-estimate-fee-usd

rating89us:
When estimated fee is less than 0.01 USD, we should display it as (<0.01 USD), instead of (~0.00 USD).

> > Should we also show the USD or desired fiat current conversion for the fee?
> 
> Updated.
> 
> monero-gui-estimate-fee-usd

selsta:
I guess this will stay open for a while until we update to libwallet_api to master?

xiphon:
> When estimated fee is less than 0.01 USD, we should display it as (<0.01 USD), instead of (~0.00 USD).

Fixed.

> I guess this will stay open for a while until we update to libwallet_api to master?

Yep.

rating89us:
We should remove the extra zeros at the end of the estimated fee:
`0.000023440000 XMR` should be `0.00002344 XMR`

xiphon:
> remove the extra zeros at the end of the estimated fee

Updated.

js/Utils.js
```diff
@@ -133,3 +133,7 @@ function capitalize(s){
     if (typeof s !== 'string') return ''
     return s.charAt(0).toUpperCase() + s.slice(1)
 }
+
+function removeTrailingZeros(value) {
+    return (value + '').replace(/(\.\d*[1-9])0+/, '$1');
```
rating89us:
```suggestion
    var lastchar = value.charAt(value.length-1);
    if (lastchar == "0") {
        return (value + '').replace(/(\.\d*[1-9])0+/, '$1');
    } else {
        return value;
    }
```
I'm trying to use this function in the amount value of tx confirmation popup (which is still WIP).
When the amount is 0.000101010101, it returns 0.00010101011. So it should run only when the last digit is 0.

xiphon: Fixed

PR #2750, opened 2020-01-29
Transfer: redesign transaction confirmation popup

### LINKED ISSUE #2467: Transaction confirmation improvements
image

1) Simple mode shouldn't display `Ring size`, `Number of transactions` and `Spending address index`.

2) There should be an additional field `From:`, with the following format:
- From: "_Wallet_" / "_Account_"
- `From: John's wallet / Savings`

3) `Address:` should be renamed into `To:`

4) `Amount:` and `Fee:` should display XMR

5) Amount converted in fiat (USD) should also appear
- in a second line: `Amount (USD):` and `Amount (EUR):`
- or in the same line: `Amount: 1.0 XMR (54 USD)`

6) `Fee:` shouldn't display extra zeros at the end
- Currently: `Fee: 0.0000248200`
- Expected: `Fee: 0.00002482`

7) `Ok` button should be renamed into `Confirm transaction`

rating89us:
Closes #2467

After:
image

Before:
image

Changes:
* Added blur behind the transaction confirmation popup
* Fiat conversion to transaction amount and fee
* Added From: (spending Wallet & Account)
* Added "XMR" after amount and fee
* Total amount (transaction amount + fee)
* Amount and fee don't display extra zeros at the end anymore
* Buttons now have fontawesome icons
* "Ok" button renamed to "Confirm transaction"
* Advanced mode now displays Description, Ring size, and Spending address(es) index(es),


xiphon:
* Would be nice to avoid scrolling
* Something is wrong with the button icons (they are on the opposite sides)

rating89us:
> * Would be nice to avoid scrolling

Scrolling is only visible in advanced mode, because in this mode Description, Ring size, and Spending address(es) index(es) are visible. Increasing the height of the popup would make it strange in Simple mode.

> * Something is wrong with the button icons (they are on the opposite sides)

It was done on purpose. See trezor example:
image

Issue #2544, opened 2019-12-03
New wallet: password fields aren't displaying lock icon


rating89us:
Current:
![image](https://user-images.githubusercontent.com/45968869/70032648-23f56900-15ae-11ea-83cd-45d72dde9ce7.png)

Expected:
![image](https://user-images.githubusercontent.com/45968869/70032660-2d7ed100-15ae-11ea-879c-bc02f2e1bd2c.png)

xiphon:
Can't reproduce

selsta:
I will fix it

PR #2752, opened 2020-01-29
[WIP] Transfer: redesign transaction flow

### LINKED ISSUE #2467: Transaction confirmation improvements

rating89us:
Closes #2467

After:
image

Before:
image

Changes:
* Added blur behind the transaction confirmation popup
* Added From: (spending Wallet & Account)
* Amount with larger font
* Amount and fee: removed extra zeros at the end + "XMR" suffix + fiat conversion
* Buttons have the same width
* "Cancel" button renamed to "Back"
* "Ok" button renamed to "Confirm" + confirm icon
* Removed payment ID, ring size, transaction count and spending address(es) index(es)

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -806,21 +806,30 @@ ApplicationWindow {
                     + ", fee: " + walletManager.displayAmount(transaction.fee));
 
             // here we show confirmation popup;
-            transactionConfirmationPopup.title = qsTr("Please confirm transaction:\n") + translationManager.emptyString;
-            transactionConfirmationPopup.text = "";
-            transactionConfirmationPopup.text += (address === "" ? "" : (qsTr("Address: ") + address));
-            transactionConfirmationPopup.text += (paymentId === "" ? "" : (qsTr("\nPayment ID: ") + paymentId));
-            transactionConfirmationPopup.text +=  qsTr("\n\nAmount: ") + walletManager.displayAmount(transaction.amount);
-            transactionConfirmationPopup.text +=  qsTr("\nFee: ") + walletManager.displayAmount(transaction.fee);
-            transactionConfirmationPopup.text +=  qsTr("\nRingsize: ") + (mixinCount + 1);
-            transactionConfirmationPopup.text +=  qsTr("\n\nNumber of transactions: ") + transaction.txCount
-            transactionConfirmationPopup.text +=  (transactionDescription === "" ? "" : (qsTr("\nDescription: ") + transactionDescription))
-            for (var i = 0; i < transaction.subaddrIndices.length; ++i){
-                transactionConfirmationPopup.text += qsTr("\nSpending address index: ") + transaction.subaddrIndices[i];
-            }
+			var transactionAmount = Number(walletManager.displayAmount(transaction.amount));
+			var transactionFee = Number(walletManager.displayAmount(transaction.fee));
 
-            transactionConfirmationPopup.text += translationManager.emptyString;
-            transactionConfirmationPopup.icon = StandardIcon.Question
+            function showFiatConversion(valueXMR) {
+                return (fiatApiConvertToFiat(valueXMR) === "0.00" ? "&#60;0.01 " + fiatApiCurrencySymbol() : "~" + fiatApiConvertToFiat(valueXMR) + " " + fiatApiCurrencySymbol());
+            }
+			
+            transactionConfirmationPopup.title = qsTr("Confirm send") + translationManager.emptyString
+            transactionConfirmationPopup.text =  "";
+            transactionConfirmationPopup.text += "<style>td{padding:7px;}
+                                                         td.amount{padding:3px; font-size:40px; font-weight:bold; text-align:center;}
+                                                         td.amountfiat{padding:0px; height:70px; font-size:17px; text-align:center; vertical-align:top;}
+                                                         td.bold{font-size:16px; font-weight:bold; width:20%;}
+                                                         td.normal{font-size:15px; width:80%;}
+                                                  </style>"
+                                              +  "<table>"
+                                              +  "<tr><td class='amount' colspan=2>" + transactionAmount + " XMR</td>"
+                                              +  "<tr><td class='amountfiat' colspan=2>" + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionAmount) : "") + "</td></tr>"
+                                              +  "<tr><td class='amountfiat' colspan=2></td></tr>"
+                                              +  "<tr><td class='bold'>" + qsTr("From") + ":</td><td class='normal'>" + walletName + " (" + currentWallet.getSubaddressLabel(currentWallet.currentSubaddressAccount, 0) + ")" + "</td></tr>"
+                                              +  "<tr><td class='bold'>" + qsTr("To") + ":</td><td class='normal'>" + address + "</td></tr>"
+                                              +  "<tr><td class='bold'>" + qsTr("Fee") + ":</td><td class='normal'>" + transactionFee + " XMR " + (persistentSettings.fiatPriceEnabled ? "(" + showFiatConversion(transactionFee) + ")" : "")+ "</td></tr>"
+                                              +  "</table>"
+                                              +  translationManager.emptyString
             transactionConfirmationPopup.open()
         }
     }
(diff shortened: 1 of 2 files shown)


selsta:
Almost good now. Please look into the `StandardDialog.qml` property aliases. There already is `property alias okText: okButton.text` and `property alias cancelText: cancelButton.text` which you can set from main.qml. Also add a width and label/font alias.

`StandardDialog.qml` should only contain the default implementation, not `transactionConfirmationPopup.visible`.

xiphon:
The order should be 
1 From
2 To
3 Send
4 etc..

main.qml
```diff
@@ -806,19 +806,33 @@ ApplicationWindow {
...
+            var transactionFee = Number(walletManager.displayAmount(transaction.fee));
+            var totalAmount = Number(transactionAmount + transactionFee);
+
+            function showFiatConversion(valueXMR) {
+                return (fiatApiConvertToFiat(valueXMR) === "0.00" ? "(<0.01 " + fiatApiCurrencySymbol() + ")" : "(~" + fiatApiConvertToFiat(valueXMR) + " " + fiatApiCurrencySymbol() + ")");
             }
 
+            transactionConfirmationPopup.title = qsTr("Confirm transaction") + translationManager.emptyString;
+            transactionConfirmationPopup.text =  ""; //cleaning text because user may have canceled the confirmation popup previously
+            transactionConfirmationPopup.text += "Send:\n" + transactionAmount + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionAmount) : "");
```
xiphon: qsTr

main.qml
```diff
@@ -806,19 +806,33 @@ ApplicationWindow {
...
+            transactionConfirmationPopup.text += "\n\nFrom:\n" + walletName + " (" + currentWallet.getSubaddressLabel(currentWallet.currentSubaddressAccount, 0) + ")";
+            transactionConfirmationPopup.text += (address === "" ? "" : (qsTr("\n\nTo:\n") + address));
+            transactionConfirmationPopup.text +=  qsTr("\n\nFee:\n") + transactionFee + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionFee) : "");
```
xiphon: Strip the tags from translatable string

main.qml
```diff
@@ -806,19 +806,33 @@ ApplicationWindow {
...
+            transactionConfirmationPopup.text +=  qsTr("\n\nTotal amount:\n") + totalAmount + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(totalAmount) : "");
+
+            if(appWindow.walletMode >= 2) {
+                // transactionConfirmationPopup.text +=  (paymentId === "" ? "" : (qsTr("\n\nPayment ID:\n") + paymentId));
```
xiphon: Remove

main.qml
```diff
@@ -806,19 +806,33 @@ ApplicationWindow {
...
+                transactionConfirmationPopup.text +=  (transactionDescription === "" ? "" : (qsTr("\n\nDescription:\n") + transactionDescription));
+                transactionConfirmationPopup.text +=  qsTr("\n\nRingsize:\n") + (mixinCount + 1);
+                // transactionConfirmationPopup.text +=  qsTr("\n\nNumber of transactions:\n") + transaction.txCount;
```
xiphon: Remove

main.qml
```diff
@@ -806,21 +806,27 @@ ApplicationWindow {
...
+            transactionConfirmationPopup.title = qsTr("Confirm send") + translationManager.emptyString;
+            transactionConfirmationPopup.text =  "";
+            transactionConfirmationPopup.text += qsTr("From") + ":\n" + walletName + " (" + currentWallet.getSubaddressLabel(currentWallet.currentSubaddressAccount, 0) + ")";
+            transactionConfirmationPopup.text += "\n\n" + qsTr("To") + ":\n" + address;
+            transactionConfirmationPopup.text += "\n\n" + qsTr("Amount") + ":\n" + transactionAmount + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionAmount) : "");
+            transactionConfirmationPopup.text += "\n\n" + qsTr("Fee") + ":\n" + transactionFee + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionFee) : "");
+
             transactionConfirmationPopup.text += translationManager.emptyString;
             transactionConfirmationPopup.icon = StandardIcon.Question
+            transactionConfirmationPopup.cancelText = qsTr("Back") + translationManager.emptyString;
```
xiphon: Move it to `transactionConfirmationPopup` properties ->
https://github.com/monero-project/monero-gui/blob/35b7a5f6bb469583f1e78c9f35f154d312d99dc6/main.qml#L1399

main.qml
```diff
@@ -1413,6 +1414,12 @@ ApplicationWindow {
     StandardDialog {
         z: parent.z + 1
         id: transactionConfirmationPopup
+        dialogHeight: 380
+        icon: StandardIcon.Question
+        cancelText: qsTr("Back") + translationManager.emptyString
+        cancelButton.width: okButton.width
+        okButton.rightIcon: "qrc:///images/rightArrow.png"
```
xiphon: Use FontAwesome icon. For example, solid `arrowCircleRight` or `checkCircle`.

main.qml
```diff
@@ -930,12 +924,19 @@ ApplicationWindow {
...
+                return (fiatApiConvertToFiat(valueXMR) === "0.00" ? "(<0.01 " + fiatApiCurrencySymbol() + ")" : "(~" + fiatApiConvertToFiat(valueXMR) + " " + fiatApiCurrencySymbol() + ")");
+            }
+            
+            transactionConfirmationPopup.title = qsTr("Confirm transaction") + translationManager.emptyString
+            transactionConfirmationPopup.text =  "";
+            transactionConfirmationPopup.text += qsTr("Please confirm transaction to sweep unmixable outputs") + ":\n\n"
+                                              +  qsTr("Amount") + ":\n" + transactionAmount + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionAmount) : "") + "\n\n"
+                                              +  qsTr("Fee") + ":\n" + transactionFee + " XMR " + (persistentSettings.fiatPriceEnabled ? showFiatConversion(transactionFee) : "")
+                                              +  translationManager.emptyString
             transactionConfirmationPopup.icon = StandardIcon.Question
```
xiphon: Seems we don't need this anymore

main.qml
rating89us: FontAwesome doesn't work with `.rightIcon`

Also, if we change button text to `qsTr("Confirm") + FontAwesome.checkCircle`, the FontAwesome isn't displayed on Windows (see #2544).

If we try the solution of #2556 (change button fontFamily to FontAwesome.fontFamilySolid), the button text has a font that is bold and not so nice, see below:
image

I used the same png icon as the send button, but maybe we could create a png icon with a checkCircle image. Also, if it is possible, we could make the button use two font families, but I don't know if it is possible. Any other idea?

main.qml
```diff
@@ -1407,6 +1414,12 @@ ApplicationWindow {
     StandardDialog {
         z: parent.z + 1
         id: transactionConfirmationPopup
+        dialogHeight: 380
```
selsta: dialogHeight?

GBKS:
Not sure how relevant this is, but I also had a design for this some time back (actually for the whole transaction flow). [Here it is in Figma](https://www.figma.com/file/DplJ2DDQfIKiuRvolHX2hN/Monero-GUI?node-id=59%3A5597). Hope it's helpful.

main.qml
rating89us: Do you mean we should just remove this lines and have a cleaner transactionConfirmationPopup, 
or that the transactionConfirmationPopup is duplicated in the code and we should call `onTransactionCreated() `from `handleSweepUnmixable`?

xiphon: Was refferring to `transactionConfirmationPopup.icon = StandardIcon.Question` line, you already set the icon in `transactionConfirmationPopup` properties.

main.qml
xiphon: Gotcha, let's use `rightArrow.png` for now and update it to `FontAwesome.checkCircle` later.

xiphon:
> Not sure how relevant this is, but I also had a design for this some time back (actually for the whole transaction flow). [Here it is in Figma](https://www.figma.com/file/DplJ2DDQfIKiuRvolHX2hN/Monero-GUI?node-id=59%3A5597). Hope it's helpful.

Yeah, would prefer to use table layout.

GBKS:
Would it be possible to combine this review overlay and the previous one that shows a loader while the transaction is being prepared? I also mocked this up in Figma, the buttons are replaced with a “Preparing transaction...” load indicator until it’s done. That way users don’t have to wait and look at a spinner, they can use that time to review the transaction details.

main.qml
rating89us: I decided to add a checkCircle.png as icon

xiphon:
Please don't add new `png` images, we are steadily dropping such files and replacing them with FontAwesome glyphs instead.

main.qml
```diff
@@ -924,13 +933,19 @@ ApplicationWindow {
                     + ", fee: " + walletManager.displayAmount(transaction.fee));
 
             // here we show confirmation popup;
+            var transactionAmount = Number(walletManager.displayAmount(transaction.amount));
```
selsta: This results in small amounts getting displayed like this: `1e-8 XMR`, the average user doesn’t know what this means.

rating89us:
@GBKS Thanks, I'm redesigning the whole transaction flow according to your design. Soon I'll update it.

GBKS:
@rating89us awesome! Let me know if you need anything. I bet there are some details I missed.

rating89us:
Please check new PR (#2779)

PR #2759, opened 2020-02-01
MenuButton: fix gradient

selsta:
Before
Screenshot 2020-02-01 at 23 02 17
After
Screenshot 2020-02-01 at 23 00 35


xiphon:
I guess the gradient here is to highlight the parent menu item.
Could you do a screenshot of `Advanced` -> `Sing/verify` selected?

selsta:
> I guess the gradient here is to highlight the parent menu item.

Yes but IMO it looks off/broken.

Screenshot 2020-02-02 at 00 35 05

PR #2773, opened 2020-02-13
macOS: add version to info.plist

### LINKED ISSUE #1578: version numbers missing on macOS
i've downloaded Version '0.12.3.0' but when selecting it in the Finder and using 'Get Info' it doesn't show any Version-number at all. i guess the **CFBundleVersion** and **CFBundleShortVersionString** entries in your **Info.plist** file are missing.
all apps on macOS are supposed to contain valid version information.

### LINKED ISSUE #2334: Add version to Monero wallet file properties
Based on a [support question](https://www.reddit.com/r/monerosupport/comments/cl9osc/cant_find_the_version_of_the_monero_wallet_gui/) that came up on MoneroSupport, it seems that people might try to look at the file properties when determining the wallet version. Supporting the display of the version number here seems like a great UX feature to me.

selsta:
Screenshot 2020-02-13 at 03 45 42

Ref: https://stackoverflow.com/a/19981753

Fixes #2334
Fixes #1578

### DIFF
--- a/share/Info.plist
+++ b/share/Info.plist
@@ -26,6 +26,12 @@
   <key>CFBundleIdentifier</key>
   <string>org.monero-project.monero-wallet-gui</string>
 
+  <key>CFBundleVersion</key>
+  <string>@VERSION_LONG@</string>
+
+  <key>CFBundleShortVersionString</key>
+  <string>@VERSION@</string>
+
   <key>NSSupportsAutomaticGraphicsSwitching</key>
   <true/>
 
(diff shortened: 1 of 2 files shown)


share/Info.plist
```diff
@@ -26,6 +26,12 @@
   <key>CFBundleIdentifier</key>
   <string>org.monero-project.monero-wallet-gui</string>
 
+  <key>CFBundleVersion</key>
+  <string>@VERSION_LONG@</string>
```
xiphon: Shouldn't start with a `v` prefix

xiphon:
> You can include more integers but the system ignores them.
> This key is required by the App Store and is used throughout the system to identify the version of the build. For macOS apps, increment the build version before you distribute a build.

Makes me think we might have some (unknown to me) problems. `0.15.0.1` and `0.15.0.3` will be identified as the same `0.15.0` version.

Issue #2757, opened 2020-02-01
Request: addressBook function getDescription(address)


rating89us:
It would be nice to display the name of an address book contact besides its address in pages other than the Address book, like Transfer page (address field, transaction confirmation pop up) and Transactions page.

In order to do that, it is necessary a function like `getDescription(address)`, which accepts an address, checks if it is included in Address Book, and then returns the address name (description).

Issue #2595, opened 2019-12-15
Address book: different contacts can use the same address


rating89us:
It is possible to create two or more different contacts, all using the same Monero address.

xiphon:
That's okay

selsta:
Closing as intended behaviour.

PR #2775, opened 2020-02-15
History: display Address Book names for known destination addresses

xiphon:
Closes https://github.com/monero-project/monero-gui/issues/2757

monero-gui-history-names

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -633,7 +633,17 @@ Rectangle {
                                 MoneroComponents.TextPlain {
                                     font.family: MoneroComponents.Style.fontRegular.name
                                     font.pixelSize: 15
-                                    text: isout ? qsTr("Sent") : qsTr("Received") + translationManager.emptyString
+                                    text: {
+                                        if (!isout) {
+                                            return qsTr("Received") + translationManager.emptyString;
+                                        }
+                                        const addressBookName = currentWallet ? currentWallet.addressBook.getDescription(address) : null;
+                                        if (!addressBookName)
+                                        {
+                                            return qsTr("Sent") + translationManager.emptyString;
+                                        }
+                                        return addressBookName;
+                                    }
                                     color: MoneroComponents.Style.historyHeaderTextColor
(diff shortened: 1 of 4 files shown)


selsta:
I wonder if Sent / Received could still be there + the address book name. Unless you know what the green / orange mark means it is not clear if the transaction is incoming / outgoing.

But changing it to `Sent to John Doe` would take up a lot of space.

pages/History.qml
```diff
@@ -633,7 +633,17 @@ Rectangle {
                                 MoneroComponents.TextPlain {
                                     font.family: MoneroComponents.Style.fontRegular.name
                                     font.pixelSize: 15
-                                    text: isout ? qsTr("Sent") : qsTr("Received") + translationManager.emptyString
+                                    text: {
+                                        if (!isout) {
+                                            return qsTr("Received") + translationManager.emptyString;
+                                        }
+                                        const addressBookName = currentWallet.addressBook.getDescription(address);
```
selsta:
```
2020-02-15 15:11:22.911	W qrc:/pages/History.qml:640: TypeError: Cannot read property 'addressBook' of undefined
```
shows up when closing the wallet.

xiphon: Updated

xiphon:
> I wonder if Sent / Received could still be there + the address book name. Unless you know what the green / orange mark means it is not clear if the transaction is incoming / outgoing.
> 
> But changing it to `Sent to John Doe` would take up a lot of space. Not sure what the best solution is here.

Feel free to propose any other design option.

rating89us:
I can't test it right now, but have you considered that a single address can correspond to multiple contacts in address book (#2595)?
In this case I believe it should return the contacts separated by comma: "Alice, Bob"

xiphon:
> we should stop allowing multiple contacts to have the same address

We can't.

> have you considered that a single address can correspond to multiple contacts in address book?

Yes. It is intended to show the most recently added contact.

PR #2780, opened 2020-02-17
Transfer: add Address Book inline button + improvements in Address Book

rating89us:
Changes:
- Remove "(Address Book)" link from address field label
- Add address book inline button with icon + adjust right margin of address field
- Adjust margins for titles and subtitles of Address Book subpages
- Add "Select an address from your Address Book" subtitle + "Back" button when address book is called from Transfer page

Before:
image

After:
inlineaddressbutton


selsta:
Please rebase.

Did you check if this works together with the QR code scanner icon?

xiphon:
And please split it and send unrelated commit as a separate PR

rating89us:
See #2831

PR #2802, opened 2020-03-08
Include icon for xdg menu entry

tobtoht:
Closes #2292

### DIFF
--- a/qml.qrc
+++ b/qml.qrc
@@ -236,6 +236,7 @@
         <file>images/themes/white/question.svg</file>
         <file>components/effects/ColorTransition.qml</file>
         <file>components/effects/GradientBackground.qml</file>
+        <file>images/appicons/64x64.png</file>
         <file>images/check-white.svg</file>
         <file>images/copy.svg</file>
         <file>images/edit.svg</file>
--- a/src/qt/utils.cpp
+++ b/src/qt/utils.cpp
@@ -28,6 +28,7 @@
 
 #include <QtCore>
 #include <QApplication>
+#include <QPixmap>
 
 #include "TailsOS.h"
 #include "utils.h"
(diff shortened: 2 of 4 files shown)


selsta:
Does this get written every time the GUI gets opened? Is this ok? Looks good otherwise.

tobtoht:
>Does this get written every time the GUI gets opened?

Hm, yeah. It's the same behavior as the .desktop, but it's not really necessary for the icon. I can have it check for existence.

selsta:
@xiphon Is writing on every start ok? Would you prefer existence check?

src/qt/utils.cpp
```diff
@@ -100,7 +114,6 @@ void registerXdgMime(QApplication &app){
 
     if (TailsOS::detect() && TailsOS::detectDotPersistence() && TailsOS::usePersistence) {
         TailsOS::persistXdgMime(filePath, mime);
-        return;
```
xiphon: ?

qml.qrc
```diff
@@ -241,5 +241,6 @@
         <file>images/edit.svg</file>
         <file>images/arrow-right-in-circle-outline-medium-white.svg</file>
         <file>images/tails-grey.png</file>
+        <file>images/appicons/64x64.png</file>
```
xiphon: Could you please move this before the line 239 ` <file>images/check-white.svg</file>`

src/qt/utils.cpp
```diff
@@ -58,6 +59,19 @@ bool fileWrite(QString path, QString data) {
     return false;
 }
 
+bool pixmapWrite(QString path, QPixmap pixmap) {
```
xiphon: Take `path` by const ref or std::move it, take `pixmap` by const ref.

xiphon:
> @xiphon Is writing on every start ok? Would you prefer existence check?

I don't like malware-like behavior, i.e. not prompting a user and silently dropping some files to the system at every startup.
And there is no way to cleanup the system afterwards.

But the code doing this is already there.
Would say it's out of scope of the current PR.

Would be nice to have existence check at least.

src/qt/utils.cpp
tobtoht: desktop file and icon should also be written to default locations on Tails.

tobtoht:
Replaced by #2984

PR #2817, opened 2020-03-31
network: provide common HTTP GET functionality with js callbacks

xiphon:


src/qt/network.cpp
```diff
@@ -37,19 +39,18 @@
...
 {
 }
 
-void Prices::getJSON(const QString url) const
+void Network::get(const QString url, const QJSValue& callback, const QString &content_type) const
 {
     qDebug() << QString("Fetching: %1").arg(url);
 
-    m_scheduler.run([this, url] {
+    m_scheduler.run([this, url, content_type] {
```
selsta:
```suggestion
    m_scheduler.run([url, content_type] {
```

See macOS CI, fails due to werror

xiphon: Done

PR #2819, opened 2020-04-04
UpdateDialog: implement update download functionality

xiphon:
Will implement and commit signature and hash sum verification in a separate PR.

monero-gui-update-download

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1436,6 +1436,14 @@ ApplicationWindow {
         }
     }
 
+    MoneroComponents.UpdateDialog {
+        id: updateDialog
+
+        allowed: !passwordDialog.visible && !inputDialog.visible && !splash.visible
+        x: (parent.width - width) / 2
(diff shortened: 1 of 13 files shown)


main.qml
```diff
@@ -1181,7 +1181,7 @@ ApplicationWindow {
     }
 
     function fiatApiJsonReceived(url, resp, error) {
-        if (error) {
+        if (error) {
```
selsta, line 1184: Small thing I noticed: appears that this only removes carriage return. Is this intended? Other source documents like src/qt/downloader.cpp/h are full of carriage return.

xiphon: > appears that this only removes carriage return. Is this intended?

Nope, did it accidentally. But... let's fix it anyway.

> src/qt/downloader.cpp/h are full of carriage return.

Fixed.

xiphon:
Updated, dropped unused Notifier component.

PR #2824, opened 2020-04-07
'--verify-update', shasum support, OpenPGP signatures verification

xiphon:
Implemented `--verify-update` command-line option.

The code will be used in Monero GUI one-click update downloading and verification mechanism.

Verifies update binary using `shasum`-compatible (SHA256 algo) output signed by two maintainers.  
* Requires `hashes.txt` - signed `shasum` output generated by a maintainer:
  `gpg -o hashes.txt --clear-sign <shasum_output>`
* Requires `hashes.txt.sig` - detached signature of `hashes.txt` generated by another maintainer, signing the whole `hashes.txt` contents (including the first maintainer's signature):
  `gpg -b hashes.txt`

Hardcoded @binaryFate, @fluffypony , @luigi1111  public keys.
Update verification requires using any 2 of them.
```
monero/utils/gpg_keys/binaryfate.asc
monero/utils/gpg_keys/fluffypony.asc
monero/utils/gpg_keys/luigi1111.asc
```

Sample output:
```
$ ./monero-gui --verify-update /some/folder/monero-gui-win-x64-v0.15.0.4.zip
successfully verified, signed by "binaryFate <binaryfate@getmonero.org>" and "luigi1111 <luigi1111w@gmail.com>"
```


binaryFate:
IMO it's a solid scheme for automatic updates, and should not interfere with people checking hashes "by hand" or confuse them.

selsta:
Can you add license headers to the new files?

xiphon:
> Can you add license headers to the new files?

Done

src/qt/updater.cpp
```diff
@@ -0,0 +1,95 @@
...
+    hasher << epee::span<const uint8_t>(reinterpret_cast<const uint8_t *>(data), size);
+    const std::vector<uint8_t> hash = hasher.finish();
+    return QByteArray(reinterpret_cast<const char *>(&hash[0]), hash.size());
+}
+
+QByteArray Updater::parseShasumOutput(const QString &message, const QString &filename) const
+{
+    for (const auto &line : message.splitRef("\n"))
+    {
+        if (!line.startsWith(filename))
```
selsta, line 59: Does this match with the current hashes.txt format? Shouldn’t the line start with the hash, then double space and then the filename?

I added my key to the source, detached signed the hashes.txt and tried it out:
```
selsta@mbpR ~/d/monero-gui (2824) [1]> build/release/bin/monero-wallet-gui.app/Contents/MacOS/monero-wallet-gui --verify-update ~/Downloads/monero-gui-mac-x64-v0.15.0.4.tar.bz2
2020-04-08 00:03:57.244	W app startd (log: /Users/selsta/Library/Logs/monero-wallet-gui.log)
2020-04-08 00:03:57.261	E hash not found
```

xiphon: Fixed. Supports both formats now.

garlicgambit:
Some general thoughts about automatic updates and its implementation.  
  
Automatic updates are disabled by default, correct? Will that policy change in the future? Will it support an unattended update mode?   
  
A 2/3 scheme for automatic updates is a pretty low security margin for such an important task. The signers could get hacked or be compelled to do malicious things. Five or more geographically separated people with good opsec would give us @garlicgambit a reasonable level of comfort. The signers don't have to be project maintainers. Reputable devs can also be part of this group of signers.   
  
What happens when this check fails? How robust is it? Can you do a denial of service to prevent it from ever checking or downloading updates? What type of message will the user get if the download or checksum fails?  
  
Will users be warned to verify the authenticity of the update out-of-band (check getmonero.org, reddit, chat)? Please add a call to action to always verify out-of-band.  
  
Will this also be ported to *BSD platforms? There is currently limited to no GUI support on them, but that will hopefully change in the future.

selsta:
@garlicgambit Considering that the GUI does not have reproducible builds yet, the auto updater is not the attack vector. If the binaries themselves are compromised then it does not matter if 2, 4 or 100 people sign them. If your threat models considers multiple core team members getting fully compromised at the same time then I would consider compiling from source from your own repository where you verify every commit yourself.

> Automatic updates are disabled by default, correct? Will that policy change in the future? Will it support an unattended update mode?

See #2819. You still have to manually install the new version.

>  What happens when this check fails? How robust is it? Can you do a denial of service to prevent it from ever checking or downloading updates? What type of message will the user get if the download or checksum fails?

If the check fails then it does not notify you about a new update.

src/qt/updater.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+    for (const auto &maintainer : m_maintainers)
+    {
+        if (signature.verify(data, maintainer))
+        {
+            return QString::fromStdString(maintainer.user_id());
+        }
+    }
+
+    throw std::runtime_error("not singed by a maintainer");
```
selsta, line 129:
```suggestion
    throw std::runtime_error("not signed by a maintainer");
```

xiphon:
Updated, fixed a typo

src/openpgp/hash.h
```diff
@@ -0,0 +1,101 @@
...
+  hash &operator=(const hash &) = delete;
+
+  hash(uint8_t algorithm)
+    : algorithm(algorithm)
+  {
+    if (gcry_md_open(&md, algorithm, 0))
+    {
+      throw std::runtime_error("failed to create message digest object");
+    }
+  }
```
moneromooo-monero, line 57: consumed not initialized anywhere

xiphon: Fixed

xiphon:
Updated. Simplified the text parsing code.

src/openpgp/openpgp.cpp
```diff
@@ -0,0 +1,380 @@
...
+}
+
+s_expression signature_rsa::hash_bytes(const epee::span<const uint8_t> message, size_t public_key_bits) const
+{
+  const std::vector<uint8_t> plain_hash = (hash(m_hash_algorithm) << message << m_hashed_appendix).finish();
+  if (plain_hash.size() < sizeof(m_hash_left16_bits))
+  {
+    throw std::runtime_error("insufficient message hash size");
+  }
+  if (*reinterpret_cast<const uint16_t *>(&plain_hash[0]) != m_hash_left16_bits)
```
moneromooo-monero, line 313: That looks endian dependent (might not care ?)

src/openpgp/s_expression.h
```diff
@@ -0,0 +1,78 @@
...
+  {
+    if (gcry_sexp_build(&data, nullptr, args...) != GPG_ERR_NO_ERROR)
+    {
+      throw std::runtime_error("failed to build S-expression");
+    }
+  }
+
+  s_expression(s_expression &&other)
+  {
+    std::swap(data, other.data);
```
moneromooo-monero, line 56: That's an interesting way to do it. Thanks I'll remember it :)

src/openpgp/openpgp.cpp
xiphon: Update. Is a pair of bytes now.

moneromooo-monero: Sorry to complain again, but AFAIK function parameter evaluation order is undefined (and that's a function in disguise).

PR #2828, opened 2020-04-10
TitleBar: fix close icon fallback

selsta:


xiphon:
Don't you want to drop the image and always use Font Awesome icon?

selsta:
No, the font awesome icon is too thick and does not really fit in, we just use it as a backup because is seems like `QMLSCENE_DEVICE=softwarecontext` can’t render SVGs.

The current ones were designed by GKBS for the GUI specifically.

PR #2829, opened 2020-04-12
trezor: support new passphrase entry mechanism, requires #6446

ph4r05:
- device/host decision is now made on the host (due to the change in Trezor firmware)
- passphrase can be prompted also when wallet is initialized (thus I had to add passphrase support to the wallet listener)
- requires https://github.com/monero-project/monero/pull/6446 to work

GUI now asks where does user want to enter the passphrase hardware wallet or host.
Screenshot 2020-04-12 at 19 42 33

If host is selected, passphrase prompt is displayed
Screenshot 2020-04-12 at 19 39 27

If device is selected the app behaves in the same way as previously. 
Screenshot 2020-04-11 at 12 32 09

Unrelated to this, but https://github.com/monero-project/monero/pull/6446 fixes also missing fee information:
Screenshot 2020-04-11 at 13 49 46


selsta:
Can you blur the background? https://github.com/monero-project/monero-gui/blob/86252506f01a8069ffbf01bed2ed8f44123c5233/main.qml#L1694

ph4r05:
> Can you blur the background when passphraseDialog is visible?
> 
> https://github.com/monero-project/monero-gui/blob/86252506f01a8069ffbf01bed2ed8f44123c5233/main.qml#L1694

Thanks for the suggestion. I changed it, but now I realized I have to make small adjustments to the passphrase logic so I switched PR to draft.

ph4r05:
Done

ph4r05:
Ah. I've noticed QScopeGuard is available only since Qt 5.12. I will address this.

ph4r05:
@selsta CI builds are failing as PR monero dependency is not resolved. 
Any idea why this is not working in CI?

https://github.com/monero-project/monero-gui/blob/cd9ca2daecdf3d81c4c7d65de79e97f2387d6942/get_libwallet_api.sh#L38

selsta:
I don’t know but once the PR is merged you can force push to rerun the CI.

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) 2014-2019, The Monero Project
```
selsta, line 1: 2020 :)

ph4r05:
> * Can you remove the black rectangle box from here? Our password dialogs usually only have blur with no background.
> Can you remove the black rectangle box from here? Our password dialogs usually only have blur with no background.

@selsta do you mean background for password box or background around the whole dialog?

selsta:
@ph4r05 The black rectangle background for the password box. The StandardDialog is okay.

selsta:
Screenshot 2020-05-05 at 21 48 22
Our password entry looks like this

ph4r05:
@selsta like this?
Screenshot 2020-05-06 at 09 27 23
Screenshot 2020-05-06 at 09 27 29

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,158 @@
...
+            onRejectedPassphrase: {
+                if (onRejectedCallback)
+                    onRejectedCallback();
+            }
+            onCloseCallback: {
+                root.close()
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
ph4r05: done

rating89us:
Your password textboxes are still black. We use it transparent with blur, see: 
image

selsta:
@ph4r05 Can you try to rebase with master? This should make them transparent.

ph4r05:
> Your password textboxes are still black. We use it transparent with blur, see:

Isn't it QT thing? I see it this way:
Screenshot 2020-05-06 at 09 55 00

ph4r05:
This PR should not affect theme of other components (like PasswordDialog)

selsta:
> This PR should not affect the theme of other components (like PasswordDialog)

Correct. Once rebased it looks like the other screenshots:

Screenshot 2020-05-06 at 10 22 32

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+        id: entryChooserDialog
+        title: qsTr("Hardware wallet passphrase") + translationManager.emptyString
+        text: qsTr("Please select where you want to enter passphrase.\nIt is recommended to enter passphrase on the hardware wallet for better security.") + translationManager.emptyString
+        cancelVisible: true
+        okVisible: true
+        okText: qsTr("Hardware wallet")
+        cancelText: qsTr("Host")
+        closeVisible: true
+        icon: StandardIcon.Question
+        property var isClosedWithoutAction: false
```
selsta, line 86:
```suggestion
```

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+        z: root.z + 1
+        id: entryChooserDialog
+        title: qsTr("Hardware wallet passphrase") + translationManager.emptyString
+        text: qsTr("Please select where you want to enter passphrase.\nIt is recommended to enter passphrase on the hardware wallet for better security.") + translationManager.emptyString
+        cancelVisible: true
+        okVisible: true
+        okText: qsTr("Hardware wallet")
+        cancelText: qsTr("Host")
+        closeVisible: true
+        icon: StandardIcon.Question
```
selsta, line 85:
```suggestion
```

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+    StandardDialog {
+        z: root.z + 1
+        id: entryChooserDialog
+        title: qsTr("Hardware wallet passphrase") + translationManager.emptyString
+        text: qsTr("Please select where you want to enter passphrase.\nIt is recommended to enter passphrase on the hardware wallet for better security.") + translationManager.emptyString
+        cancelVisible: true
+        okVisible: true
+        okText: qsTr("Hardware wallet")
+        cancelText: qsTr("Host")
+        closeVisible: true
```
selsta, line 84:
```suggestion
```

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+            passphraseDialog.close()
+    }
+
+    StandardDialog {
+        z: root.z + 1
+        id: entryChooserDialog
+        title: qsTr("Hardware wallet passphrase") + translationManager.emptyString
+        text: qsTr("Please select where you want to enter passphrase.\nIt is recommended to enter passphrase on the hardware wallet for better security.") + translationManager.emptyString
+        cancelVisible: true
+        okVisible: true
```
selsta, line 81:
```suggestion
```

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+        if (passphraseDialog.visible)
+            passphraseDialog.close()
+    }
+
+    StandardDialog {
+        z: root.z + 1
+        id: entryChooserDialog
+        title: qsTr("Hardware wallet passphrase") + translationManager.emptyString
+        text: qsTr("Please select where you want to enter passphrase.\nIt is recommended to enter passphrase on the hardware wallet for better security.") + translationManager.emptyString
+        cancelVisible: true
```
selsta, line 80:
```suggestion
```

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+        z: root.z + 1
+        anchors.fill: parent
+        passphraseDialogMode: true
+
+        onAcceptedPassphrase: {
+        if (onAcceptedCallback)
+            onAcceptedCallback(passphraseDialog.password);
+        }
+        onRejectedPassphrase: {
+        if (onRejectedCallback)
```
selsta, line 114: indentation looks wrong, also above

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        onCloseCallback: {
+            root.close()
+        }
+    }
+
+    PasswordDialog {
+        id: passphraseDialog
+        z: root.z + 1
```
selsta, line 105:
```suggestion
```

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import QtQuick 2.9
+import QtQuick.Controls 2.0
```
selsta, line 30: Remove imports: `QtQuick.Controls 2.0`, `QtQuick.Dialogs 1.2`, `QtQuick.Layouts 1.1`, `QtQuick.Controls.Styles 1.4`, `QtQuick.Window 2.0` `FontAwesome 1.0`, `"effects/" as MoneroEffects`, `"../js/Utils.js" as Utils`

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+import "." as MoneroComponents
+import "effects/" as MoneroEffects
+import "../js/Utils.js" as Utils
+
+Item {
+    id: root
+    anchors.fill: parent
+    z: parent.z + 1
+
+    property var canEnterOnDevice
```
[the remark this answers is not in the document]
ph4r05: done

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,121 @@
...
+    function close() {
+        root.visible = false;
+        if (entryChooserDialog.visible)
+            entryChooserDialog.close()
+        if (passphraseDialog.visible)
+            passphraseDialog.close()
+    }
+
+    StandardDialog {
+        z: root.z + 1
```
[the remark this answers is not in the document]
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

ph4r05:
@selsta thanks for a great feedback!

btw I've just figured out that StandardDialog resets the ok/cancel button labels on close. Any ideas where to set 

```
okText: qsTr("Hardware wallet")
cancelText: qsTr("Host")
```

variables so it is set correctly? Override open()?

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,106 @@
...
+    property var onAcceptedCallback
+    property var onWalletEntryCallback
+    property var onRejectedCallback
+
+    function open(canEnterOnDevice_) {
+        var canEnterOnDevice = canEnterOnDevice_ !== null ? canEnterOnDevice_ : canEnterOnDevice
+        root.visible = true;
+
+        if (canEnterOnDevice) {
+            entryChooserDialog.open()
```
selsta, line 46: You can set the `okText` and `cancelText` here and remove them from the component itself.

ph4r05: done

selsta:
@xiphon has to review the C++ part

main.qml
```diff
@@ -1537,14 +1550,13 @@ ApplicationWindow {
...
-        onRejectedPassphrase: {
-            if (onRejectedPassphraseCallback)
-                onRejectedPassphraseCallback();
-        }
+    }
+
+    DevicePassphraseDialog {
+        id: devicePassphraseDialog
+        visible: false
+        z: parent.z + 1
```
selsta, line 1558:
```suggestion
```

main.qml
```diff
@@ -1537,14 +1550,13 @@ ApplicationWindow {
...
+        anchors.fill: parent
```
selsta, line 1559:
```suggestion
```

src/libwalletqt/Wallet.h
```diff
@@ -348,6 +349,10 @@ class Wallet : public QObject
     Q_INVOKABLE void segregationHeight(quint64 height);
     Q_INVOKABLE void keyReuseMitigation2(bool mitigation);
 
+    // Passphrase entry for hardware wallets
+    Q_INVOKABLE void onPassphraseEntered(const QString &passphrase, bool enter_on_device, bool entry_abort=false);
+    Q_INVOKABLE virtual void onWalletPassphraseNeeded(bool on_device) override;
```
xiphon, line 354: Do you plan to invoke this from QML?

ph4r05: Currently, `onPassphraseEntered` is called by the QML, `onWalletPassphraseNeeded` is not (is called by passphrase helper from the wallet listener, causes signal emission to the QML, displays prompt). Should I remove `Q_INVOKABLE` from `onWalletPassphraseNeeded` then?

src/libwalletqt/PassphraseHelper.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    on_device = m_passphrase_on_device;
+    if (!on_device) {
+        auto tmpPass = m_passphrase.toStdString();
+        m_passphrase = QString();
+        return Monero::optional<std::string>(tmpPass);
+    } else {
+        return Monero::optional<std::string>();
+    }
```
xiphon, line 49: Looks like this should be guarded by the mutex as well, no?

src/libwalletqt/PassphraseHelper.h
```diff
@@ -0,0 +1,149 @@
...
+#endif
+QScopeGuard<F> qScopeGuard(F f)
+{
+    return QScopeGuard<F>(std::move(f));
+}
+
+QT_END_NAMESPACE
+
+#endif // QSCOPEGUARD_H
+#endif
```
xiphon, line 147: Please move `QScopeGuard` stuff to a separate header in `src/qt`

src/libwalletqt/Wallet.cpp
```diff
@@ -1021,6 +1037,20 @@ void Wallet::keyReuseMitigation2(bool mitigation)
...
 }
 
+void Wallet::onWalletPassphraseNeeded(bool on_device)
+{
+    emit this->walletPassphraseNeeded(on_device);
+}
+
+void Wallet::onPassphraseEntered(const QString &passphrase, bool enter_on_device, bool entry_abort)
+{
+    PassphraseDelegate * pdelegate = dynamic_cast<PassphraseDelegate*>(m_walletListener);
```
xiphon, line 1047: `dynamic_cast` is not needed here.

src/libwalletqt/PassphraseHelper.cpp
ph4r05: After the `onWalletPassphraseNeeded` returns the passphrase entry is finished and we just return the value and set `on_device`, I think the mutex is not needed. 
I use mutex in the PassphraseHelper just to guard state when using `QWaitCondition` to make sure it works properly.

src/libwalletqt/PassphraseHelper.h
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Implements receiver of the passphrase responsible for passing it back to the wallet,
+ * typically wallet listener.
+ */
+class PassphraseDelegate {
+public:
+    virtual void onPassphraseEntered(const QString &passphrase, bool enter_on_device, bool entry_abort) = 0;
+};
+
+class PassphraseHelper {
```
xiphon, line 57: Shouldn't it inherit `PassphraseHandler` and  `PassphraseDelegate`?

src/libwalletqt/Wallet.cpp
ph4r05: without it I cannot use the `PassphraseDelegate`: `Monero::WalletListener *m_walletListener;` 

The `m_walletListener` is instantiated with `class WalletListenerImpl : public  Monero::WalletListener, public PassphraseDelegate` but without dynamic cast I think I cannot reach the `PassphraseDelegate`

src/libwalletqt/PassphraseHelper.h
ph4r05: PassphraseHelper is only a simple class wrapping mutex/condition variable logic to wait until user enters the passphrase (as passphrase entry is blocking in the Monero wallet API).

the `PassphraseHandler` is then implemented by `{Wallet, WalletManager}` as they are responsible for showing passphrase prompt to the user, while

the `PassphraseDelegate` is implemented by Monero wallet listener as it is responsible for receiving passphrase from the user and passing it back to the monero api.

src/libwalletqt/Wallet.cpp
xiphon: Do you see reason to not make it `WalletListenerImpl *m_walletListener;`?

src/libwalletqt/PassphraseHelper.h
ph4r05: Maybe we can rename the interfaces `PassphraseHandler -> PassprasePrompter` and `PassphraseDelegate -> PassphraseReceiver`, is it more understandable?

src/libwalletqt/PassphraseHelper.cpp
xiphon: `onDevicePassphraseRequest` and `onPassphraseEntered` are concurrent. Have a good chance to corrupt the state then.

src/libwalletqt/Wallet.h
xiphon: Yep, would prefer removing `Q_INVOKABLE ` from `onWalletPassphraseNeeded` if we don't call it from QML.

src/libwalletqt/PassphraseHelper.h
xiphon: > `PassphraseHandler -> PassprasePrompter` and `PassphraseDelegate -> PassphraseReceiver`, is it more understandable?

Sounds good

src/libwalletqt/Wallet.cpp
ph4r05: We would have to move `WalletListenerImpl` out of the `Wallet.cpp`, probably to a separate `.h, .cpp` files.

src/libwalletqt/PassphraseHelper.cpp
ph4r05: I think I still don't see the scenario that could cause the corruption. 
The call flow goes like 

- `thread-wallet-api`: `onDevicePassphraseRequest` -> `onWalletPassphraseNeeded` mutex & cond wait for UI, wait
- `thread-gui`: UI calls `onPassphraseEntered`, fills in data, wakes condition variable which causes unblocking of the `thread-wallet-api`
- `thread-wallet-api`: continues after line `onWalletPassphraseNeeded`, returns eventually. 

For a race to happen GUI would have to submit another `onPassphraseEntered` somehow out of order. 

To be extra safe we could add 

```cpp
    m_mutex_pass.lock();
    auto cleanup = qScopeGuard([&] { m_mutex_pass.unlock; });
    if (m_passphrase_abort)
    {
        throw std::runtime_error("Passphrase entry abort");
    }

    on_device = m_passphrase_on_device;
    if (!on_device) {
        auto tmpPass = m_passphrase.toStdString();
        m_passphrase = QString();
        return Monero::optional<std::string>(tmpPass);
    } else {
        return Monero::optional<std::string>();
    }
```

What do you think?

src/libwalletqt/Wallet.cpp
xiphon: That's okay

src/libwalletqt/PassphraseHelper.cpp
```diff
@@ -0,0 +1,78 @@
...
+    qDebug() << __FUNCTION__;
+    m_mutex_pass.lock();
+    m_passphrase_on_device = true;
+    m_passphrase_abort = false;
+
+    if (m_handler != nullptr){
+        m_handler->onWalletPassphraseNeeded(on_device);
+    }
+
+    qDebug() << __FUNCTION__ << " entered";
```
xiphon, line 63: Can't find any benefit of having this debug message.

src/libwalletqt/PassphraseHelper.cpp
xiphon: Looks better, I would use `QMutexLocker` instead of `qScopeGuard` though.

What do you think about dropping `onWalletPassphraseNeeded` (moving it to `onDevicePassphraseRequest`)?

ph4r05: Done, pls check

src/libwalletqt/Wallet.h
ph4r05: Done, pls check

src/libwalletqt/PassphraseHelper.h
ph4r05: Done, pls check and resolve when ok.

src/libwalletqt/Wallet.cpp
ph4r05: Done, pls check and resolve when ok.

src/libwalletqt/PassphraseHelper.h
ph4r05: Done, pls check and resolve when ok.

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,106 @@
...
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import QtQuick 2.9
+import "." as MoneroComponents
+
+Item {
+    id: root
+    anchors.fill: parent
+    z: parent.z + 1
```
xiphon, line 35:
```suggestion
```

src/libwalletqt/PassphraseHelper.cpp
```diff
@@ -0,0 +1,71 @@
...
+    QMutexLocker locker(&m_mutex_pass);
+    m_passphrase_on_device = true;
+    m_passphrase_abort = false;
+
+    if (m_prompter != nullptr){
+        m_prompter->onWalletPassphraseNeeded(on_device);
+    }
+
+    m_cond_pass.wait(&m_mutex_pass);
+    locker.relock();
```
xiphon, line 45: Is not needed

src/qt/ScopeGuard.h
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2014-2020, The Monero Project
```
xiphon, line 1: We are not the authors of original code from Qt

src/libwalletqt/WalletManager.h
```diff
@@ -185,14 +186,14 @@ class WalletManager : public QObject
     // clear/rename wallet cache
     Q_INVOKABLE bool clearWalletCache(const QString &fileName) const;
 
-    Q_INVOKABLE void onWalletPassphraseNeeded(Monero::Wallet * wallet);
-    Q_INVOKABLE void onPassphraseEntered(const QString &passphrase, bool entry_abort=false);
+    Q_INVOKABLE void onPassphraseEntered(const QString &passphrase, bool enter_on_device, bool entry_abort=false);
+    Q_INVOKABLE virtual void onWalletPassphraseNeeded(bool on_device) override;
```
xiphon, line 190: This is not getting invoked from QML too

src/libwalletqt/Wallet.h
xiphon: Looks okay to me.
Only the PR author can `Resolve` it.

src/libwalletqt/PassphraseHelper.cpp
xiphon: Seems fine to me now

src/libwalletqt/PassphraseHelper.h
xiphon: Seems fine to me now

src/libwalletqt/WalletManager.h
ph4r05: done

src/qt/ScopeGuard.h
ph4r05: done

src/libwalletqt/PassphraseHelper.cpp
ph4r05: done

components/DevicePassphraseDialog.qml
ph4r05: done

src/qt/ScopeGuard.h
```diff
@@ -0,0 +1,77 @@
+#ifndef MONERO_GUI_SCOPEGUARD_H
```
selsta, line 1: Now that xiphon mentioned it, might be a good idea to add following to the top:

```
Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com, author Sérgio Martins <sergio.martins@kdab.com>
Contact: https://www.qt.io/licensing/
```

src/libwalletqt/WalletManager.cpp
```diff
@@ -539,22 +534,15 @@ WalletManager::~WalletManager()
...
 {
-    m_mutex_pass.lock();
-    m_passphrase = passphrase;
-    m_passphrase_abort = entry_abort;
-
-    m_cond_pass.wakeAll();
-    m_mutex_pass.unlock();
+    if (m_passphraseReceiver != nullptr)
+    {
+        m_passphraseReceiver->onPassphraseEntered(passphrase, enter_on_device, entry_abort);
```
xiphon, line 546: Should protect `m_passphraseReceiver` with a mutex.

src/qt/ScopeGuard.h
xiphon: I would say you should copy the whole license header from the original Qt code.

src/libwalletqt/PassphraseHelper.h
xiphon: Seems fine to me now. Feel free to `Resolve` it.

src/qt/ScopeGuard.h
ph4r05: Heh this becomes a problem now. The license is https://www.gnu.org/licenses/lgpl-3.0.html
Is it compatible with Monero-GUI?

selsta:
```
/****************************************************************************
**
** Copyright (C) 2018 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com, author Sérgio Martins <sergio.martins@kdab.com>
** Contact: https://www.qt.io/licensing/
**
** This file is part of the QtCore module of the Qt Toolkit.
**
** $QT_BEGIN_LICENSE:LGPL$
** Commercial License Usage
** Licensees holding valid commercial Qt licenses may use this file in
** accordance with the commercial license agreement provided with the
** Software or, alternatively, in accordance with the terms contained in
** a written agreement between you and The Qt Company. For licensing terms
** and conditions see https://www.qt.io/terms-conditions. For further
** information use the contact form at https://www.qt.io/contact-us.
**
** GNU Lesser General Public License Usage
** Alternatively, this file may be used under the terms of the GNU Lesser
** General Public License version 3 as published by the Free Software
** Foundation and appearing in the file LICENSE.LGPL3 included in the
** packaging of this file. Please review the following information to
** ensure the GNU Lesser General Public License version 3 requirements
** will be met: https://www.gnu.org/licenses/lgpl-3.0.html.
**
** GNU General Public License Usage
** Alternatively, this file may be used under the terms of the GNU
** General Public License version 2.0 or (at your option) the GNU General
** Public license version 3 or any later version approved by the KDE Free
** Qt Foundation. The licenses are as published by the Free Software
** Foundation and appearing in the file LICENSE.GPL2 and LICENSE.GPL3
** included in the packaging of this file. Please review the following
** information to ensure the GNU General Public License requirements will
** be met: https://www.gnu.org/licenses/gpl-2.0.html and
** https://www.gnu.org/licenses/gpl-3.0.html.
**
** $QT_END_LICENSE$
**
****************************************************************************/
```

appears to be the full license

https://github.com/RSATom/Qt/blob/master/qtbase/src/corelib/tools/qscopeguard.h

src/libwalletqt/WalletManager.cpp
ph4r05: That it does not become null after the check?
It is set only in `WalletManager::openWallet` and `WalletManager::createWalletFromDevice`, both are protected by locking `m_mutex`. 

The scenario this could happen is: GUI somehow triggers (without previous request) `onPassphraseEntered`, meanwhile one of the two functions ends, scope guard sets `m_passphraseReceiver` to null in race. 

Should we then to protect from this add a new mutex, that is called from the scope guard, before setting `m_passphraseReceiver` to null?

components/DevicePassphraseDialog.qml
```diff
@@ -0,0 +1,104 @@
...
+    property var onWalletEntryCallback
+    property var onRejectedCallback
+
+    function open(canEnterOnDevice_) {
+        var canEnterOnDevice = canEnterOnDevice_ !== null ? canEnterOnDevice_ : canEnterOnDevice
+        root.visible = true;
+
+        if (canEnterOnDevice) {
+            entryChooserDialog.okText = qsTr("Hardware wallet")
+            entryChooserDialog.cancelText = qsTr("Host")
```
xiphon, line 45: Could we replace `Host` with something GUI users more familiar with?

ph4r05: sure, I will set there any string you suggest

src/qt/ScopeGuard.h
selsta: looks ok?

https://opensource.stackexchange.com/questions/6405/lgpl-licensed-code-reuse-in-mit-licensed-project

src/libwalletqt/WalletManager.cpp
xiphon: > Should we then to protect from this add a new mutex, that is called from the scope guard, before setting `m_passphraseReceiver` to null?

Yes, please do

src/qt/ScopeGuard.h
ph4r05: This seems problematic at this point. We like to have something like scope guard to have RAII resource management, but without upgrading the Qt I think we cannot use this source without forcing the whole Monero-GUI to be LGPLv3, am I right?

Do we need to reimplement scope guard by ourselves? And if it is the case, it probably cannot be called `qScopeGuard` anymore as it won't have the same properties. 
Then we need to have something ours, which fallbacks to `qScopeGuard` if it is present. Am I right?

xiphon: > Heh this becomes a problem now. The license is https://www.gnu.org/licenses/lgpl-3.0.html
> Is it compatible with Monero-GUI?

At the first glance, we ain't doing anything malicious, nor we violate the license.

We don't claim the copyright on the code and we copy the original license and copyright.

selsta: There is also this: https://spdx.org/licenses/Qt-LGPL-exception-1.1.html

components/DevicePassphraseDialog.qml
xiphon: Let me summon @erciccione

src/qt/ScopeGuard.h
ph4r05: I am removing this scope guard and using https://github.com/ricab/scope_guard

ph4r05: Any ideas where to put the scope guard related files? it does not seem right to put it in `src/qt`, should I create `src/misc` ?

selsta: FWIW we already have a Qt file in the source code tree (MoneroSettings) so if we violate the license we did for a while.

components/DevicePassphraseDialog.qml
erciccione: I think `Computer` would be better. Sounds dumb, but it's very easy to understand for both users and translators.

src/qt/ScopeGuard.h
xiphon: > I am removing this scope guard and using https://github.com/ricab/scope_guard

Great, that is a good one.

>Any ideas where to put the scope guard related files? it does not seem right to put it in `src/qt`, should I create `src/misc` ?

That's correct. But we already have some unrelated things in `src/qt`, please place it there for now.

ph4r05: Hmm I still think that having LGPL sources in the app is problematic. This LGPL has to be stated somewhere in the app (about), the app is distributed also in the binary forms. I am not sure whether this is safe. LGPL states 

> A program that contains no derivative of any portion of the Library, but is designed to work with the Library by being compiled or linked with it, is called a "work that uses the Library". Such a work, in isolation, is not a derivative work of the Library, and therefore falls outside the scope of this License.

I think that means that whole Monero-GUI would have to be LGPL when we use the source file directly. https://www.quora.com/If-my-code-uses-some-code-licensed-under-LGPL-can-I-release-the-whole-software-under-MIT

ph4r05: So what do you think guys? I would strongly prefer to use Unlicensed https://github.com/ricab/scope_guard to avoid such license pains. Do you agree?

If yes, should I create `src/misc` or `src/utils` where to put these new files or should I use some existing dir?

xiphon: > I would strongly prefer to use Unlicensed https://github.com/ricab/scope_guard to avoid such license pains. Do you agree?

Yes.

Already replied to this, see https://github.com/monero-project/monero-gui/pull/2829#discussion_r422211422

ph4r05: > Already replied to this, see [#2829 (comment)](https://github.com/monero-project/monero-gui/pull/2829#discussion_r422211422)

Ah I see, GitHub didn't refresh the conversation.

src/libwalletqt/Wallet.cpp
xiphon: Looks okay to me

src/libwalletqt/WalletManager.cpp
ph4r05: ok done, pls check

src/qt/ScopeGuard.h
ph4r05: ok done, pls check

src/qt/ScopeGuard.h
xiphon: Hm. I still think we shouldn't put Monero copyright here.

src/qt/ScopeGuard.h
```diff
@@ -0,0 +1,206 @@
...
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
xiphon, line 27: Hm. I still think we shouldn't put Monero copyright here.

ph4r05: Its matter of preference I think, It is licensed as "Unlicense" which means we can freely relicense it to a common Monero license. We also mention the author. But I basically don't care, do you want to put there just author name as it was there previously? Or the unlicense text?

src/libwalletqt/WalletManager.cpp
```diff
@@ -186,6 +182,11 @@ Wallet *WalletManager::createWalletFromDevice(const QString &path, const QString
 {
     QMutexLocker locker(&m_mutex);
     WalletPassphraseListenerImpl tmpListener(this);
+    m_passphraseReceiver = &tmpListener;
```
xiphon, line 185: Please guard the assignment the `m_mutex_passphraseReceiver` too

ph4r05: ah, true, done

src/qt/ScopeGuard.h
xiphon: Would prefer having author name, github link and Unlicense text here

ph4r05: ok done

src/libwalletqt/WalletManager.cpp
```diff
@@ -532,29 +537,26 @@ WalletManager::WalletManager(QObject *parent)
     , m_scheduler(this)
 {
     m_pimpl =  Monero::WalletManagerFactory::getWalletManager();
+    m_mutex_passphraseReceiver.lock();
```
xiphon, line 540: The mutex is not needed here. Initializing `m_passphraseReceiver(nullptr)` will work just fine

src/libwalletqt/WalletManager.cpp
xiphon: Is okay now

src/libwalletqt/WalletManager.cpp
xiphon: Is okay now

ph4r05:
Thanks @xiphon !

xiphon:
> Thanks @xiphon !

No problem. Thank you.

PR #2831, opened 2020-04-14
Transfer: add Address Book inline button in Address field

rating89us:
Changes:
- Remove "(Address Book)" link from address field label
- Add address book inline button with icon + adjust right margin of address field
- Add second inline button (scan QR code button)
- Change color of scan QR code button

Address book inline button as single inline button:
inlineaddressbutton3

With scan QR code button enabled:
inlinewithqrcode

### DIFF
--- a/components/InlineButton.qml
+++ b/components/InlineButton.qml
@@ -53,6 +53,7 @@ Item {
     property alias fontPixelSize: inlineText.font.pixelSize
     property alias fontFamily: inlineText.font.family
     property alias buttonColor: rect.color
+    property alias buttonHeight: rect.height
     signal clicked()
 
     function doClick() {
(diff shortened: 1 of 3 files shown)


xiphon:
And another not mentioned change is removing "Scan QR code" inline button

rating89us:
I thought the scan QR code button was disabled so I commented those lines. Do you think I should implement this address book button as a 2nd inline button and keep the scan QR code button?

selsta:
You can still compile with QR scanner enabled.

rating89us:
Updated

rating89us:
In my opinion scan QR code button should be always visible (when enabled). Currently it is hidden when the address field isn't empty. What do you think?

xiphon:
> In my opinion scan QR code button should be always visible (when enabled).

Yep, agree

selsta:
What is the opinion on adding a `Scan QR code` menu item instead of a second inline button?

Screenshot 2020-04-16 at 03 17 24

xiphon:
Think nobody will ever know the feature exists then.

PR #2832, opened 2020-04-14
updater: fetch signed hashes from getmonero.org, verify downloads

xiphon:
monero-gui-update-download 1
monero-gui-update-download 12
monero-gui-update-download 92

Related: https://github.com/monero-project/monero-gui/pull/2824 https://github.com/monero-project/monero-gui/pull/2819


src/qt/updater.cpp
```diff
@@ -39,37 +40,73 @@ Updater::Updater()
     m_maintainers.emplace_back(fileGetContents(":/monero/utils/gpg_keys/luigi1111.asc").toStdString());
 }
 
-QPair<QString, QString> Updater::verifySignaturesAndHashSum(
+QByteArray Updater::fetchSignedHash(
+    const QString &binaryFilename,
+    const QByteArray &hashFromDns,
+    QPair<QString, QString> &signers) const
+{
+    static constexpr const char hashesTxtUrl[] = "https://web.getmonero.org/downloads/hashes.txt";
```
erciccione, line 48: Wouldn't be better to fetch the hashes from the repository instead? https://raw.githubusercontent.com/monero-project/monero-site/master/downloads/hashes.txt.

It's easier to spot a compromised file on GitHub. Maybe github could be the primary source and getmonero a fallback option in case github is down?

selsta: We require the signature of two different maintainers, it does not matter if getmonero.org would be hacked.

xiphon: This and also:

users doing manual update will use getmonero.org.

If getmonero.org gets hacked, there will be appropriate errors in the GUI console/logs, so we will react way faster on a hack.

If we use github as a source, we lose this opportunity.

PR #2837, opened 2020-04-16
Save wallet on money received

tobtoht:
Displaying an accurate balance is fundamental to the concept of a wallet, and it is therefore imperative that the balance shown when the wallet is exited is always the same amount that is shown when the wallet is opened.

I've about had it with helping people that write anxiety ridden reports about "losing" their balance and fearing that it got stolen / lost / disappeared into the void because they forcefully closed their wallet / it crashed before it got saved.

I've tried to argue for wallet consistency (for subaddresses and accounts) before in #2639, but balance is more important.

The wallet is already saved after sending a transaction, but not after receiving one. This PR saves the wallet whenever money is received.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -735,6 +735,8 @@ ApplicationWindow {
 
         if(middlePanel.state == "History")
             middlePanel.historyView.update();
+
+        currentWallet.store();
     }
 
     function onWalletUnconfirmedMoneyReceived(txId, amount) {


rating89us:
Why not save the wallet when transaction isn't confirmed yet?

xiphon:
The patch opens the door to DoS-ing the wallet (intentionally or unintentionally).

xiphon:
`Autosave every N minutes` option would work way better.

tobtoht:
>The patch opens the door to DoS-ing the wallet

You're right, I had not considered that.

>Autosave every N minutes option would work way better.

What would an acceptable interval be?

xiphon:
> > Autosave every N minutes option would work way better.
> 
> What would an acceptable interval be?

Thinking about 10 or 15 mins by default. Would be nice to have a slider like `Lock wallet on inactivity`.
Let me know if you want to implement this, but if you don't have enough free time, i'm up to work on it.

rating89us:
It could save when it receives the first transaction, and from then on every 10-15 min. to avoid DoS attacks.

xiphon:
Implemented autosave https://github.com/monero-project/monero-gui/pull/2879

tobtoht:
Thanks.

PR #2846, opened 2020-04-22
Merchant: various bug fixes / small improvements

selsta:
- Confirmations would not show with untrusted remote nodes
- tx_pool check was buggy
- payment URL did not display
- Point out when amount is unconfirmed

### DIFF
--- a/pages/merchant/Merchant.qml
+++ b/pages/merchant/Merchant.qml
@@ -33,6 +33,9 @@ Item {
     property var    hiddenAmounts: []
 
     function onPageCompleted() {
+        if (appWindow.currentWallet) {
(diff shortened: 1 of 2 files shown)


pages/merchant/Merchant.qml
```diff
@@ -616,15 +615,12 @@ Item {
 
                 var in_txpool = false;
                 var confirmations = 0;
-                var displayAmount = 0;
+                var displayAmount = displayAmount = model.data(idx, TransactionHistoryModel.TransactionDisplayAmountRole);
```
xiphon, line 618: .
```suggestion
                var displayAmount = model.data(idx, TransactionHistoryModel.TransactionDisplayAmountRole);
```

selsta: fixed

PR #2848, opened 2020-04-22
LanguageSidebar: reenable

### LINKED ISSUE #2525: Language settings page isn't scrollable
Some languages aren't displayed when window size is small.
image

selsta:
Any reason why we shouldn’t do this? It was disabled due to Qt 5.7 but now that we use Qt 5.9 it can be enabled again.

Resolves #2525


rating89us:
When the wizard is open, I prefer the way it is now (toolbar).
When the wallet is open, most users won't change the language, so it's better to have a dropdown menu in Settings > Layout.

selsta:
Not sure if I understand your comment. This PR does not touch toolbar.

PR #2850, opened 2020-04-23
checkUpates: installer support (Win), use GUI buildTag and version

xiphon:


src/main/oshelper.cpp
```diff
@@ -164,3 +164,47 @@ QString OSHelper::temporaryPath() const
...
+        &installLocation[0],
+        &size);
+    if (status != ERROR_SUCCESS)
+    {
+        qCritical() << "RegGetValueW Failed (read)" << status;
+        return false;
+    }
+
+    const QDir installDir(QString(reinterpret_cast<const QChar *>(&installLocation[0])));
+    return installDir == QDir(QCoreApplication::applicationDirPath());
```
selsta, line 206:
```
#include <QCoreApplication>
```

If I understand the CI error right.

xiphon: Yes, forgot to stage and commit this line

PR #2855, opened 2020-04-24
main: add --disable-check-updates flag

selsta:
Checkbox in Settings -> Interface will be added in a separate PR.


main.qml
```diff
@@ -1989,7 +1989,9 @@ ApplicationWindow {
     }
 
     function checkUpdates() {
-        walletManager.checkUpdatesAsync("monero-gui", "gui")
+        if (!disableCheckUpdatesFlag) {
```
selsta, line 1992: I could also additionally disable `updatesTimer`.

xiphon: I would omit this change, drop `checkUpdates` from `onCompleted` and use `triggeredOnStart` and `running` `updatesTimer`'s properties instead

selsta: Updated.

PR #2866, opened 2020-04-26
ProcessingSplash: new design

selsta:
Screenshot 2020-04-27 at 01 36 26
Screenshot 2020-04-27 at 01 33 21


components/ProcessingSplash.qml
```diff
@@ -56,44 +58,57 @@ Rectangle {
...
         }
 
 
         MoneroComponents.TextPlain {
-            id: heightProgress
-            font {
-                pixelSize: 18
-            }
+            id: messageTitle
+            text: "Please wait..."
```
xiphon, line 105: Translation

components/Style.qml
```diff
@@ -58,6 +58,7 @@ QtObject {
     property string titleBarCloseSource: blackTheme ? _b_titleBarCloseSource : _w_titleBarCloseSource
     property string titleBarButtonHoverColor: blackTheme ? _b_titleBarButtonHoverColor : _w_titleBarButtonHoverColor
 
+    property string moneroLogoSource: "qrc:///images/monero-vector.svg"
```
xiphon, line 61: ?

components/ProcessingSplash.qml
selsta: updated

components/Style.qml
selsta: updated

wasn’t sure where to add it, I skipped an extra Style.qml entry now

components/ProcessingSplash.qml
```diff
@@ -29,21 +29,23 @@
...
 import QtQuick.Layouts 1.1
 
 import "../components" as MoneroComponents
 
 Rectangle {
     id: root
-    color: MoneroComponents.Style.blackTheme ? "white" : "transparent"
+    color: MoneroComponents.Style.blackTheme ? "black" : "white"
     visible: false
+    radius: 10
```
xiphon, line 41: Looks inconsistent with the other dialogs

selsta: Plan is to update the other dialogs too over time:

https://www.figma.com/file/DplJ2DDQfIKiuRvolHX2hN/Monero-GUI?node-id=2%3A4925

xiphon: Ok, let's use radius.

selsta:
@GBKS Can you upload the spinner from Figma as an .svg?

GBKS:
SVG can't do conic gradients unfortunately, would a PNG do as well? I'm attaching 1x and 2x versions here. 
load-spinner
load-spinner@2x

selsta:
@GBKS Could you upload the 1x version with 80x80 pixels (and the corresponding 2x)? Thanks!

selsta:
@GBKS The 66x66 current size is okay! No need for a new upload. Thanks again.

selsta:
@GBKS New problem now, can you check if the busy indicator is a proper circle? It does not seems so.
Screenshot 2020-04-29 at 09 39 59
The red line is taken from above, it should be the same size as the top margin but the bottom margin between logo and busy indicator is smaller.

GBKS:
Good catch. The width was off by one pixel. Here are the updated files.

load-spinner
load-spinner@2x

selsta:
@GBKS thanks, the new ones are centred now.

Updated.

GBKS:
Woohoo!

Issue #2700, opened 2020-01-06
Password fields missing unmask password icon


rating89us:
The unmask password icon (eye icon)...
![image](https://user-images.githubusercontent.com/45968869/71831832-75e74e00-3088-11ea-86ea-63866c06cbcc.png)

...is missing in some password fields:
![image](https://user-images.githubusercontent.com/45968869/71832011-c9f23280-3088-11ea-9600-3b1ccc51d42c.png)

![image](https://user-images.githubusercontent.com/45968869/71831877-8b5c7800-3088-11ea-8b03-522cabc87479.png)

PR #2870, opened 2020-04-28
LineEdit: password mode, linking. PasswordDialog: use LineEdit

xiphon:
Related: https://github.com/monero-project/monero-gui/issues/2700
monero-gui-password

### DIFF
--- a/components/LineEdit.qml
+++ b/components/LineEdit.qml
@@ -26,6 +26,7 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import FontAwesome 1.0
 import QtQuick 2.9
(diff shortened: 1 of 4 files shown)


components/LineEdit.qml
```diff
@@ -210,6 +239,27 @@ Item {
             onTextChanged: item.textUpdated()
             topPadding: 10
             bottomPadding: 10
+            echoMode: isPasswordHidden() ? TextInput.Password : TextInput.Normal
+
+            MoneroComponents.Label {
+                visible: password || passwordLinked
+                fontSize: eyeMouseArea.containsMouse ? 24 : 20
```
selsta, line 246:
```suggestion
                fontSize: eyeMouseArea.containsMouse ? 22 : 20
```

24 feels a bit too much

xiphon: Is a copy-paste of the current code.

TBH, I don't get why it gets bigger on hover. Would prefer to drop this. Thoughts?

selsta: Yes, I prefer dropping this and only changing color.

xiphon: Updated

PR #2871, opened 2020-04-28
WizardAskPassword: drop custom password fields, use LineEdit

xiphon:
Fixes https://github.com/monero-project/monero-gui/issues/2700
Requires https://github.com/monero-project/monero-gui/pull/2870.

monero-gui-password 0


xiphon:
Rebased

PR #2872, opened 2020-04-28
Remove caps from multiple strings

rating89us:
Changes: 
- Remove unnecessary caps in multiple strings
- Rename `IP` to `IP address` in Settings > Node page


selsta:
That would mean it has to get retranslated to all languages. Not sure if it’s worth it to change the capitalization of 2 words.

rating89us:
I think it's worth it. There are only 3 strings with wrong caps remaining in the GUI. We won't need to do this again.

erciccione:
I don't think it worth retranslating more than 150 strings (5 words for +30 languages) because the capitalization of some words is off. Probably better to keep them as they are.

rating89us:
So we will never have a perfect GUI, because there will always be the fear of having to retranslate some strings. Also, there strings are definitely not critical for 99% of users, therefore it won't be a disaster if they stay untranslated for some time.

erciccione:
@rating89us Please understand that while i share your wish to have a perfect and professional looking GUI, you have to consider that what might look like a 5 seconds change, for others can be hours of work.

QT is not smart and if something in a string changes, the entire string will be marked "untranslated". That means that somebody (guess who) will have to go check all +150 strings and remove the "untranslated" flags one by one and if a string is already translated, a translator will have to check it and confirm it's valid.

Now, there have been a lot of string changes lately and translators will have to work on a lot of stuff. If you consider that i'm pretty sure they won't have a lot of time to translate the changed/new strings, you understand that adding 150 strings to the count is not helpful.

If we had a crazy amount of translators that wouldn't be a problem, but since probably many strings or entire languages will not receive updates (because we don't have translators for those languages, at least for now) you have to consider that some strings will simply stay untranslated and manual work will be needed to fix them.

My question is: Does this change justify all the downsides listed above? I think it doesn't, but i can understand why people might want it anyway. Let's just be real, there will be always a comma off, a dot too much or a better phrasing :)

rating89us:
> Let's just be real, there will be always a comma off, a dot too much or a better phrasing :)

Phishing emails and other scam messages often contain capitalization and grammatical errors. These type of errors are red flags. We should always strive for excellence. 

> will have to go check all +150 strings and remove the "untranslated" flags one by one

You don't have to check all strings one by one. Let the translators do it, it's only 5 strings for each translator.

> My question is: Does this change justify all the downsides listed above? 

In my opinion, yes, it does.

erciccione:
> You don't have to check all strings one by one. Let the translators do it, it's only 5 strings for each translator.

We don't have translators for every language and that's not how Weblate works.

Let's merge this after next release so everyone is happy.

selsta:
Okay, let’s merge it after the upcoming release.

PR #2876, opened 2020-05-01
StandardDropdown: drop Qt 5.8 workaround

xiphon:


pages/settings/SettingsLayout.qml
```diff
@@ -304,10 +305,6 @@ Rectangle {
             i += 1;
         }
 
-        fiatPriceProviderDropDown.update();
```
selsta, line 307: Screenshot 2020-05-01 at 14 37 28

xiphon: Can't reproduce. Do you see any errors in the logs? Could you try clean build?
Does the dropdown contain any options? What happens if you select one and restart the GUI?

selsta: > Do you see any errors in the logs?

No

>  Could you try clean build?

Same issue

> Does the dropdown contain any options?

Yes, the dropdown looks correct

> What happens if you select one and restart the GUI?

Same issue.

selsta: Note that the other dropdown menus are correct.

xiphon: Fixed. Thanks for catching this.

Issue #2888, opened 2020-05-06
Simple mode stop daemon loop


selsta:
Steps to reproduce:

- Have the wallet in simple mode
- Start wallet
- Enter a wrong password
- Stop daemon popup starts to loop

![loop](https://user-images.githubusercontent.com/7697454/81160184-0deb3a80-8f8a-11ea-875f-a23a92d7cacc.gif)

Can reproduce with and without #2884.

PR #2891, opened 2020-05-07
SimpleMode: use connection timer 'running' prop, fix daemon restart

xiphon:
Closes https://github.com/monero-project/monero-gui/issues/2888


LuckyProot:
🤖✨✨✨✨✨✨🤔

PR #2893, opened 2020-05-09
SettingsLayout: use minute(s) to simplify for translation

rating89us:


pages/settings/SettingsLayout.qml
```diff
@@ -109,7 +109,10 @@ Rectangle {
             stepSize: 1
             to: 60
             value: persistentSettings.autosaveMinutes
-            text: "%1 %2 %3".arg(qsTr("Every")).arg(value).arg(qsTr("minute(s)")) + translationManager.emptyString
+            text: {
+                var minutes = value > 1 ? qsTr("minutes") : qsTr("minute");
+                return qsTr("Every ") + value + " " + minutes + translationManager.emptyString;
```
selsta, line 114:
```suggestion
                return "%1 %2 %3".arg(qsTr("Every")).arg(value).arg(minutes) + translationManager.emptyString;
```

Else translators have to retranslate `"Every"`.

xiphon (changes requested):
I don't like the patch. It only bloats the code and adds more work to translators.

And i guess some languages have more complex rules.

@erciccione

erciccione (changes requested):
Singulars / plurals are always hard to deal with, especially because of the little context translators can extrapolate. This string is a perfect example. I think would be better to just leave this string as it is, to give translators more flexibility.

selsta:
`qsTr("Every %n second(s)", "0", value)` solves the plural problem. See #2915

selsta:
Superseded by #4591

PR #2898, opened 2020-05-11
SettingsNode: dropdown (stop/keep local node when closing GUI)

rating89us:
Changes: 

StandardDialog:
- changed to new design schema (black background, rounded borders)
- cancel button is secondary (grey color)

SettingsNode:
- added dropdown to select whether to stop or keep local node running when closing GUI

main.qml:
- the dialog is not full screen anymore.
- text with better explanation


xiphon:
Please split into separate PRs

PR #2771, opened 2020-02-11
WizardRestoreWallet1: seed - strip whitespaces and line breaks

xiphon:

PR #2899, opened 2020-05-11
Filter out multiple blanks in seed input when counting words

rbrunner7:
This uses a regular expression for the JavaScript method `split` instead of simply `" "` to deal with any multiple blanks in seed input, without any undesirable side effects like it seems this earlier attempt to solve the same problem had: #2771

### DIFF
--- a/js/Wizard.js
+++ b/js/Wizard.js
@@ -97,10 +97,6 @@ function tr(text) {
     return qsTr(text) + translationManager.emptyString
 }
 
-function lineBreaksToSpaces(text) {
-    return text.trim().replace(/(\r\n|\n|\r)/gm, " ");
-}
-
 function usefulName(path) {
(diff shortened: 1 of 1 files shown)


js/Wizard.js
```diff
@@ -110,7 +110,7 @@ function usefulName(path) {
 
 function checkSeed(seed) {
     console.log("Checking seed")
-    var wordsArray = lineBreaksToSpaces(seed).split(" ");
+    var wordsArray = lineBreaksToSpaces(seed).split(/\s+/);
```
xiphon, line 113: Consider to drop `lineBreaksToSpaces` entirely

rbrunner7: You are right of course, regular expression `s` alone already nicely covers any kind of whitespace, not only blanks. PR is updated.

PR #2895, opened 2020-05-10
Transfer: new TxConfirmationDialog

rating89us:
Closes #2467

**Changes:**

New TxConfirmationDialog component:
- From field: 
  - Wallet name ("Ledger", "Trezor" or "My wallet")
  - Account # + Account label (if the wallet has more than one account)
- To field:
  - Address book icon + name (if address is in address book)
  - Destination address
- Added fiat conversion to amount and fee
- Removed extra zeros from the end of amount and fee
- Keyboard can be used to cancel (ESC key) or to confirm dialogs (enter/return keys)
- When the dialog is waiting for user action (hardware wallet button press), a blinking text message will appear

Normal transaction:
image

PR #2903, opened 2020-05-13
Wallet: add isTrezor() function

rating89us:
We already have `isLedger()` and `isHwBacked()` boolean functions.

### DIFF
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -239,6 +239,11 @@ bool Wallet::isLedger() const
     return m_walletImpl->getDeviceType() == Monero::Wallet::Device_Ledger;
 }
 
+bool Wallet::isTrezor() const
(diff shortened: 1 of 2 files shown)


rating89us:
PR #2895

TxConfirmationDialog.qml line 261
```
                       if (currentWallet) {
                            var walletTitle = function() {
                                if (currentWallet.isHwBacked() === true && currentWallet.isLedger() === true) {
                                    return "Ledger";
                                } else if (currentWallet.isHwBacked() === true && currentWallet.isLedger() === false) {
                                    return "Trezor";
                                } else {
                                    return qsTr("My wallet");
                                }
                            }
```

selsta:
We might have new hw device backends in the future so I think this would be ok to add. The alternative would be an enum with `HW_NONE`, `HW_LEDGER`, `HW_TREZOR`.

PR #2904, opened 2020-05-13
StandardDialog: new design

rating89us:
New design schema according to [Figma](https://www.figma.com/file/DplJ2DDQfIKiuRvolHX2hN/Monero-GUI?node-id=2%3A4925): 
 - Black background
 - Round borders
 - Grey cancel button (secondary button)


components/StandardDialog.qml
```diff
@@ -38,8 +38,11 @@ import "effects/" as MoneroEffects
 
 Rectangle {
     id: root
-    color: "transparent"
+    color: MoneroComponents.Style.blackTheme ? "black" : "white"
     visible: false
+    radius: 10
+    border.color: MoneroComponents.Style.blackTheme ? Qt.rgba(255, 255, 255, 0.25) : Qt.rgba(0, 0, 0, 0.25)
```
xiphon, line 44: Not sure we should use transparency here (there is `appWindowBorderColor`, `grey`, etc.)

selsta: That’s what I did in 5265e52

rating89us: Changed to `MoneroComponents.Style.moneroGrey` and `MoneroComponents.Style._w_appWindowBorderColor`, which are the same colors used in Figma design schema.

components/StandardDialog.qml
```diff
@@ -38,8 +38,11 @@ import "effects/" as MoneroEffects
 
 Rectangle {
     id: root
-    color: "transparent"
+    color: MoneroComponents.Style.blackTheme ? "black" : "white"
     visible: false
+    radius: 10
+    border.color: MoneroComponents.Style.blackTheme ? MoneroComponents.Style.moneroGrey : MoneroComponents.Style._w_appWindowBorderColor
```
xiphon, line 44: Could we use ProcessingSplash border color here as well?

xiphon:
ping @rating89us

components/StandardDialog.qml
rating89us: ProcessingSplash border color is:
`MoneroComponents.Style.blackTheme ? Qt.rgba(255, 255, 255, 0.25) : Qt.rgba(0, 0, 0, 0.25)`

It works. But it's the same as the last review, where you said
> Not sure we should use transparency here (there is `appWindowBorderColor`, `grey`, etc.)

xiphon: Gotcha

xiphon (approved):

selsta:
Not sure if I understand this. 

Is this how it is intended to look?

Screenshot 2020-09-15 at 13 25 07

PR #2914, opened 2020-05-16
Redesign wallet mode selection

rating89us:
**Change summary**:

**1) Wallet mode selection page** 
- At first, only two options appears: simple and advanced mode
- Simple mode icon is simpler than the advanced mode icon
- Added (recommended) tag to simple mode
- Description of simple mode includes only features, and doesn't explain connection options
- Included transactions history, address book and hardware wallet support in simple mode description. We could also add merchant mode and accounts, but decided not to include them.
- Advanced mode description now includes almost all advanced features available

image

**2) After selecting simple mode, connection options for simple mode are displayed**
- Added (recommended) tag to Local Node + Bootstrap node option (wallet mode 1)
- Moved wallet mode 1 to the first option, since it's the recommended one
- Added FontAwesome icons (the same used in Settings > Node page)
- Added "remote node" term in the description of both options, to make it clear that both options of simple mode connect to remote node

image

**3) Warning after selecting Local node & Bootstrap node**
- Better explanation overall
- Precise instructions for users who prioritize privacy: they should use an advanced mode and start a local node without setting a bootstrap node (previously it was only said that those users should use a "full node")

image

**4) Warning after selecting Remote node**
- Better explanation overall
- Precise instructions for users who prioritize privacy: they should use an advanced mode and start a local node (previously it was only said that those users should use a "full node")
image


wizard/WizardModeSelection.qml
```diff
@@ -65,52 +66,67 @@ Rectangle {
             WizardMenuItem {
                 opacity: appWindow.persistentSettings.nettype == 0 ? 1.0 : 0.5
                 Layout.topMargin: 20
-                headerText: qsTr("Simple mode") + translationManager.emptyString
+                headerText: qsTr("Simple mode") + (appWindow.persistentSettings.nettype == 0 ? " " + qsTr("(recommended)") : "") + translationManager.emptyString
```
erciccione, line 69: I disagree with this change. We should (and we do) recommend users to run their own node, not to rely on a third party one. I would just take off "recommended" from the simple mode.

selsta: From the screenshots simple mode (bootstrap) gets recommended which means users ending up with their own node.

rating89us: Newbie users usually just click on the reccomended options.
In my design, the user who clicks on both (recommended) options (Simple mode and Local node & Bootstrap node) in the end will have their own local node running.

rbrunner7:
I think it's certainly useful to give people more information at hand for those very first and important decision how to run their wallet.

But the added "(Recommended)" hints make me uneasy. I understand the intention, but nevertheless in my personal opinion it's just *that* little step too far for the leading private cryptocurrency to go as far as recommending its users the use of remote nodes.

I guess users might also find it a little strange that first something gets recommended and then afterwards they have to explicitely confirm to understand and accept some "implications". They may think, yeah, couldn't that clever wallet recommend something without such "implications"?

Remote nodes are of course very useful, and probably a "necessary evil" today for most new Monero users. But recommending them?

wizard/WizardModeSelection.qml
erciccione: My point is that using bootstrap mode they will transact on the network using a remote node until their local node is synced, which may never happen. I don't know if we should specifically suggest it.

rating89us:
When we discuss the recommended tag we should be clear whether it's recommended for usability (essential features x full featured wallet) or privacy (local node x remote node).

If we consider only the usability (wallet features) and ignore the privacy issue, I would always recommend simple mode to most users.

The problem is that simple mode options (wallet mode 0 and 1) always connect to a remote node. There is no option to use a simple wallet with full privacy (without connecting to a remote node).

If there were three options in simple mode, 1) Local node without Bootstrap node, 2) Local node with Bootstrap node and 3) Remote node, we would all agree that the first option is the best one to recommend for privacy reasons, but at the cost of having to wait for the blockchain to download.

rbrunner7:
> When we discuss the recommended tag we should be clear whether it's recommended for usability (essential features x full featured wallet) or privacy (local node x remote node).

Well, we here, the devs, can discuss what we want, under whatever discussion assumptions that are useful; but IMHO this does not change the fact that info about the two very different possible meanings of "recommended" will not reach those Monero beginners reading those screens.

They will only see "(Recommended)" and understand that the Monero community recommends them to choose that particular option. I for one don't think the Monero community should do this, i.e. make such a recommendation.

> The problem is that simple mode options (wallet mode 0 and 1) always connect to a remote node. There is no option to use a simple wallet with full privacy (without connecting to a remote node).

I see. Not sure whether the wallet should learn that, but anyway: The current absence of a simple mode with full privacy still does not tip the balance to the "recommending" side for me.

rating89us:
> The current absence of a simple mode with full privacy still does not tip the balance to the "recommending" side for me.

Sure, you conviced me that with the current wallet state it's better to remove the (recommended) tags. But I think we should further explore these ideas in future PRs, like allowing wallets in easy mode to start a local node without bootstrap node. For example:

Wallet modes (interface modes):
 - Easy mode: interface with essential features.
 - Advanced mode: interface with advanced features.

Connection options in easy mode:
  - Local node without Bootstrap node (slow & no privacy limitations)
  - Local node with Bootstrap node (pre-set bootstrap node) (fast & temporary privacy limitations)
  - Remote node (pre-set remote node) (fast & permanent privacy limitations)

Connection options in advanced mode:
  - Local node without Bootstrap node (slow & no privacy limitations)
  - Local node with Bootstrap node (custom bootstrap node) (fast & temporary privacy limitations)
  - Remote node (custom remote node) (fast & permanent privacy limitations)

xmrdsc:
I like this PR except for those FontAwesome icons. They don't fit the other images displayed on that page. Maybe @GBKS is willing to create 2 new icons :D!

PR #2915, opened 2020-05-16
Utils: simplify for translations

selsta:
Test commit to see how this looks in weblate.

### DIFF
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -55,36 +55,14 @@ function ago(epoch) {
     var now = new Date().getTime() / 1000;
     var delta = now - epoch;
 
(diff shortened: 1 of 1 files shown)


erciccione:
I tried this on my weblate instance after using `lupdate` on all language files. Seems to be working very good. singulars and plurals get picked up correctly by weblate:

en

More complex languages seem to be working great as well:

complex.

xiphon:
Do we have other similar places? If so, would be nice to fix them all the once.

erciccione:
@xiphon yes. See https://translate.getmonero.org/translate/monero/gui-wallet/en/?q=check:optional_plural

PR #2919, opened 2020-05-20
OpenGL: Windows - fall back to software renderer if OpenGL < 2.1

xiphon:
Report: https://www.reddit.com/r/Monero/comments/gmhxl6/im_having_trouble_running_monero_gui_wallet_on_my/

We don't need `start-low-graphics-mode.bat` anymore. Could keep it for a while and drop it later though.

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -75,6 +75,10 @@
 #include "daemon/DaemonManager.h"
 #endif
 
+#if defined(Q_OS_WIN)
+#include <QOpenGLContext>
+#endif
+
 #ifdef WITH_SCANNER
 #include "QR-Code-scanner/QrCodeScanner.h"
 #endif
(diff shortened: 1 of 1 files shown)


src/main/main.cpp
```diff
@@ -75,6 +75,11 @@
 #include "daemon/DaemonManager.h"
 #endif
 
+#if defined(Q_OS_WIN)
+#include <QOpenGLContext>
+#include <QOpenGLFunctions>
```
selsta, line 80: Is this include required?

xiphon: Yes. That's why i put it there.

xiphon: Ah, sorry. that was from another version of the patch. Can drop it now.

xiphon: Updated. Thanks

PR #2922, opened 2020-05-25
Wizard: redesign daemon settings page

rating89us:


selsta:
One minor point: It is recommended to run your own node. It does not matter if it is a local or remote node. A remote node is not necessary a third party node and thus not always less secure.

rating89us:
> A remote node is not necessary a third party node and thus not always less secure.

You're right. I've updated the descriptions.

PR #2925, opened 2020-05-26
Account: account page redesign

rating89us:
Changes:
- remove first address of each account (and copy to clipboard button): displaying a single address for each account incentivizes address reuse, discourages use of subaddresses, and gives a false idea that each account has an "official" or "main" address
- remove "Balance:" label
- add XMR to balances
- align balances
- selected account now has highlighted background color

Before:
image

After:
accountselect


rating89us:
See #3128

PR #2928, opened 2020-05-29
workflows: Windows CMake 'release-win64' target CI

xiphon:


xiphon:
Updated monero submodule head to include libboost build fix https://github.com/monero-project/monero/pull/6612

Issue #2024, opened 2019-03-18
GUI Re-design improvements. 


Electricsheep01:
![Asset 1](https://user-images.githubusercontent.com/32292415/54509730-80afe580-4985-11e9-8d5a-009d1877bb55.png)

I've done some re-designing of the GUI of the past few days of which I would greatly appreciate some feedback on. I have broken down each main aspect of the GUI I have changed below. If there is too much individual changes to be made for a single issue I'd be happy to break this thread up into multiple issues. I understand not everyone will like each change I have made but don't look at this as a package deal, rather each element has merit on its own independent from my overall design. 

1) The side bar is what I first set out to re-design. Frankly the original sidebar is a bit of a eye sore from the clunky nature of the menus when navigating them to the use of multiple colored dots (not sure why this was done, basic design mistake). The side, top and bottom sections are hex color #23272a, a slight change from the original #1a1a1a. Deep blacks like #1a1a1a never look that great and look even worse when on a very high or very low end monitor. All icons are from font awesome, they are colored white with a 50% opacity to blend some background color in. 

2) The latest GUI update introduced the 'Account' menu option - I really don't see the purpose of this as it is basically the same as the Receive but it calls addresses accounts instead. This just adds unnecessary cognitive load to the user, I suggest having a Dashboard section instead (which I will do a mock up design of soon) which is hierarchical in its nature as I have in my design. 
This dashboard section would give a brief one page overview of your wallet. Everything from Balance / unlocked balance, ability to switch between the different wallet modes (though I also think this feature is unnecessary, simple mode should be the default option with the 'advanced' features simply part of settings), syncing details, mining H/s (if enabled), addresses (list maybe 5 or so) and tx history overview (latest 5-10 tx's for example). The dashboard button has a slight gradient going from a light orange #ff6c37 at the top to a darker orange #bf5129 at the bottom - it is suble but looks a lot nicer than a static orange. 

3) A simple addition, adding a visibility toggle for your balance next to your balance. Clicking the eye hides the balance and unlocked balance.

4) Moved the syncing section to a new bottom menu. Looks a lot less cluttered than having it in the side menu. A total syncing % should also be added (combining both wallet / daemon syncs). Users like to be able to quickly see the progress status - just listing x amount of blocks remaining isn't that meaningful to most people. If the user is using a remote node / simple mode this area will say just that instead. This area also has quite a lot of extra space available that we could add other useful information (mining H/s being one example). 

[...]

PR #2936, opened 2020-06-04
Receive: redesign receive page

rating89us:
Changes:
- Every time the receive page is opened, a fresh address is visible and ready to use.
- A fresh address is an address which the full address was never displayed (as text or QR code) or copied to the clipboard.
- A new fresh address is generated when the fresh address is displayed/copied to the clipboard (address is generated in the background) or when the create address button is clicked. Fresh addresses are generated without label.
- Previous addresses are previous fresh addresses that were displayed/copied to the clipboard. They not necessarily have received funds.
- Previous addresses are always hidden when the receive page is opened. This is a private by default design: if the user is clueless, he/she will probably use the option that is more prominent (fresh address), and will avoid reusing a previous address.
- Added scrolling to previous addresses list (right column with QR code is always visible).
- Known limitation: currently it's not possible to edit the first address (#0) label, because it would edit the account label
- I decided to always display the first address (#0) of every account, even for account #0, which starts with 4...
- removed "monero:" before the address that is copied to clipboard with the Copy to clipboard button on the right column, therefore this button now has the same function as the copy button in the fresh address and previous address list.

Some gifs:
receivepage

receivepage2

- Instructions added to question mark icon (they are displayed for the first 3 addresses generated, to educate users):
helptext


SamsungGalaxyPlayer:
> Previous addresses are previous fresh addresses that were displayed/copied to the clipboard. They not necessarily have received funds.

We definitely need ways to filter only by addresses that have received funds or have labels. Else some addresses that may matter could be buried.

Can we also provide an export of the addresses and labels?

selsta:
FWIW we have to make sure that this PR does not get too large else we can’t review it. Both of the suggestions above can be done in separate PRs at a later point.

rating89us:
Updated:
filteradddress

xmrdsc:
Probably gonna take a shot at reviewing this soon even though it's quite large :-P!

xmrdsc:
@rating89us I have not looked at the code properly, but can you confirm this also increases the wallet's subaddress lookahead automatically? 

Another thing I'm wondering; if, by general usage, you've managed to generate 5000 sub-addresses, and you import your wallet on another computer (via mnemonic seed), then you would need to take extra precaution by setting the lookahead values correctly, correct?

garlicgambit:
Like the way this looks.  
  
@rating89us, can you break this PR in smaller parts to make it easier to review and merge?  
  
Couple suggestions:  
  
- Put the help instructions permanently on the page. Or increase the number to 50 or 100 addresses. Our preference would be to make it permanent.   
- Create new addresses with labels [without a new popup window](https://github.com/monero-project/monero-gui/issues/3069). Integrate the label input section on the page.

MoneroArbo:
Seeing this PR, I still think it would be good to have options to hide used addresses specifically and/or to reverse the sort order, as well as a way to quickly visually distinguish between used and unused addresses.

Maybe a way to "pin" certain addresses that you specifically intend for re-use could also be considered. I get that we don't want to encourage re-use, but some exchanges make you register each and every address you want to withdraw to, so that generating new addresses each time isn't necessarily worth it.

xanoni:
**It would be nice to also have a way to manually hide addresses. This can be useful, for example, if you attached an address to your identity but then the payment never arrives. We don't want this address to be shown as untainted.**

But overall, this is great. We desperately need a way to hide used addresses, otherwise we encourage address reuse for no reason. Not as bad as in Bitcoin, but still bad.

xanoni:
Just wanted to check that you are all aware of this PR: https://github.com/monero-project/monero-gui/issues/2024

Hope there are no conflicts between them. Haven't checked in detail.

elibroftw:
I know this is stale, but if this is important, please make sure that the default shows addresses with labels. People like me have specific addresses for say: donations, or exchanges (I know Kraken requires email confirmation to add a withdraw address).

PR #2937, opened 2020-06-07
workflows: update msys2 setup action v0 -> v1

selsta:
Decreases msys2 setup time from ~12 minutes to ~4 minutes.

Will rebase after #2928 is merged.


eine:
It seems that the same enhancement can be applied to [monero-project/monero](https://github.com/monero-project/monero).

selsta:
Yep, testing it currently :)

PR #2943, opened 2020-06-10
openpgp: implement public subkeys support

xiphon:


moneromooo-monero:
I'm not familiar at all with this code or the pgp api, but I've been asked to review, so all I can say is that I do not see anything obviously wrong here.

PR #2947, opened 2020-06-11
main: update balance only when wallet is synced

### LINKED ISSUE #2945: On restore from seed, balance changes from ?.?? to 0.0000... mid-scan
Steps to reproduce:

1. Run monero-gui-mac-x64-v0.16.0.0
2. Restore a mainnet wallet from seed, with no restore height specified. Use a random seed generated by https://moneroaddress.org/. Use a remote node, such as node.supportxmr.com
3. Observe that the balance will show as ?.?? for several hundred thousand blocks, and then suddenly switch to display 0.000000000000000 even though there are hundreds of thousands of blocks left to scan.

I'm concerned that people may misinterpret this as the wallet suddenly completing the sync and reporting an empty balance.

rating89us:
fixes #2945

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -416,7 +416,7 @@ ApplicationWindow {
 
         var balance = "?.??";
         var balanceU = "?.??";
-        if(!hideBalanceForced && !persistentSettings.hideBalance){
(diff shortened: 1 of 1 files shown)


selsta:
This makes problems when opening a wallet :/ It stays at "?.??" for a couple seconds.

PR #2948, opened 2020-06-12
QML: fix some Qt 5.15 warnings

selsta:
Untested with Qt 5.9, the `function onFoo() {}` syntax might not be available.

### DIFF
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -265,11 +265,11 @@ Rectangle {
     Connections {
         ignoreUnknownSignals: false
         target: transferView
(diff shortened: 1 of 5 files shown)


xmrdsc:
`function foo()` is valid javascript, works fine on 5.9.*

xiphon:
Actually it doesn't work on 5.9.*, 'Send' and 'Sweep' buttons don't react. https://github.com/monero-project/monero-gui/pull/3014

selsta:
i guess dsc was talking more about the syntax itself

@xiphon thanks for fixing

PR #2949, opened 2020-06-12
PasswordDialog: remove already defined z value

selsta:
PasswordDialog already defines z as parent.z + 2


xiphon:
We better drop this form PasswordDialog instead.

selsta:
@xiphon z+2 or z+1?

PR #2950, opened 2020-06-13
LeftPanel: simpifly color binding

selsta:
Didn’t know about `Binding on value` when writing this code.

Edit: I guess this could have been done with a simple if too but this might be a bit more clear.


xmrdsc:
nice find.

PR #2959, opened 2020-06-16
SettingsLayout: ask password for password relevant setting

selsta:
https://www.reddit.com/r/Monero/comments/h9dhcy/maam_monero_ask_anything_monday_june_15_2020/fuxaej6/


rating89us:
Maybe we could ask for the wallet password only when unchecking the setting, which is what an attacker would do.

Issue #2509, opened 2019-11-27
"Primary address" & "Primary account" are not being translated in any language


rating89us:
![image](https://user-images.githubusercontent.com/45968869/69725957-bbc30500-111f-11ea-8504-cb5500b500e8.png)

![image](https://user-images.githubusercontent.com/45968869/69726022-ddbc8780-111f-11ea-8880-7aae84c5178e.png)

erciccione:
Looks correctly translated in the language file: https://github.com/monero-project/monero-gui/blob/master/translations/monero-core_es.ts#L390

I guess it doesn't get picked for some reason.

rating89us:
> Looks correctly translated in the language file: https://github.com/monero-project/monero-gui/blob/master/translations/monero-core_es.ts#L390
> 
> I guess it doesn't get picked for some reason.

The translations are in History page. 
What is not being translated is in Account and Receive pages, as well as in Balance card.

PR #2965, opened 2020-06-22
main.qml: add "primary account" translation in balance card

rating89us:
fixes part of #2509


rating89us:
@selsta I think this can be merged in the next release

selsta:
@rating89us 

The reason I didn’t add this yet is because it’s a bit messy:

- Only the account label on the balance card gets translated, the one on the Account page stays in English resulting in a inconsistency.
- If someone decides to call a random account "Primary account" it will suddenly display translated text, even if they entered it in English.

I don’t have a better solution but I’m not happy with this one.

rating89us:
updated

PR #2966, opened 2020-06-23
build: set submodule to v0.16.0.1

selsta:


xiphon:
You also have to set monero submodule head to v0.16.0.1

PR #2968, opened 2020-06-26
workflows: add experimental mac cmake

selsta:
`continue-on-error` set to true for now because cmake on macOS does not fully work yet.


selsta:
Can’t get it to work properly..

PR #2969, opened 2020-06-26
workflows: add experimental mac cmake

selsta:
The Mac cmake CI will show as success until Mac cmake support is fully added.

Not sure if we want this, might be useful to see how cmake changes are on macOS. Can close if not wanted.


selsta:
#2991

PR #2973, opened 2020-06-28
RemoteNodeEdit: fix treatment of IPv6 addresses

### LINKED ISSUE #2971: Bug using IPV6
#### Setup in the remote daemon is:
- daemon version: `monero-x86_64 linux-gnu-v0.16.0.0`
- IP address: `fc00:db9::1` 
```
# monerod.conf
public-node=1
rpc-use-ipv6=1
rpc-bind-ip=0.0.0.0                                         
rpc-bind-ipv6-address=::0                                     
rpc-restricted-bind-port=18089  
confirm-external-bind=1
```
#### In my computer:
- GUI version: `monero-x86_64 linux-gnu-v0.16.0.0`
- IP address: `fc00:db8::1`

#### Steps: 

- Run the GUI
- Click on `Remote Node` under `node` tab
- Set address to:  `[fc00:db9::1]`, port to: `18089` and click on `connect`

The connection seems successful, but the fields `address` and `port` are not showing the right info, so please look at the screen capture.

image

**Note**: if we put the address without `[]`, the connection doesn't seem to be successful

jeandudey:
Now RemoteNodeEdit items check if the host is an IPv6 address and fixes
the return of `getAddress()`, for example, if it's an IPv6 address, it
returns [2001:db8::1]:18081 instead of 2001:db8::1:18081 which is
ambiguous and also showed up incorrectly on the GUI.

Previously the user had to enter the address inside square brackets in
order to connect to an IPv6 remote node (and it succeed, ofc) however
the GUI didn't parsed it correctly.

These bugs were found during the test of the **Locha Mesh** network (which
uses `fc00::/16` unique local addresses) as can be seen on the
[demonstration].

[demonstration]: https://twitter.com/Locha_io/status/1276930442030133248

The incorrect parsing is showed on issue #2971

The `getAddressParts()` will try to guess, depending on the input, how
the address is structured, if it is:

1. IPv6 with a port.
2. IPv6 without a port.
3. IPv4/DNS with a port.
4. IPv4/DNS without a port.
5. Treat anything else as an error and use the entire string as the
host.

Fixes #2971 

Signed-off-by: Jean Pierre Dudey <jeandudey@hotmail.com>

### DIFF
--- a/components/RemoteNodeEdit.qml
+++ b/components/RemoteNodeEdit.qml
@@ -73,6 +73,12 @@ GridLayout {
         if(addr === "" || addr.length < 2) return "";
         if(!Utils.isNumeric(port)) return "";
 
+        // IPv6 addresses get special treatment, they're enclosed in square
+        // brackets, e.g.: [fe80:cafe::1]: 18081.
+        if (Utils.isIPv6(addr)) {
+            return "[" + addr + "]:" + port;
+        }
+
         return addr + ":" + port;
     }
 
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -137,3 +137,69 @@ function capitalize(s){
 function removeTrailingZeros(value) {
     return (value + '').replace(/(\.\d*[1-9])0+$/, '$1');
 }
+
+function ipv6Regexes(){
+    return {
+        common: /^(::)?((?:[0-9a-f]+::?)+)?([0-9a-f]+)?(::)?(%[0-9a-z]{1,})?$/i,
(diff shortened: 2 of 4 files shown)


xiphon:
We better just store remote/bootstrap node host and port separately.

jeandudey:
> We better just store remote/bootstrap node host and port separately.

Seems better, still the IP address needs to be tested if it's IPv6 to format it correctly.

For now I see this handling more manageable, doesn't change behavior at all, just the IPv6 bug.

xiphon:
Please test https://github.com/monero-project/monero-gui/pull/2978 PR and let me know if something doesn't work as expected.

PR #2974, opened 2020-06-29
docker: implement Docker Linux 'release-static' build

xiphon:


selsta:
My findings (not sure if related to this PR or cmake in general)

- monero-gui does not display an icon in Ubuntu
- Version number does not work (compiling v0.16.0.2 results in e.g. 10d4ee7fd for both monero and monero-gui version number in Settings -> Info)

I would suspect these are cmake related.

xiphon:
> monero-gui does not display an icon in Ubuntu

Updated, should be fixed now.

> Version number does not work (compiling v0.16.0.2 results in e.g. 10d4ee7fd for both monero and monero-gui version number in Settings -> Info)

Fixed via https://github.com/monero-project/monero-gui/pull/2977

PR #2977, opened 2020-07-01
cmake: fix git tag + commit detection and version.js generation

xiphon:


selsta:
cmake:

Screenshot 2020-07-01 at 22 52 04

qmake:

Screenshot 2020-07-01 at 22 54 12

Is the difference intended? Otherwise works good.

xiphon:
> Is the difference intended? Otherwise works good.

Yes. Want to always include commit hash in the version string.

PR #2984, opened 2020-07-03
Prevent automatic monero-gui.desktop file creation

### LINKED ISSUE #2598: Use xdg-desktop-menu to install menu shortcut instead of writing to the file.
Some menus will not refresh when a .desktop file is updated until the system is restarted.

If the menu entry is used before the system is restarted it is overwritten back to the older version causing confusion.

See: https://linux.die.net/man/1/xdg-desktop-menu

### LINKED ISSUE #2815: Prevent monero-gui.desktop file creation
It seems that every time I run Monero GUI, it creates the following desktop file: `~/.local/share/applications/monero-gui.desktop`. I delete it, but it always creates the file when running. Can this be disabled somehow? I don't want that file (I don't want this custom entry to show up in my app menu) and the resulting file doesn't even work in my system. Is there any way to disable it?

xmrdsc:
Disables automatic writing of xdg entry on Linux machines to `~/.local/share/applications/monero-gui.desktop`. 

- User needs to give consent to write to xdg location(s).
- GUI will keep the xdg entry updated to latest version if the path already exists.
- GUI will refresh xdg with `update-desktop-database`
- Added xdg icon

Closes #2815
Closes #2802
Closes #2598

### DIFF
--- a/pages/settings/SettingsLayout.qml
+++ b/pages/settings/SettingsLayout.qml
@@ -251,14 +251,29 @@ Rectangle {
             }
         }
 
-        MoneroComponents.StandardButton {
-            visible: !persistentSettings.customDecorations
-            Layout.topMargin: 10
-            small: true
-            text: qsTr("Change language") + translationManager.emptyString
+        RowLayout{
+            spacing: 20
 
-            onClicked: {
-                appWindow.toggleLanguageView();
+            MoneroComponents.StandardButton {
+                visible: !persistentSettings.customDecorations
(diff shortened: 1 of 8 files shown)


selsta:
Do you think we could hide the button once a desktop entry is created?

xmrdsc:
Good idea, fixed.

xmrdsc:
Might need #2598 to also refresh xdg

xmrdsc:
closed due lack of interest.

PR #2986, opened 2020-07-06
QR code for the address book

### LINKED ISSUE #2929: Address Book: add option to display QR code
There could be a display QR code button for every contact in the address book. 

This would be useful if you want to send Monero from a mobile wallet to a contact that is only available in the GUI wallet.

xmrdsc:
- Show QR code for address book entries like on the receive page
- Introduces new QML component `QRComponent.qml`

fixes #2929

### DIFF
--- a/components/QRComponent.qml
+++ b/components/QRComponent.qml
@@ -1,0 +1,127 @@
+// Copyright (c) 2014-2020, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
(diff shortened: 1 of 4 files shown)


components/QRComponent.qml
```diff
@@ -0,0 +1,130 @@
...
+import QtQuick.Layouts 1.1
+import QtQuick.Dialogs 1.2
+import FontAwesome 1.0
+
+import "../components" as MoneroComponents
+import "../components/effects/" as MoneroEffects
+
+import moneroComponents.Clipboard 1.0
+import moneroComponents.Wallet 1.0
+import moneroComponents.WalletManager 1.0
```
selsta, line 41: Line 29 - 46: some of these imports are unused.

xmrdsc:
Thanks, fixed

components/QRComponent.qml
xmrdsc: fixed

rating89us:
In my opinion this page is looking very similar to the receive page now. Maybe some users could mistake the address book with the receive page and give the wrong address to the sender.

xmrdsc:
closed due lack of interest.

PR #2989, opened 2020-07-07
Support comma seperator for amount field, dynamically change to dot. 

### LINKED ISSUE #2951: Transfer: amount field should also accept comma as decimal separator
If the user types 0,01 in the amount field, it should be accepted as valid, or the field should replace the comma with a dot (changing it to 0.01).

This is useful for languages that use comma as a decimal separator, specially when typing the value using the numeric keypad.

xmrdsc:
Fixes #2951

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -296,6 +296,7 @@ Rectangle {
                   inlineButtonText: qsTr("All") + translationManager.emptyString
                   inlineButton.onClicked: amountLine.text = "(all)"
                   onTextChanged: {
+                        amountLine.text = amountLine.text.replace(",", ".");
                         const match = amountLine.text.match(/^0+(\d.*)/);
                         if (match) {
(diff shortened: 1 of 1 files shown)


pages/Transfer.qml
```diff
@@ -311,7 +312,7 @@ Rectangle {
                   }
 
                   validator: RegExpValidator {
-                      regExp: /^(\d{1,8})?([\.]\d{1,12})?$/
+                      regExp: /^(\d{1,8})?([\.|,]\d{1,12})?$/
```
xiphon, line 315: Treats `1|1` as a valid amount

xmrdsc: Thanks, my regex sucks, please check my fix.

xiphon: Should be `^(\d{1,8})?([\.,]\d{1,12})?$`

xmrdsc: Ehh, yes. Thanks.

PR #2992, opened 2020-07-08
history: add "In" and "To" fields + add address label

### LINKED ISSUE #2703: Transactions: Search field doesn't filter transactions when there is a pending transaction
GUI v0.15.0.2 (Qt 5.9.7)

To reproduce:
1) Send a transaction
2) Go to Transactions page
3) Use search filter. Filter doesn't work while transaction status is Pending
5) When transaction has 1 confirmation, filter starts working again

### LINKED ISSUE #2987: Feature request: Display subaddress label and subaddress against each incoming transaction
Current situation:

User creates a subaddress to receive a payment, and assigns a label such as "Rent from Bob"
User sees incoming transaction, then has to click the 'down arrow' and then click the 'i' button to see which subaddress it came from. Then the user needs to keep a note of that address, go back to the 'Receive' tab, and then scroll through the list of subaddresses to match the subaddress to the label that they assigned. Only then will they know it was from Bob. If lots of subaddresses have been issued, this will be a very tedious process.

Proposal:
In the transactions section, show the abbreviated subaddress and full subaddress label next to each transaction. Then, when a transaction is received, the user can immediately see it was "Rent from Bob" without further effort.

rating89us:
Changes:
- removed trailing zeros in Fee

Changes in Received transations:
- added "In" field to the Blockheight field position. This field will display the receiving address index + label (from the receive page) or the truncated address (when the address has no label) (fixes #2987)
- Blockheight field moved to the bottom.
- increased height of collapsed received transactions to the same as sent transactions (320 pixels), in order to display the blockheight. As a side effect, transaction key is also being displayed for received transactions.

Changes in Sent transactions:
- moved "To" field to the Blockheight field position. This field will now display either an address book icon + a destination address label (from the address book) or a truncated destination address (when the destination address has no address book entry).
- Blockheight field moved to the bottom.
- changed "Waiting for transaction to leave txpool" message to "Waiting confirmation..." in order to fit in the smaller space and because it's easier to understand.
- transactions with amount = 0 XMR will display destination as "My wallet", since they are seen in self transactions after a wallet is rescanned or restored from keys/seed.
- in restored wallets, transactions sending to the same wallet will display "Unknown amount", since we can't recover the amount information from the blockchain

image

image

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -573,8 +573,7 @@ Rectangle {
                 anchors.right: parent ? parent.right : undefined
                 height: {
                     if(!collapsed) return 60;
-                    if(isout && delegate.address !== "") return 320;
-                    return 220;
+                    return 320;
                 }
                 color: {
                     if(!collapsed) return "transparent"
@@ -618,6 +617,7 @@ Rectangle {
                             spacing: 0
                             clip: true
                             Layout.preferredHeight: 120
+                            Layout.minimumWidth: 180
 
                             Rectangle {
                                 color: "transparent"
@@ -633,17 +633,7 @@ Rectangle {
                                 MoneroComponents.TextPlain {
                                     font.family: MoneroComponents.Style.fontRegular.name
                                     font.pixelSize: 15
-                                    text: {
-                                        if (!isout) {
-                                            return qsTr("Received") + translationManager.emptyString;
-                                        }
-                                        const addressBookName = currentWallet ? currentWallet.addressBook.getDescription(address) : null;
-                                        if (!addressBookName)
-                                        {
-                                            return qsTr("Sent") + translationManager.emptyString;
-                                        }
-                                        return addressBookName;
-                                    }
+                                    text: (isout ? qsTr("Sent") : qsTr("Received")) + translationManager.emptyString
                                     color: MoneroComponents.Style.historyHeaderTextColor
                                     anchors.verticalCenter: parent.verticalCenter
                                     themeTransitionBlackColor: MoneroComponents.Style._b_historyHeaderTextColor
@@ -659,7 +649,7 @@ Rectangle {
                                 MoneroComponents.TextPlain {
                                     font.family: MoneroComponents.Style.fontRegular.name
                                     font.pixelSize: 15
-                                    text: displayAmount
+                                    text: (amount == 0 ? qsTr("Unknown amount") : displayAmount) + translationManager.emptyString
                                     color: MoneroComponents.Style.defaultFontColor
                                     anchors.verticalCenter: parent.verticalCenter
 
@@ -711,7 +701,7 @@ Rectangle {
                                     font.pixelSize: 15
                                     text: {
                                         if(!isout && confirmationsRequired === 60) return qsTr("Yes") + translationManager.emptyString;
-                                        if(fee !== "") return fee + " XMR";
+                                        if(fee !== "") return Utils.removeTrailingZeros(fee) + " XMR";
                                         return "-";
                                     }
 
@@ -739,6 +729,7 @@ Rectangle {
                             spacing: 0
                             clip: true
                             Layout.preferredHeight: 120
+                            Layout.minimumWidth: 230
 
                             Rectangle {
                                 color: "transparent"
@@ -754,7 +745,7 @@ Rectangle {
                                 MoneroComponents.TextPlain {
                                     font.family: MoneroComponents.Style.fontRegular.name
                                     font.pixelSize: 15
-                                    text: qsTr("Blockheight") + translationManager.emptyString
+                                    text: (isout ? qsTr("To") : qsTr("In")) + translationManager.emptyString
                                     color: MoneroComponents.Style.historyHeaderTextColor
                                     themeTransitionBlackColor: MoneroComponents.Style._b_historyHeaderTextColor
                                     themeTransitionWhiteColor: MoneroComponents.Style._w_historyHeaderTextColor
@@ -768,15 +759,44 @@ Rectangle {
                                 Layout.preferredHeight: 20
 
                                 MoneroComponents.TextPlain {
+                                    id: addressField
                                     font.family: MoneroComponents.Style.fontRegular.name
-                                    font.pixelSize: 14
-                                    text: blockheight > 0 ? blockheight : qsTr('Pending') + translationManager.emptyString;
+                                    font.pixelSize: 15
+                                    text: {
+                                        if (isout) {
+                                            if (address) {
+                                                const addressBookName = currentWallet ? currentWallet.addressBook.getDescription(address) : null;
+                                                return (addressBookName ? FontAwesome.addressBook + " " + addressBookName : TxUtils.addressTruncate(address, 8));
+                                            }
+                                            if (amount != 0) {
+                                                return (blockheight ? qsTr("Unknown recipient") : qsTr("Waiting confirmation...")) + translationManager.emptyString;
+                                            } else {
+                                                return qsTr("My wallet") + translationManager.emptyString;
+                                            }
+                                        } else {
+                                            const receivingAddress = currentWallet ? currentWallet.address(subaddrAccount, subaddrIndex) : null;
+                                            const receivingAddressLabel = currentWallet ? appWindow.currentWallet.getSubaddressLabel(subaddrAccount, subaddrIndex) : null;
+                                            if (receivingAddress) {
+                                                if (subaddrIndex == 0) {
+                                                    return qsTr("Address") + " #0" + " (" + qsTr("Primary address") + ")" + translationManager.emptyString;
+                                                } else {
+                                                    if (receivingAddressLabel) {
+                                                        return qsTr("Address") + " #" + subaddrIndex + " (" + receivingAddressLabel + ")" + translationManager.emptyString;
+                                                    } else {
+                                                        return qsTr("Address") + " #" + subaddrIndex + " (" + TxUtils.addressTruncate(receivingAddress, 4) + ")" + translationManager.emptyString;
+                                                    }
+                                                }
+                                            } else {
+                                                return qsTr("Unknown address") + translationManager.emptyString;
+                                            }
+                                        }
+                                    }
 
                                     color: MoneroComponents.Style.defaultFontColor
                                     anchors.verticalCenter: parent.verticalCenter
 
                                     MouseArea {
-                                        state: "copyable"
+                                        state: isout ? "copyable_address" : "copyable_receiving_address"
                                         anchors.fill: parent
                                         hoverEnabled: true
                                         onEntered: parent.color = MoneroComponents.Style.orange
@@ -847,6 +867,7 @@ Rectangle {
                             spacing: 0
                             clip: true
                             Layout.preferredHeight: 120
+                            Layout.minimumWidth: 130
 
                             Rectangle {
                                 color: "transparent"
@@ -909,9 +930,15 @@ Rectangle {
                                 Layout.preferredHeight: 10
                             }
 
+                            Rectangle {
+                                color: "transparent"
+                                Layout.fillWidth: true
+                                Layout.preferredHeight: 10
+                            }
+
                             Item {
                                 Layout.fillWidth: true
-                                Layout.preferredHeight: 60
+                                Layout.preferredHeight: 50
 
                                 MoneroComponents.StandardButton {
                                     id: btnDetails
@@ -919,7 +946,7 @@ Rectangle {
                                     small: true
                                     label.font.family: FontAwesome.fontFamily
                                     fontSize: 18
-                                    width: 28
+                                    width: 34
 
                                     MouseArea {
                                         state: "details"
@@ -949,7 +976,7 @@ Rectangle {
                                     small: true
                                     label.font.family: FontAwesome.fontFamilyBrands
                                     fontSize: 18
-                                    width: 36
+                                    width: 34
 
                                     MouseArea {
                                         state: "proof"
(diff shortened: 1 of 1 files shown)


xiphon:
Please split into separate PRs

rating89us:
> Please split into separate PRs

Done

selsta:
Screenshot 2020-09-29 at 03 34 01

Would like a bit more margin between In | Date

selsta:
Looks good now IMO, please squash the commits.

pages/History.qml
```diff
@@ -769,14 +760,44 @@ Rectangle {
...
+                                        if (isout) {
+                                            if (address !== "") {
+                                                const addressBookName = currentWallet ? currentWallet.addressBook.getDescription(address) : null;
+                                                if (!addressBookName) {
+                                                    return TxUtils.addressTruncate(address, 8);
+                                                } else {
+                                                    return FontAwesome.addressBook + " " + addressBookName;
+                                                }
+                                            }
+                                            if (blockheight === 0) {
```
selsta, line 774: return (blockheight ? qsTr("Unknown recipient") : qsTr("Waiting confirmation...")) + translationManager.emptyString;

pages/History.qml
```diff
@@ -769,14 +760,44 @@ Rectangle {
 
                                 MoneroComponents.TextPlain {
                                     font.family: MoneroComponents.Style.fontRegular.name
-                                    font.pixelSize: 14
-                                    text: blockheight > 0 ? blockheight : qsTr('Pending') + translationManager.emptyString;
+                                    font.pixelSize: 15
+                                    text: {
+                                        if (isout) {
+                                            if (address !== "") {
```
selsta, line 766: minor but just `if (address)` seems a bit easier to read

inside the if you could also use the ternary operator

pages/History.qml
```diff
@@ -769,14 +760,44 @@ Rectangle {
...
+                                            }
+                                            if (blockheight === 0) {
+                                                return qsTr("Waiting confirmation...") + translationManager.emptyString;
+                                            } else {
+                                                return qsTr("Unknown recipient") + translationManager.emptyString;
+                                            }
+                                        } else {
+                                            const receivingAddress = currentWallet ? currentWallet.address(subaddrAccount, subaddrIndex) : null;
+                                            const receivingAddressLabel = currentWallet ? appWindow.currentWallet.getSubaddressLabel(subaddrAccount, subaddrIndex) : null;
+                                            if (receivingAddress) {
```
selsta, line 782: else missing, resulting in

```
2020-09-29 21:15:35.309	W qrc:/pages/History.qml:764:37: Unable to assign [undefined] to QString
```

rating89us: fixed

pages/History.qml
rating89us: i agree

pages/History.qml
rating89us: I agree

rating89us:
Added a new thing:
If you rescan your wallet (change wallet restore height) or restore it from keys/seed, your wallet will lose details from past transactions and all transactions that were sent to your own wallet will display an amount 0.000000000000 XMR. 

So in transactions with amount = zero, "To:" field will now display "My wallet"

image

selsta:
Unknown recipient -> You can click on it, it turns orange and it says "copied to clipboard" but nothing is in clipboard.

rating89us:
Fixed. It will now copy the text "My wallet" or "Unknown recipient".

Even in "My wallet" we don't know the exact receiving address, because it can be either the main address or a subaddress.

pages/History.qml
```diff
@@ -633,17 +633,7 @@ Rectangle {
...
-                                            return qsTr("Received") + translationManager.emptyString;
-                                        }
-                                        const addressBookName = currentWallet ? currentWallet.addressBook.getDescription(address) : null;
-                                        if (!addressBookName)
-                                        {
-                                            return qsTr("Sent") + translationManager.emptyString;
-                                        }
-                                        return addressBookName;
-                                    }
+                                    text: isout ? qsTr("Sent") : qsTr("Received") + translationManager.emptyString
```
xiphon, line 636: `1 ? 2 : 3 + 4` == `2`

pages/History.qml
```diff
@@ -768,15 +759,44 @@ Rectangle {
...
+                                                return (blockheight ? qsTr("Unknown recipient") : qsTr("Waiting confirmation...")) + translationManager.emptyString;  
+                                            } else {
+                                                return qsTr("My wallet") + translationManager.emptyString;
+                                            }
+                                        } else {
+                                            const receivingAddress = currentWallet ? currentWallet.address(subaddrAccount, subaddrIndex) : null;
+                                            const receivingAddressLabel = currentWallet ? appWindow.currentWallet.getSubaddressLabel(subaddrAccount, subaddrIndex) : null;
+                                            if (receivingAddress) {
+                                                if (subaddrIndex == 0) {
+                                                    return qsTr("Address #0") + " (" + qsTr("Primary address") + ")" + translationManager.emptyString;
```
xiphon, line 781: IMO you'd better move " #0" out of the translatable string

rating89us: fixed

pages/History.qml
rating89us: i don't see what's wrong here

selsta: add () around the ternary operator so that + emptyString gets applied to both

rating89us:
Some changes: 
1) I changed the 0.000000000000 XMR to "Unknown amount" in transactions sent to the same wallet:
- The amount entered by the user is always >0 XMR, and we can't recover it from the blockchain.
- The net balance of the transacion = the transaction fee, therefore it's negative. Displaying 0 XMR seems that the net balance = 0 XMR.

2) Removed trailing zeros from Fee

3) Added/improved details in transaction details popup

pages/History.qml
rating89us: + emptyString is already being applied to both Sent and Received

selsta: no, it only gets applied to Received due to + having precedence over the ternary operator

rating89us:
Currently the To field can only display:
- To: _address string_ (this address can be from my wallet or external wallet)
- To: "My wallet" (transaction with amount 0.00000000 XMR in a recovered wallet)
- To: "Unknown recipient" (transaction with amount different from 0.00000 XMR in a recovered wallet)

To better differentiate case 1, we need a new function to detect if a given address is from the own wallet. The input would be the address string and the function would return true if the address is from the wallet or false if it's from an external wallet.

The resulting To field possibilities would be:
- To: "My wallet (_address string_)"
- To: "External address (_address string_)"
- To: "My wallet" (transaction with amount 0.00000000 XMR in a recovered wallet)
- To: "Unknown recipient" (transaction with amount different from 0.00000 XMR in a recovered wallet)

pages/History.qml
```diff
@@ -768,15 +759,44 @@ Rectangle {
...
-                                    font.pixelSize: 14
-                                    text: blockheight > 0 ? blockheight : qsTr('Pending') + translationManager.emptyString;
+                                    font.pixelSize: 15
+                                    text: {
+                                        if (isout) {
+                                            if (address) {
+                                                const addressBookName = currentWallet ? currentWallet.addressBook.getDescription(address) : null;
+                                                return (addressBookName ? FontAwesome.addressBook + " " + addressBookName : TxUtils.addressTruncate(address, 8));
+                                            }
+                                            if (amount != 0.000000000000) {
```
xiphon, line 771: Could we use `0` instead of `0.000000000000`?

pages/History.qml
rating89us: in my wallet the Sent string is being translated normally. And I've already seen this type of `A ? B : C + translationManager.emptyString` many times in monero GUI code, always working fine.

pages/History.qml
rating89us: I tried, but it doesn't seem to work

pages/History.qml
```diff
@@ -768,15 +759,44 @@ Rectangle {
...
+                                            if (receivingAddress) {
+                                                if (subaddrIndex == 0) {
+                                                    return qsTr("Address") + " #0" + " (" + qsTr("Primary address") + ")" + translationManager.emptyString;
+                                                } else {
+                                                    if (receivingAddressLabel) {
+                                                        return qsTr("Address") + " #" + subaddrIndex + " (" + receivingAddressLabel + ")" + translationManager.emptyString;
+                                                    } else {
+                                                        return qsTr("Address") + " #" + subaddrIndex + " (" + TxUtils.addressTruncate(receivingAddress, 4) + ")" + translationManager.emptyString;
+                                                    }
+                                                }
```
xiphon, line 788: Account number / label is ignored

rating89us: Transactions page only display transactions from the currently selected account

pages/History.qml
selsta: > in my wallet the Sent string is being translated normally.

Yes because `translationManager.emptyString` is technically not needed when using the ternary operator, not because the () are not needed.

> And I've already seen this type

Then it is wrong in these places too, would appreciate it if you could update it here, else it is wrong code that works by accident.

pages/History.qml
selsta: Yea, I think account stuff is not needed here, also there is not enough space for it.

pages/History.qml
selsta: are you sure? seems to work here

pages/History.qml
rating89us: My code is:
`isout ? qsTr("Sent") : qsTr("Received") + translationManager.emptyString`

Actually the computer reads this (but somehow translation is working):
`(isout ? qsTr("Sent") : qsTr("Received") + translationManager.emptyString)`

And you want me to change to this:
`(isout ? qsTr("Sent") : qsTr("Received")) + translationManager.emptyString`

Right?

selsta: > but somehow translation is working

Due to technically `translationManager.emptyString` not being needed at all in this case. But it is better if we still add it.

> And you want me to change to this:

Yes.

rating89us: Why is translationManager.emptyString not needed in this case? I thought that I should add `+ translationManager.emptyString` at the end of every line that contains a `qsTr("");`, so that the string gets added to the translation.

selsta: translationManager.emptyString is a workaround for a Qt bug, AFAIK it is only needed in cases like this:

```
text: amount
```

We add it to every string just to be sure.

pages/History.qml
xiphon: Gotcha

pages/History.qml
```diff
@@ -1510,7 +1527,7 @@ Rectangle {
             root.txModelData.push({
                 "i": i,
                 "isout": isout,
-                "amount": Number(amount),
+                "amount": Utils.removeTrailingZeros(displayAmount.toFixed(12)),
```
xiphon, line 1530: Why this change is needed?

rating89us: I removed Number() because amounts like 0.000001 were returning 0.

xiphon (commented):
`showTxDetails` changes look completely unrelated to the PR. if so, please submit unrelated changes in a separate PR.

pages/History.qml
rating89us: fixed (i was wrong)

pages/History.qml
xiphon: > I removed Number() because amounts like 0.000001 were returning 0.

How can i reproduce the issue?

rating89us: In this example, transactions #1 to #6 the wallet knows the amount. Transaction #7 the wallet doesn't know the amount (wallet was restored at this point).

`"amount": Utils.removeTrailingZeros(displayAmount.toFixed(12))`: all transactions with known amount  (#1-6) display the amount correctly. 
image


` "amount": Number(amount)`: some transactions (#3-6) with known amount display as Unknown amount. 
image

xiphon: Can't reproduce. Works well with `"amount": Number(amount)` (Qt 5.9.7). What Qt version do you use?

xiphon: Did you send the txes in question to the same wallet but to the different account or did you send them to a third party wallet?

selsta: I checked with Qt 5.12.9 and 5.15.1 and also can’t reproduce this issue. @rating89us can you double check?

rating89us: My wallet main address (account #1) is 48rg...
The first two transactions are from account #1 (48rg...) to account #2's primary address (89HP...) and its first subaddress (88PS...).
The remaining transactions are from account #1 to account #1 (48rg... to 48rg...)

I'm using Qt 5.12.9.
I'll check again if I can identify what's causing this.

rating89us: There is a comment in line 1510 :
```
// *sometimes* amount is 0, while the 'destinations string'
// has the correct amount, so we try to fetch it from that instead.
```

In my investigation "amount = 0" occurs when you send from account #1 to any address of account #1 (either main address or subaddress).

If I send from account #1 to account #2 (either main address or subaddress), the amount is correct.

xiphon: Reproduced the issue. Looking into it.

xiphon: https://github.com/monero-project/monero-gui/pull/3127 resolves the issue. Please confirm.

rating89us: Yes, it's working now. Thanks

rating89us: I won't fix this in this PR, since it will be fixed by your PR #3127 anyway

xiphon (approved):

Issue #2990, opened 2020-07-07
Feature idea: Show solo mining block reward and daily chance of winning


knaccc:
The GUI currently says "Mining also gives you a small chance to earn some Monero".

It would be much more compelling if it presented the user with the size of reward and probability of winning.

For example, if you were able to mine 1500H/s, it could tell you that you have a 1 in 1416 daily chance of winning 1.5813 XMR (those are real stats for a high-end laptop). That's much more exciting!

selsta:
#2994

PR #2994, opened 2020-07-08
Mining: Indicate probability of finding a block in one day

glv2:
For issue #2990

### DIFF
--- a/pages/Mining.qml
+++ b/pages/Mining.qml
@@ -251,7 +251,19 @@ Rectangle {
 
     function updateStatusText() {
         if (appWindow.isMining) {
-            statusText.text = qsTr("Mining at %1 H/s").arg(walletManager.miningHashRate()) + translationManager.emptyString;
+            var userHashRate = walletManager.miningHashRate();
+            if (userHashRate === 0) {
+                statusText.text = qsTr("Mining temporarily suspended.") + translationManager.emptyString;
+            }
+            else {
+                var blockTime = 120;
+                var blocksPerDay = 86400 / blockTime;
+                var globalHashRate = walletManager.networkDifficulty() / blockTime;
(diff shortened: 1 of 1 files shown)


pages/Mining.qml
```diff
@@ -251,7 +251,14 @@ Rectangle {
 
     function updateStatusText() {
         if (appWindow.isMining) {
-            statusText.text = qsTr("Mining at %1 H/s").arg(walletManager.miningHashRate()) + translationManager.emptyString;
+            var blockTime = 120;
+            var blocksPerDay = 86400 / blockTime;
+            var globalHashRate = walletManager.networkDifficulty() / blockTime;
+            var userHashRate = walletManager.miningHashRate();
```
xiphon, line 257: Doesn't handle zero hashrate case.

glv2: I added a "mining temporarily suspended" message for the zero hashrate case.

SamsungGalaxyPlayer: Why temporarily suspended? Why not something like "You are not currently mining"

glv2: This message shows when background mining is enabled, but temporarily suspended because some other program is using a lot of CPU. There is a different message when mining is not enabled ("Not mining", if I recall correctly).

pages/Mining.qml
```diff
@@ -251,7 +251,19 @@ Rectangle {
 
     function updateStatusText() {
         if (appWindow.isMining) {
-            statusText.text = qsTr("Mining at %1 H/s").arg(walletManager.miningHashRate()) + translationManager.emptyString;
+            var userHashRate = walletManager.miningHashRate();
+            if (userHashRate == 0) {
```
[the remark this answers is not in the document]
glv2: Done, and squashed.

xmrdsc:
naisu

PR #2995, opened 2020-07-08
history: change sent transaction dot color from orange to red

rating89us:
Red is the most commonly used color for outgoing transactions in cryptocurrency wallets.
image

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -595,7 +595,7 @@ Rectangle {
                         width: 10
                         height: 10
                         radius: 8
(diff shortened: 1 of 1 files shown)


xmrdsc:
I'm not sure if this is needed. @GBKS designed the color scheme.

xiphon:
IMO red dots look somewhat inconsistent

GBKS:
My issue with red is that in UI design it usually means that an error happened, which is not the case here. I'd suggest sticking with orange to indicate outgoing transactions.

elibroftw:
This should be closed then

PR #2996, opened 2020-07-08
history: fix search filter not working when there is a pending tx

### LINKED ISSUE #2703: Transactions: Search field doesn't filter transactions when there is a pending transaction

rating89us:
- Search filter now works when there is a pending transaction (fixes #2703)


xmrdsc:
nice

PR #2991, opened 2020-07-08
cmake: implement MacOS 'release' build + CI support

xiphon:

PR #2998, opened 2020-07-08
workflows: fix macOS build

selsta:


xiphon:
Do you mean `libgcrypt` is installed by default on MacOS?

selsta:
No but the macOS image of Github Actions has a lot of stuff preinstalled, looks like they added libgcrypt now.

selsta:
#2991

PR #3000, opened 2020-07-09
cmake: fix zxcvbn option description

selsta:


selsta:
Hmm, seems like this is failing. I was quite sure that I tested this but maybe not.

selsta:
Updated to only fix the description.

PR #3001, opened 2020-07-09
cmake: set current monero-gui version

selsta:
Useful for macOS bundle.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,10 +3,10 @@ project(monero-gui)
 
 message(STATUS "Initiating compile using CMake ${CMAKE_VERSION}")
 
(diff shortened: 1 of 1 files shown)


xiphon:
Don't like hardcoded version approach. Should be set automatically. Will implement this in a while.

selsta:
Would setting automatically also work in environments without git, e.g. when downloading the source code manually?

xiphon:
> Would setting automatically also work in environments without git, e.g. when downloading the source code manually?

Typically source code archive filename contains version suffix. Thinking about obtaining version info from the directory name in that case.

selsta:
Good point.

PR #3004, opened 2020-07-10
version: use MONERO_VERSION_FULL const as monero core version

xiphon:
There is a difference now in version strings format.
```
GUI version: v0.16.0.2-e0faddf96 (Qt 5.9.7)
Embedded Monero version: 0.15.0.0-596b71b29
```

We could either:
1. Strip "v" prefix from the GUI version
2. Add "v" prefix to Monero verison
3. Do nothing

I'm okay with option 1 or 3.


selsta:
I would prefer 1

xiphon:
Rebased

PR #3007, opened 2020-07-14
KeysFiles: fix WalletKeysFilesModel, should be instantiable QML type

xiphon:
Requires https://github.com/monero-project/monero-gui/pull/3006


xiphon:
Fixed commit message typo

PR #3010, opened 2020-07-14
WalletManager: displayAmount should be static

xiphon:


xiphon:
Updated. Found another one place.

PR #3012, opened 2020-07-14
SettingsLog: add command history

rating89us:
Allow users to navigate with up/down arrows the last commands entered in Settings > Log (during a session).

### DIFF
--- a/pages/settings/SettingsLog.qml
+++ b/pages/settings/SettingsLog.qml
@@ -213,9 +213,27 @@ Rectangle {
         MoneroComponents.LineEdit {
             id: sendCommandText
             Layout.fillWidth: true
+            property var lastCommands: []
+            property int currentCommandIndex
             fontBold: false
             placeholderText: qsTr("command + enter (e.g 'help' or 'status')") + translationManager.emptyString
             placeholderFontSize: 16
+            Keys.onUpPressed: {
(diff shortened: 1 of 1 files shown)


pages/settings/SettingsLog.qml
```diff
@@ -213,9 +213,27 @@ Rectangle {
...
+            Keys.onUpPressed: {
+                if (currentCommandIndex != 0) {
+                    sendCommandText.text = lastCommands[currentCommandIndex - 1]
+                    currentCommandIndex = currentCommandIndex - 1
+                }
+            }
+            Keys.onDownPressed: {
+                if (currentCommandIndex == lastCommands.length - 1) {
+                    currentCommandIndex = lastCommands.length;
+                    return text = "";
```
xiphon, line 230: ?

rating89us: 1) User enter values "zero", "one", "two", "three" that are stored in the array.
2) Use the up arrow to navigate to "three" (last index of array).
3) Pressing down arrow will display a blank line, which is the default behavior in a Ubuntu terminal. In a Windows prompt command, pressing down arrow would keep displaying "three", but I prefer Ubuntu behavior.

selsta: Maybe write it like this
```
text = "";
return;
```

PR #3050, opened 2020-08-26
cmake: fix DEV_MODE, checkout master in submodule

selsta:
Edit, seems like there is still an issue with the version number.

Also fails on Windows..


xiphon:
IMO `DEV_MODE` flag is meant to explicitly disable all the version/revision checks.

selsta:
There is also -DMANUAL_SUBMODULES=1, but not sure what the initial intention of DEV_MODE was.

xiphon (approved):
Looking at `execute_process(COMMAND cd monero && git checkout origin/master)` command it appears `DEV_MODE` is actually intended to `checkout master in submodule`.

xiphon (approved):

PR #3089, opened 2020-09-15
main: command line option for socks5 proxy

selsta:


pages/settings/SettingsLayout.qml
```diff
@@ -265,16 +266,17 @@ Rectangle {
 
         MoneroComponents.RemoteNodeEdit {
             id: proxyEdit
+            enabled: socksProxyFlag.length === 0
             Layout.leftMargin: 36
             Layout.topMargin: 6
             Layout.minimumWidth: 100
             placeholderFontSize: 15
-            visible: persistentSettings.proxyEnabled
+            visible: persistentSettings.proxyEnabled || socksProxyFlag.length > 0
```
xiphon, line 274: Use `proxyCheckbox.checked`

selsta: done, also updated the enabled part

pages/settings/SettingsLayout.qml
```diff
@@ -265,16 +266,17 @@ Rectangle {
 
         MoneroComponents.RemoteNodeEdit {
             id: proxyEdit
+            enabled: proxyCheckbox.enabled
```
xiphon, line 269: Shouldn't allow changing the proxy address if socksProxyFlag is set.

main.qml
```diff
@@ -1423,13 +1423,14 @@ ApplicationWindow {
         property string proxyAddress: "127.0.0.1:9050"
         property bool proxyEnabled: false
         function getProxyAddress() {
-            if (!proxyEnabled) {
+            var proxyAddressSetOrForced = socksProxyFlagSet ? socksProxyFlag : proxyAddress;
```
xiphon, line 1426: Could be declared at line 1430

main.qml
```diff
@@ -1423,13 +1423,14 @@ ApplicationWindow {
...
+            if (socksProxyFlagSet ? !socksProxyFlag : !proxyEnabled) {
```
xiphon, line 1427: IMO it would be better to not use ternary operator here

xiphon (approved):

PR #3092, opened 2020-09-18
cmake: support macdeployqt

selsta:
Replaces "make deploy" from current qmake build system.

https://sodocumentation.net/qt/topic/5857/deploying-qt-applications


selsta:
Replaced by #3103

PR #3103, opened 2020-09-22
cmake: add deploy target (mac / windows)

selsta:
thanks @iDunk5400 for testing


cmake/Deploy.cmake
```diff
@@ -0,0 +1,103 @@
+if(APPLE OR WIN32)
```
xiphon, line 1: Please exclude static windows builds

selsta: done

xiphon (approved):

PR #3121, opened 2020-09-29
Transfer: Create SuccessfulTxDialog

rating89us:
New SuccessfulTxDialog component:
- Green circle checkmark animation
- Open folder button (view-only wallet): opens the folder where the transaction file was saved

Normal transaction:
txflownew1

After creating transaction file in view-only wallet (offline transaction signing feature):
image


xiphon:
> Requires #2895

Why does it require #2895? Couldn't we just move the code showing new dialog to the current PR?

rating89us:
My mistake. I think all code is moved now

components/SuccessfulTxDialog.qml
```diff
@@ -0,0 +1,191 @@
...
+                KeyNavigation.tab: doneButton
+                Keys.enabled: openFolderButton.visible
+                Keys.onReturnPressed: openFolderButton.onClicked
+                Keys.onEnterPressed: openFolderButton.onClicked
+                Keys.onEscapePressed: {
+                    root.close()
+                    root.rejected()
+                }
+                onClicked: {
+                    oshelper.openContainingFolder(walletManager.urlToLocalPath(saveTxDialog.fileUrl))
```
xiphon, line 167: Could we use `onCloseCallback` for that?

main.qml
```diff
@@ -1499,6 +1501,19 @@ ApplicationWindow {
         }
     }
 
+    // Transaction successfully sent popup
+    SuccessfulTxDialog {
+        // dynamically change onclose handler
+        property var onCloseCallback
```
xiphon, line 1507: Looks like we can drop this

xiphon:
What do you think about adjusting the font size to fit transaction id on a single line?

rating89us:
Font size would need to be too small to fit in current dialog width.
And increasing dialog width would make it too large.
So I prefer the way it is now.

xiphon (approved):

selsta (approved):

PR #3123, opened 2020-09-30
cmake: update version

selsta:
bundle version still gets taken from here


xiphon (approved):
Should be the last commit before we tag the release

PR #3128, opened 2020-10-03
Account: account page redesign

rating89us:
Changes:
- remove first address of each account (and copy to clipboard button): displaying a single address for each account incentivizes address reuse, discourages use of subaddresses, and gives a false idea that each account has an "official" or "main" address
- remove "Balance:" label
- add XMR to balances
- align balances
- selected account now has highlighted background color

Before:
image

After:
accountselect


xiphon:
Please split into multiple PRs like:
1. remove "Balance:" label
    add XMR to balances
2. selected account now has highlighted background color and white account name

and so on..

rating89us:
> Please split into multiple PRs

This PR is now only about removing addresses and copy to clipboard button. The remaining changes I moved to PRs #3165 and #3164

pages/Account.qml
```diff
@@ -136,6 +137,7 @@ Rectangle {
 
                 MoneroComponents.TextPlain {
                     id: unlockedBalanceAll
+                    Layout.rightMargin: 60
```
xiphon, line 140: Seem you should adjust line 215 `anchors.rightMargin: 80` as well

jonathancross:
Is there a discussion of these changes somewhere?

I would like to hear the rational about the removal of these features (which I've used extensively in the past).  Churning funds between different accounts for example is easier when one does not need to switch to receive tab each time.  That is a small annoyance, but just wondering if there is more justification.

> address reuse

Safe address reuse (via stealth addresses) is one of the UX benefits that Monero can offer over Bitcoin for example.  Address reuse is not always appropriate, but is perfectly fine for accepting donations, sending funds to yourself and many other situations where the sender already knows that payments are going to the same recipient.  Is there a general agreement that users should be discouraged from ever reusing addresses?

> gives a false idea that each account has an "official" or "main" address

Well, there is a "Primary" address which is different than the others via 4... prefix. Unfortunatly.

PR #3133, opened 2020-10-06
LeftPanel: add hide balance icon in card

rating89us:
hide balance


selsta:
I would argue that this is quite a niche feature while everyone will always having this button displayed. Also it duplicates the functionality form the settings page.

How can we add this functionality for power user while not cluttering up the UI for the average user? What would be your use case for this button?

rating89us:
Some ideas in my mind:
- Many bank mobile apps and cryptocurrency wallets have this function easily accessible in the balance card, and some hide this function in the settings page, but I prefer the first option.
- Displaying this icon in balance card makes it easier for users to learn this function exists (it's important for users to know all functions that increases privacy).
- The changing icon makes it easier for users to understand why their balance is hidden. Many times I've watched my ?.?? balance and I thought there was a problem with sync (I forgot it was hidden). 
- Balances should be hidden in all pages. Users should be able to open their wallets, change account, check transaction history and send a new transaction, all without revealing its current balance. 
- If we're going to hide a lot of balances in many pages of the wallet, the unhide function should be easily accessible and always visible, and the balance card is the perfect place for it. See example below:

hideb2

tficharmers:
Whilst I would like the show/hide functionality to not be hidden in the settings, I'm not a fan of the eye icon cluttering the UI. So instead of the 'eye' icon to show/hide the balance, why not make the actual balance the show/hide toggle? Currently clicking on the balance copies it to the clipboard; how often does someone really need to do that? The same full balance copying mechanism is currently on the 'Accounts' page anyway.

Nice work rating89us. I like your thinking on these things.

Issue #3151, opened 2020-10-12
Cannot open view-only hardware wallet


viperperidot:
I created a view-only wallet by clicking the button in the Settings GUI interface. The wallet created successfully yet when I try to open it I get the message: 'Couldn't open wallet: file [view-only wallet file path] does not correspond to [view-only.keys path]'

I am using a Ledger Nano X. Not sure what went wrong but I have tried deleting the view-only wallet and keys and creating again with the same issue happening.

xiphon:
Ledger doesn't support creating view-only wallets (yet?).

viperperidot:
Oh really? Ok I didn't know that, because you can make watch-only wallets in Bitcoin using Ledger so they must have not implemented it yet for XMR. I could put a feature request on the Ledger GitHub I suppose but I am pretty sure it's probably already on their list.

xiphon:
Closed via https://github.com/monero-project/monero-gui/pull/3160

ksdhans:
As posted elsewhere, it looks like extracting the data needed from Ledger is possible:
https://shawnhogan.com/2021/07/how-to-get-monero-private-view-key-from-ledger.html

It would be great if exporting view only wallets for Ledger were added (with appropriate warnings). I know people argue that you shouldn't do it. However, there are situations where it's needed. For example, if you have a payment server (like BTCPay), then it needs the view keys in order to operate. It isn't practical to have the Ledger device permanently plugged in with the Monero app active.

PR #3160, opened 2020-10-17
SettingsWallet: disable creating view-only wallet for Ledger

rating89us:
Ledger doesn't support creating view-only wallet. See #3151 and https://www.reddit.com/r/monerosupport/comments/jcaet1/error_couldnt_open_wallet_file_mywallet_viewonly/


selsta (approved):

ksdhans:
It looks like extracting the data needed from Ledger is possible:
https://shawnhogan.com/2021/07/how-to-get-monero-private-view-key-from-ledger.html

It might be time to revisit this...

PR #3164, opened 2020-10-17
Account/receive: highlight background and text of selected item

### LINKED ISSUE #2572: Selected account/address is not highlighted enough
When you select an account in Account page or an address in Receive page, the selected item is not highlighted enough, because only the number (`#1`, `#2`, `#3`, etc.) changes color.

I suggest we change the background color of the selected item to the same one used when you hover the mouse cursor over it.

Also, all text in the selected item should change color, not only the index number. 

Dark theme:
selecting
 
Light theme:
selecting2

### LINKED ISSUE #3067: Improve highlighting of currently selected receive address and account
The GUI only highlights the number at the beginning of the row. Example: `#0`. Suggest to highlight the entire row with a different color. This can prevent accidental address reuse. When you hoover over a row it will use a different color scheme. Apply this to the currently selected entry, but with a different color.  
  
Or redesign the interface in such a way that the selected entry is always visible. Like how you can configure a spreadsheet to freeze a specific row that will always be visible even when you scroll.

rating89us:
Closes #2572
Closes #3067

Changes:
- selected account in account page now has highlighted background and text label
- selected address in receive page now has highlighted background and text label

beforeacc1

### DIFF
--- a/pages/Account.qml
+++ b/pages/Account.qml
@@ -190,7 +190,7 @@ Rectangle {
                         height: subaddressAccountListRow.subaddressAccountListItemHeight
                         width: parent ? parent.width : undefined
                         Layout.fillWidth: true
-                        color: "transparent"
+                        color: itemMouseArea.containsMouse || index === currentAccountIndex ? MoneroComponents.Style.titleBarButtonHoverColor : "transparent"
 
                         Rectangle {
                             color: MoneroComponents.Style.appWindowBorderColor
@@ -226,7 +226,7 @@ Rectangle {
 
                             MoneroComponents.Label {
                                 id: nameLabel
-                                color: MoneroComponents.Style.dimmedFontColor
+                                color: index === currentAccountIndex ? MoneroComponents.Style.defaultFontColor : MoneroComponents.Style.dimmedFontColor
                                 anchors.verticalCenter: parent.verticalCenter
                                 anchors.left: idLabel.right
                                 anchors.leftMargin: 6
@@ -276,11 +276,10 @@ Rectangle {
                             }
 
                             MouseArea {
+                                id: itemMouseArea
                                 cursorShape: Qt.PointingHandCursor
                                 anchors.fill: parent
                                 hoverEnabled: true
-                                onEntered: tableItem2.color = MoneroComponents.Style.titleBarButtonHoverColor
-                                onExited: tableItem2.color = "transparent"
                                 onClicked: {
                                     appWindow.currentWallet.switchSubaddressAccount(index);
                                     if (selectAndSend)
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -107,7 +107,7 @@ Rectangle {
                         height: subaddressListRow.subaddressListItemHeight
                         width: parent ? parent.width : undefined
                         Layout.fillWidth: true
-                        color: "transparent"
+                        color: itemMouseArea.containsMouse || index === appWindow.current_subaddress_table_index ? MoneroComponents.Style.titleBarButtonHoverColor : "transparent"
 
                         Rectangle{
                             anchors.right: parent.right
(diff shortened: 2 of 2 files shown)


pages/Account.qml
```diff
@@ -192,6 +192,14 @@ Rectangle {
                         Layout.fillWidth: true
                         color: "transparent"
 
+                        Rectangle {
+                            visible: itemMouseArea.containsMouse || index === currentAccountIndex
+                            height: subaddressAccountListRow.subaddressAccountListItemHeight
+                            width: parent ? parent.width : undefined
+                            Layout.fillWidth: true
+                            color: MoneroComponents.Style.titleBarButtonHoverColor
+                        }
```
xiphon, line 201: Can't we use `tableItem2` `color` instead of introducing another rectangle? Like `color: itemMouseArea.containsMouse || index === currentAccountIndex ? MoneroComponents.Style.titleBarButtonHoverColor : "transparent"`

rating89us: It seems the obvious solution, but I tried this, and it doesn't work.

rating89us: hmm i'll see if it works. just a moment

rating89us: I decided to add a new rectangle because Mouse area has two properties that deal with mouse hover : 
```
onEntered: tableItem2.color = MoneroComponents.Style.titleBarButtonHoverColor
onExited: tableItem2.color = "transparent"
```

and using the following code doesn't work:
```
onEntered: tableItem2.color = (index === currentAccountIndex ? MoneroComponents.Style.titleBarButtonHoverColor : "transparent")
onExited: tableItem2.color = (index === currentAccountIndex ? MoneroComponents.Style.titleBarButtonHoverColor : "transparent")
```

xiphon: > It seems the obvious solution, but I tried this, and it doesn't work.
> I decided to add a new rectangle because Mouse area has two properties that deal with mouse hover and change tableItem2.color value:

Don't add new rectangle. Remove Mouse area `onEntered` and `onExited` handlers.
The approach i suggest works pretty well.

```diff
diff --git a/pages/Account.qml b/pages/Account.qml
index 49c06156..d0bd5e43 100644
--- a/pages/Account.qml
+++ b/pages/Account.qml
@@ -190,7 +190,9 @@ Rectangle {
                         height: subaddressAccountListRow.subaddressAccountListItemHeight
                         width: parent ? parent.width : undefined
                         Layout.fillWidth: true
-                        color: "transparent"
+                        color: itemMouseArea.containsMouse || index === currentAccountIndex
+                            ? MoneroComponents.Style.titleBarButtonHoverColor
+                            : "transparent"
 
                         Rectangle {
                             color: MoneroComponents.Style.appWindowBorderColor
@@ -226,7 +228,9 @@ Rectangle {
 
                             MoneroComponents.Label {
                                 id: nameLabel
-                                color: MoneroComponents.Style.dimmedFontColor
+                                color: index === currentAccountIndex
+                                    ? MoneroComponents.Style.defaultFontColor
+                                    : MoneroComponents.Style.dimmedFontColor
                                 anchors.verticalCenter: parent.verticalCenter
                                 anchors.left: idLabel.right
                                 anchors.leftMargin: 6
@@ -276,11 +280,10 @@ Rectangle {
                             }
 
                             MouseArea {
+                                id: itemMouseArea
                                 cursorShape: Qt.PointingHandCursor
                                 anchors.fill: parent
                                 hoverEnabled: true
-                                onEntered: tableItem2.color = MoneroComponents.Style.titleBarButtonHoverColor
-                                onExited: tableItem2.color = "transparent"
                                 onClicked: {
                                     appWindow.currentWallet.switchSubaddressAccount(index);
                                     if (selectAndSend)
```

rating89us: Yes you're right, I tested it here. I'm just doing some modifications now, will updated new code soon.

rating89us:
removed rectangle and added colors for the 4 possible states

xiphon:
> and added colors for the 4 possible states

I would approve without this change.

rating89us:
Do you prefer this style (1 color + transparent)
beforeacc1

instead of this one (3 colors + transparent)? 
accafter

The second one bloats the code, but I think it's worth it.

xiphon:
I prefer `"transparent" or MoneroComponents.Style.titleBarButtonHoverColor`

rating89us:
ok, reverted changes

xiphon (approved):
Please squash

xiphon:
> ok, reverted changes

Thanks

xiphon (approved):

PR #3165, opened 2020-10-17
Account: remove "Balance:" label; add XMR; align balances

rating89us:
Changes:
- remove "Balance:" label
- add XMR to balances
- align balances

### DIFF
--- a/pages/Account.qml
+++ b/pages/Account.qml
@@ -103,6 +103,7 @@ Rectangle {
 
                 MoneroComponents.TextPlain {
                     id: balanceAll
+                    Layout.rightMargin: 87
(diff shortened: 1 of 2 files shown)


pages/Account.qml
```diff
@@ -136,6 +138,7 @@ Rectangle {
 
                 MoneroComponents.TextPlain {
                     id: unlockedBalanceAll
+                    Layout.rightMargin: 87
```
xiphon, line 141: Should be 80

rating89us: my alignment is correct with 87:
image

jonathancross:
Friendly ping @xiphon

selsta (approved):

PR #3170, opened 2020-10-17
WizardWalletInput: display red border if blank field; add error messages

rating89us:
Changes:
- Display red border for blank wallet name
- Display red border for blank wallet location
- Verify wallet name field when wallet location is updated
- New error messages (Wallet name is empty, Wallet location is empty, Wallet already exists)

On the example below, a wallet file "mywallet" already exists:

errosmessagesinput

### DIFF
--- a/components/LineEdit.qml
+++ b/components/LineEdit.qml
@@ -62,7 +62,7 @@ Item {
 
     property bool borderDisabled: false
     property string borderColor: {
-        if(error && input.text !== ""){
+        if(error){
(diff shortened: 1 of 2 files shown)


selsta:
Note that on the seed input we don’t display a red border when the seed is missing. Which do you prefer?

rating89us:
Wallet name field always starts filled with a default name, while seed field always starts blank.
So I think it's ok displaying red border only when wallet name is missing.

rating89us:
superseded by #3613

Issue #1092, opened 2018-01-23
i18n issue: languages are not flags


pazos:
Hi, first and foremost I would like to say thanks to everybody involved in monero. This seems like a great community to join 👍 So far I've installed monero on Ubuntu and I'm starting to read the documentation.

There is a "little" issue on the GUI that annoys me: [languages are not flags!](http://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/)

I know there are a lot of countries which are tied to a language unspoken in the rest of the world, but is not the case for a bunch of languages (spanish, english or portuguese come to my mind).

So the whole folder [flags](https://github.com/monero-project/monero-gui/tree/master/lang/flags) just doesn't make sense for me and I rather prefer to following these   [practices for representing languages](http://www.flagsarenotlanguages.com/blog/best-practice-for-presenting-languages/)

Please, share your opinion

danrmiller:
These same thoughts come into my mind every time I see a PR about what flag to use for a translation.

erciccione:
I agree we could use a better way to show languages, but i think can be tricky. If we leave only the languages (with no flags) is going to be extremely ugly and confusing. We might use something as described at the end [of this article](http://daily.unitedlanguagegroup.com/stories/editorials/inside-design-language-selector-no-flags). "users are able to select from a variety of supported languages that are found in their chosen country."

pazos:
@erciccione: In the example you linked that makes senses because LandRover.com provides different services for different countries. The same form is used to lookup for localization and language. AFAIK monero-gui has no plans to provide such localization based services.

In the context of a web page this could work for services like "where to buy", but forcing users to choose their country to then be able to select their language doesn't seems a good point for me in a project like monero

erciccione:
Yes it's not optimal, but if we agree that a text only page would be ugly and confusing, i honestly don't see any other viable option for the GUI. Do you have any practical suggestion to improve the present way languages ar shown?

pazos:
Just a quick draft of my suggestion :)

Original:

![monero-welcome-current](https://user-images.githubusercontent.com/975883/35405600-f88cdc2e-0206-11e8-97c8-81cbfdd4698d.png)

New:

![monero-welcome-new](https://user-images.githubusercontent.com/975883/35405615-02676d04-0207-11e8-87fa-032838027a6e.png)


An optional "Please select your language"  between the globe icon and the language selector could be added to.

erciccione:
[...]

PR #3178, opened 2020-10-18
WizardHome: display icon + language instead of change language button

rating89us:
Some users that don't speak English can't understand the meaning of "Change language" button. 

Using icon + language name is more intuitive:
image

### DIFF
--- a/wizard/WizardHome.qml
+++ b/wizard/WizardHome.qml
@@ -30,6 +30,7 @@ import QtQuick 2.9
 import QtQuick.Dialogs 1.2
 import QtQuick.Layouts 1.2
 import QtQuick.Controls 2.0
+import FontAwesome 1.0
 import moneroComponents.NetworkType 1.0
 
 import "../components" as MoneroComponents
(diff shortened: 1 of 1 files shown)


erciccione (changes requested):
Using flags to list languages is not good practice. see http://www.flagsarenotlanguages.com.

We still list flags for languages in some parts of the GUI, but would be better to get rid of those as well.

selsta:
@erciccione what would you suggest if someone does not speak English? While flag are not perfect, it is still an improvement

erciccione:
@selsta It's not really an improvement, because reintroduce a bad practice we partially got rid of some time ago (#1092)

I agree the "change language" button is not optimal, i suggest to use something like https://github.com/monero-project/monero-site/pull/1218. It's common practice and internationally recognized.

rating89us:
Even though I agree with your argument (flag is not language), I prefer to keep the flags
In the future we could retrieve region/country from the OS, because this information is necessary to correctly display date (DD/MM/YYYY vs MM/DD/YYYY), time (1:00 pm vs 13:00) and decimal separator (0,00 vs 0.00).

erciccione:
> In the future we could retrieve region/country from the OS, because this information is necessary to correctly display date 

In that case the usage of flags is correct, because it's not related to a language, but to a country. The current use is not really a good idea, but i'm fine with leaving the flags we still use (we got rid of flags only in the welcome wizard). I wouldn't add new usage of flags as this PR proposes to do though. It's bad practice and it's a step back to the improvements made in past. The symbol used in monero-project/monero-site#1218 would solve the problem of visibility for that button.

xiphon:
> I agree the "change language" button is not optimal, i suggest to use something like [monero-project/monero-site#1218](https://github.com/monero-project/monero-site/pull/1218). It's common practice and internationally recognized.

Looks good to me

rating89us:
updated

xiphon:
Shouldn't it be at the top right corner?

rating89us:
updated (moved to top right corner + smaller font)

xiphon:
Seems vertical alignment is off

rating89us:
updated vertical alignment:
image

xiphon:
image

rating89us:
Updated vertical alignment:
image

rating89us:
maybe we should align with small caps ("onero")
image

xiphon (approved):

wizard/WizardHome.qml
```diff
@@ -56,10 +57,46 @@ Rectangle {
...
+                    fontFamily: FontAwesome.fontFamilySolid
+                    fontColor: MoneroComponents.Style.defaultFontColor
+                    styleName: "Solid"
+                    opacity: 0.85
+                }
+
+                MoneroComponents.TextPlain {
+                    Layout.bottomMargin: 8
+                    font.pixelSize: 14
+                    text: wizard.language_language
```
selsta, line 89:
```suggestion
                    text: wizard.language_language
                    opacity: 0.85
```

Needs initial value

wizard/WizardHome.qml
```diff
@@ -56,10 +57,46 @@ Rectangle {
...
+
+                MoneroComponents.Label {
+                    Layout.bottomMargin: 13
+                    fontSize: 26
+                    text: FontAwesome.language
+                    fontFamily: FontAwesome.fontFamilySolid
+                    fontColor: MoneroComponents.Style.defaultFontColor
+                    styleName: "Solid"
+                    opacity: 0.85
+                }
```
xiphon, line 84: Shouldn't we change icon's opacity on mouse hover?

selsta: And also clickable

rating89us: My initial idea was the same. I tried to put both Icon and language text inside a rectangle and then set a `MouseArea` with `anchors.fill: parent`, but I couldn't get rid of the error "_detected anchors inside item managed by Layout, use Layout.alignment instead_" neither make the MouseArea fill the Layout's width and height (it should overlap/occupy the same space of icon and language text items). So, instead of setting two MouseAreas (one for each item), I decided to set just one for the language text.

xiphon: Add another RowLayout containing both the icon and the text.

xiphon:
https://github.com/monero-project/monero-gui/pull/3210

PR #3282, opened 2020-12-25
Transfer: export all key images by default

### LINKED ISSUE #3246: Can't export key images - always 164 bytes output file
I'm trying to export key images. But I always get a file with size just of 164 bytes.

Tried that using three different wallet files with some balance on them on:
- Laptop #1: Ubuntu 20.10, gui wallet 0.17.1.4
- Laptop #2: Windows 10, gui wallet 0.17.1.4

As mentioned [here](https://github.com/monero-project/monero/issues/7042) "wallet just exports what it's not exported before", and in CLI workaround is to add "all" and use `export_key_images all somefilename `.

In GUI it seems like it's impossible to export key images from scratch.

benevanoff:
Resolves #3246
Depends on monero-project/monero#7185

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -718,7 +718,7 @@ Rectangle {
         selectExisting: false
         onAccepted: {
             console.log(walletManager.urlToLocalPath(exportKeyImagesDialog.fileUrl))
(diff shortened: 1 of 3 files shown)


selsta (approved):

selsta:
PR has to wait until submodule gets branched from master.

selsta (approved):

PR #3286, opened 2020-12-30
workflows: trigger cache miss, fix Windows CI

selsta:
https://github.com/msys2/setup-msys2/issues/99


selsta:
this workaround is apparently not working

PR #3287, opened 2021-01-01
installer: remove high dpi shortcut

selsta:
Qt 5.15 supports high DPI natively


rbrunner7 (approved):
Looks good. Nice to have some progress at the Qt front.

PR #3291, opened 2021-01-05
LeftPanel: allow collapsing of advanced tab

benevanoff:
Currently you have to click on a different tab to hide all the advanced subtabs and this feels clunky, especially if you're on a smaller screen where you then need to scroll in order to see all the tabs when they are uncollapsed. This allows the user to simply reclick on the advanced tab to collapse the subtabs.

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -477,9 +477,18 @@ Rectangle {
                 anchors.right: parent.right
                 text: qsTr("Advanced") + translationManager.emptyString
                 symbol: qsTr("D") + translationManager.emptyString
+                reclickable: true
+                property bool collapsed: false
                 onClicked: {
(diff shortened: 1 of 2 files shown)


selsta:
It would be cool that if you click on "Advanced twice" that the selected value jumps back to the current page.

benevanoff:
@selsta Damn I did that originally actually but then it felt kinda weird so I reverted back to just this.

selsta:
Not sure what the best way is here. If it is weird you can leave it like it is.

selsta:
@benevanoff Any reason for closing this PR?

benevanoff:
@selsta it didn't seem that important and it wasnt getting any other review so i figured i'd just close it instead of just leaving it open forever. I'll reopen it if you think it's worth it.

Issue #3228, opened 2020-11-12
monero-gui 'Daemon log' not using remote node (ever)


alberto-fc:
monero-wallet-gui configured to use a remote node (outside the LAN network) didn't route the commands to the remote node.

You can see that every command entered in the GUI is pushed to the monerod binary (```<monerod binary> <command>```):

```
DEBUG	frontend	src/wallet/api/wallet.cpp:404	sending external cmd:  ("status")
DEBUG	frontend	src/wallet/api/wallet.cpp:404	showing status message
```

```
>>> status
[12/11/2020 1:30] 2020-11-12 00:30:22.809 I Monero 'Oxygen Orion' (v0.17.1.3-release) 
Error: Couldn't connect to daemon: 127.0.0.1:18081
```

As far as I can see, there is no way to command monerod (using arguments nor the config file) to use a remote node for the RPC callings (or anything else, aka ZMQ). So even passing arguments to the monerod binary, the GUI will never read (nor show) the responses from the remote node in the 'Daemon log' dialog.

Another parallel consequence of this, you can delete the monerod binary from the GUI folder (at least using the Window version), the GUI will never show a message about the missing file, just timeout the external command without any trace about the missing binary.

![image](https://user-images.githubusercontent.com/5665863/98880994-4ca60400-2489-11eb-8283-9b56cf6a70e5.png)

xiphon:
`Log` tab is not meant to be used in Remote Node mode, Will be fixed via https://github.com/monero-project/monero-gui/pull/3309 PR.

alberto-fc:
Mice job @luigi1111, simple and elegant. Like it.

Sorry for the extra work but Thank you.

PR #3309, opened 2021-01-15
SettingsLog: disable daemon command input in Remote Node mode 

xiphon:
Closes https://github.com/monero-project/monero-gui/issues/3228

### DIFF
--- a/pages/settings/SettingsLog.qml
+++ b/pages/settings/SettingsLog.qml
@@ -215,6 +215,7 @@ Rectangle {
             Layout.fillWidth: true
             property var lastCommands: []
             property int currentCommandIndex
+            enabled: !persistentSettings.useRemoteNode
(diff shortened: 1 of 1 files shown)


selsta:
Hmm, that also disabled the option to change log level.

xiphon:
> 
> 
> Hmm, that also disables the option to change log level.

Updated, hiding just daemon console and command input now.

xiphon:
Updated. Disables daemon command input in Remote Node mode now.

selsta (approved):

alberto-fc:
This fixed #3228 as-is, what I like is the simplicity of it because remove that confusing option and really you shouldn't be using the GUI to execute Commands in the Remote daemon, even if it is **your** daemon.

Nice and elegant!

PR #3314, opened 2021-01-20
DaemonManager: enable dns block list in simple mode

selsta:


xiphon (approved):

xiphon:
Thinking about Advanced mode with `auto` bootstrap daemon, this will silently enable unwanted dns block list option.

selsta:
Will try to update it to simple mode only

selsta:
Updated to only enable block list when simple mode is used.

xiphon:
Maybe we better do something like
```
diff --git a/main.qml b/main.qml
index 82373963..42452c64 100644
--- a/main.qml
+++ b/main.qml
@@ -1938,12 +1938,13 @@ ApplicationWindow {
             return;
         }

+        const simpleModeFlags = "--enable-dns-blocklist --out-peers 16";
         if (appWindow.daemonRunning) {
             appWindow.stopDaemon(function() {
-                appWindow.startDaemon("")
+                appWindow.startDaemon(simpleModeFlags)
             });
         } else {
-            appWindow.startDaemon("");
+            appWindow.startDaemon(simpleModeFlags);
         }
     }

diff --git a/src/daemon/DaemonManager.cpp b/src/daemon/DaemonManager.cpp
index 86c1a2cc..9a7755fc 100644
--- a/src/daemon/DaemonManager.cpp
+++ b/src/daemon/DaemonManager.cpp
@@ -112,10 +112,6 @@ bool DaemonManager::start(const QString &flags, NetworkType::Type nettype, const
         arguments << "--no-sync";
     }

-    if (!flags.contains("--out-peers", Qt::CaseSensitive) && bootstrapNodeAddress == "auto") {
-        arguments << "--out-peers" << "16";
-    }
-
     arguments << "--check-updates" << "disabled";

     // --max-concurrency based on threads available.
```

selsta:
Looks simpler. Updated.

xiphon (approved):

PR #3319, opened 2021-01-30
2/2 Multisig support

benevanoff:
So this is my attempt at an implementation of specifically the 2/2 multisig scheme. I chose this limitation to begin with because it requires fewer intermediary steps for wallet and transaction creation because there's less secret data to exchange.

This build depends on monero-project/monero#7357

The steps are pretty similar to that of the CLI wallet, just visual. It could probably use some UI/UX improvements later on including a cool icon on the first page of the wizard like there is for the hardware wallet option.

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -65,6 +65,7 @@ Rectangle {
     signal miningClicked()
     signal signClicked()
     signal accountClicked()
+    signal signMultisigClicked()
 
     function selectItem(pos) {
         menuColumn.previousButton.checked = false
@@ -79,6 +80,7 @@ Rectangle {
         else if(pos === "Settings") menuColumn.previousButton = settingsButton
         else if(pos === "Advanced") menuColumn.previousButton = advancedButton
         else if(pos === "Account") menuColumn.previousButton = accountButton
+        else if (pos === "Sign Multisig") menuColum.previousButton = signMultisigButton
         menuColumn.previousButton.checked = true
     }
 
@@ -401,7 +403,7 @@ Rectangle {
                 id: transferButton
                 anchors.left: parent.left
                 anchors.right: parent.right
-                text: qsTr("Send") + translationManager.emptyString
+                text: isMultisig ? qsTr("Create Transaction") + translationManager.emptyString : qsTr("Send") + translationManager.emptyString
                 symbol: qsTr("S") + translationManager.emptyString
                 onClicked: {
                     parent.previousButton.checked = false
@@ -417,6 +419,28 @@ Rectangle {
                 anchors.leftMargin: 20
             }
 
+            // ------------- Sign Multisig Tx tab ---------------
+            MoneroComponents.MenuButton {
+                id: signMultisigButton
+                visible: isMultisig
+                anchors.left: parent.left
+                anchors.right: parent.right
+                text: qsTr("Sign Transaction") + translationManager.emptyString
+                symbol: qsTr("Q") + translationManager.emptyString
+                onClicked: {
+                    parent.previousButton.checked = false
+                    parent.previousButton = signMultisigButton
+                    panel.signMultisigClicked()
+                }
+            }
+
+            MoneroComponents.MenuButtonDivider {
+                visible: accountButton.present
+                anchors.left: parent.left
+                anchors.right: parent.right
+                anchors.leftMargin: 20
+            }
+
             // ------------- AddressBook tab ---------------
 
             MoneroComponents.MenuButton {
@@ -425,7 +449,6 @@ Rectangle {
                 anchors.right: parent.right
                 text: qsTr("Address book") + translationManager.emptyString
                 symbol: qsTr("B") + translationManager.emptyString
-                under: transferButton
                 onClicked: {
                     parent.previousButton.checked = false
                     parent.previousButton = addressBookButton
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -175,9 +175,9 @@ Rectangle {
                     visible: isMultisig
                     text: qsTr("Multisg 2/2") + translationManager.emptyString
                     anchors.top: parent.top
-                    anchors.topMargin: 8
-                    anchors.right: multisigLabel.visible ? multisigLabel.left : parent.right
-                    anchors.rightMargin: 8
+                    anchors.topMargin: 4
+                    anchors.left: multisigLabel.visible ? multisigLabel.right : parent.left
+                    anchors.leftMargin: 8
                     font.pixelSize: 12
                     font.bold: true
                     color: "#4823ff"
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -169,6 +169,20 @@ Rectangle {
                     color: "#ff9323"
                     themeTransition: false
                 }
+
+                MoneroComponents.TextPlain {
+                    id: multisigLabel
+                    visible: isMultisig
+                    text: qsTr("Multisg 2/2") + translationManager.emptyString
+                    anchors.top: parent.top
+                    anchors.topMargin: 8
+                    anchors.right: multisigLabel.visible ? multisigLabel.left : parent.right
+                    anchors.rightMargin: 8
+                    font.pixelSize: 12
+                    font.bold: true
+                    color: "#4823ff"
+                    themeTransition: false
+                }
             }
 
             Item {
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -65,6 +65,7 @@ Rectangle {
     property AddressBook addressBookView: AddressBook { }
     property Keys keysView: Keys { }
     property Account accountView: Account { }
+    property SignMultisig signMultisigView: SignMultisig { }
 
     signal paymentClicked(string address, string paymentId, string amount, int mixinCount, int priority, string description)
     signal sweepUnmixableClicked()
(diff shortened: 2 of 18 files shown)


rbrunner7:
First off all, many thanks to you that you took the initiative to bring multisig to the GUI wallet. It's only 2/2 so far alright, but one has to start somewhere, and maybe as soon as more people get exposed to Monero multisig thanks to this, things will finally pick up speed.

rbrunner7:
A general remark: I have worked quite a lot with Monero multisig, and I can say with conviction: Monero multisig is a bitch. It works if everything is perfectly ok, but if only some detail is wrong, it can fail in many and quite surprising ways. Especially syncing / exchanging key images is often hard to get right.

I would recommend that you prepare for this. For the import and output calls I would propose to add as much useful log output going into the `monero-wallet-gui.log` file as you can, so in case of trouble you have a chance to find out what went wrong.

In my testing, I am just now at such a point: I have successfully created a pair of multisig wallets, but I could not get them to sync so far after a single incoming tx, and no idea yet what might be the problem. One wallet mysteriously became a restore height set that is too high, even higher than current chain tip - no idea how that happened, but of course that disturbs syncing now ...

rbrunner7:
A general question about the GUI wallet, I am not so well versed testing it: Is there a way, e.g. by using command-line parameters, to set log-levels, so some log messages from `wallet2` go into `monero-wallet-gui.log`? I would love to see right now what `wallet2` is saying during the import of the key images that seemingly has no effect whatsoever.

rbrunner7:
I just had my second wallet getting a wrong restore height, 1664050, whereas the current chain tip is 1649830. I am not aware that I did anything in the GUI wallet that could cause this. Not sure this is connected with these multisig additions, of course.

rbrunner7:
I try now to import a key image file that the GUI wallet wrote in the CLI wallet, to be better able to see what may go wrong. It refuses to do so however, with an error message of "Error: Failed to import multisig info: Bad multisig info file magic in".

During this I noticed that this version of the GUI wallet seems to write its key image files as simple hex strings, whereas the CLI wallet writes a different file format if the option `export-format` is set to `ascii`. Here an example of such a file:

```
-----BEGIN MoneroAsciiDataV1-----
TW9uZXJvIG11bHRpc2lnIGV4cG9ydAHs6ENqyeRn5C98VSqm051z1ygV3cTEG3FG
wgPNTEYvt4kB0q9faAAROIjSdVqCFb2C/+5EAkJ33wX1nFTi8/ZoLz1972NyZ430
AuFwCJo3mQ4BlZlW6w6kyubU+HJ1MgZe98ZbwWfoXMQumrKXp8EhM3mxxcBAqdU/
Cix7ZeP4nN8T0mADnszGXkUbMVU5/J7+WNKhzfQ6w2djZsBl9IgR69h5RgCW6qun
nvCCscYeixCr+C2o+obA2lzKmFYjAVKwjxEtYRt+HzojzBIgy0ys5bj6OonHOFIN
B1Heo3IVF8pWo4IhVoutZ/bLOSfG6Hend+ES0u5Q7yUzrOQRwxQ8RT5awy4sfT0g
9QphmEI9vEZmth296T7Jk4yY5ypFTIYyZgvNVbJPmhgEBA==
-----END MoneroAsciiDataV1-----
```

Such a file format incompatibility is most unfortunate. I guess those key image file functions that you implemented in the GUI wallet are "hexing" their input and output themselves. I propose to replace this by programmatically setting the wallet option `export-format` to ascii and then let `wallet2` read and write the files in a compatible textual format.

rbrunner7:
If you create a multisig wallet it's possible to just click "Next" on the multisig keys entry screen, without entering any info from the other signer. I think the button should only get clickable after you entered any info because otherwise you can too easily end up with a normal wallet and no way to go back to that screen again, at least as far as I could see.

Of course a way to call up that multisig keys entry screen later as well, with any existing simple wallet, would be nice nevertheless.

rbrunner7:
I have now a wallet where the sync was somehow successful, seemingly after several attempts. I verified this with the help of the CLI wallet which does **not** give the warning about partial key images anymore. If I re-open the wallet in the GUI wallet, it however still has that message and the import and export buttons for key images.

Maybe the call to get the info from `wallet2` fails somehow, because I see in the wallet's log the following message to times after opening the wallet:

    WARNING	frontend	src/wallet/api/wallet.cpp:412	qrc:/pages/Transfer.qml:167: TypeError: Cannot call method 'hasMultisigPartialKeyImages' of undefined

I confess that I compiled with Qt 5.11.3 instead of the recommended Qt 5.9.7, which introduces some element of incertainty here about the source of the error. But on the other hand, I did not yet find any other function of the GUI wallet that does not run with that version - everything else seems to work.

benevanoff:
Hey @rbrunner7 thanks for taking a look! I've pushed changes to monero-project/monero#7357 that should resolve the partial key image file incompatibility between the CLI and GUI. I kinda doubt the restore thing height has anything to do with my changes but it's worth an investigation nonetheless. I'm also getting some wonkyness with the partial key image warning so I'll have to look into that more.

rbrunner7:
I finally came around to have a look at your improvements. However somehow I can't get the new code to compile after updating the Monero code. Any idea about this error?

```
/var/git/monero-core/src/libwalletqt/PendingTransaction.cpp: In member function ‘QString PendingTransaction::address() const’:
/var/git/monero-core/src/libwalletqt/PendingTransaction.cpp:98:51: error: ‘struct Monero::PendingTransaction’ has no member named ‘dest_addresses’
     std::vector<std::string> addresses = m_pimpl->dest_addresses();
                                                   ^~~~~~~~~~~~~~
make[3]: *** [src/CMakeFiles/monero-wallet-gui.dir/build.make:367: src/CMakeFiles/monero-wallet-gui.dir/libwalletqt/PendingTransaction.cpp.o] Error 1
```

benevanoff:
@rbrunner7 sorry! a code snippet got missed when staging the commit. I've pushed the necessary changes to the monero branch

rbrunner7:
I finally managed to go once through the whole process with two 2/2 multisig wallets managed by two instances of the GUI wallet app and finally saw all new dialogs supporting this: Building the wallets, funding, syncing, transacting, syncing again, and transacting again. So far, so good.

The file format issue is much improved, the files being compatible now with the CLI wallet (I tested, it works). Even better still would be additionally switching on the "save as ASCII" setting in the wallets after creating them. In the CLI wallet you handle this with the help of the `set` command, and the name of the setting is `export-format`. It's much, much easier to transport / send those file if they are not binary, but textual.

rbrunner7:
The handling of the `multisig_tx` files right now is very intransparent for the user. There is an almost total lack of feedback. The wallet does not tell you where it writes that file, and how it's named. If there is already a `multisig_tx` file, it gets overwritten without any confirmation, which seems problematic to me.

If you use the `Sign Transaction` button it signs alright if it finds the file, but there is no feedback whether signing was successful or not or even whether it was attempted at all - I don't get an error if no file is there. If you click the `Submit Transaction` button without a tx loaded and without a file present the GUI wallet just terminates: Puff, vanished :)

rbrunner7:
The *Partial key images (?)* message string is completely useless in my wallets when I test because it's simply always displayed. It's there even immediately after wallet creation where there can't possibly be any partial key images yet, for total lack of any outputs in the wallet. For testing I had to ignore that message and fly blind as far as key images are concerned.

I guess you don't have the same problem? No idea how to provoke the problem if that display is working for you, or failing only intermittently ...

benevanoff:
Thanks again for the review @rbrunner7. I've finally gotten around to addressing these issues. Now you should be able to choose where the partially signed transaction is stored, a success dialog pops up on signing and buttons are disabled when appropriate. The partial key image banner should work right now too but I'm honestly not 100% sure.

rbrunner7:
I tested today with the latest code.

Frankly, the results leave me a little exasperated. Whereas earlier the warning about partial key images always displayed, I don't see it anymore at all now. If I try to make a transaction with partial key images I get an error message of "No unmixable outputs to sweep" which does not make sense at all. The new file choose dialogs for loading and saving are nice, but confusingly some of them open in the directory of the binary, while others open in the directory of the wallets. The files still save as binary, which is a pity regarding UX; if it's currently not possible to save them in ASCII please at least detail the reasons. The wallet still crashes / vanishes if I try to sign a transaction a second time.

At that point I called it a day.

These are all errors that I was able to see during only half an hour or so of testing, and I do wonder whether we have a communication / coordination problem (maybe I test too early while you are still working on those things), whether your testing process does not cover enough, or whether my system and your system really have such a different behaviour and I run into difficulties that are simply absent while you test.

rbrunner7:
As good news I finally found out what called those mysterious "restore height in the future" problems that I had that made testing very hard for me: I use testnet wallets, and it turned out that currently the GUI wallet is not able to correctly create such wallets, not even normal / non-multisig ones. I made a PR to correct the problem:

https://github.com/monero-project/monero/pull/7682

Thus clearly nothing to do with your code, and we can forget about that problem here.

benevanoff:
@rbrunner7 Thanks for testing for me again. I’m just gonna mark this as a draft since it clearly needs more work

benevanoff:
As far as the file format goes, I’d like to add the ability for the user to choose between a QR code displayed on the screen, plain text displayed to the screen (to be copy/pasted or whatever), or as a binary file. So I’m leaving as is until then.

rbrunner7:
> As far as the file format goes, I’d like to add the ability for the user to choose between a QR code displayed on the screen, plain text displayed to the screen (to be copy/pasted or whatever), or as a binary file. So I’m leaving as is until then.

Sounds interesting. Just a remark: If you continue to transact, key images will easily reach 10 KB or more. Also transactions themselves can become quite big, if they happen to collect many unspent outputs. It might need some realistic tests how far you can get here with QR codes.

benevanoff:
Thank you, I didn't think of that. Perhaps the data can get split up into a set of QR codes. Some sort of limit will probably need to be imposed still.

benevanoff:
Hmmm, you're right the partial key image banner is still not functioning correctly. However, I still don't have a problem transacting if the key images are actually properly sync'd (which I can tell from the console output).

rbrunner7:
> Hmmm, you're right the partial key image banner is still not functioning correctly. However, I still don't have a problem transacting if the key images are actually properly sync'd (which I can tell from the console output).

Little misunderstanding: The strange error message I mentioned was produced with an attempt to start a transaction while **not** synced. This of course should either be not possible at all, e.g. with inactive buttons, or then result in a sensible error message :)

benevanoff:
Ah I see, ty!

PR #3325, opened 2021-02-02
Tx confirmation multiple addresses support

xiphon:
Only the second commit is a subject for review.

monero-tx-confirmation-multiple-recipients

### DIFF
--- a/MiddlePanel.qml
+++ b/MiddlePanel.qml
@@ -51,7 +51,7 @@ Rectangle {
     property alias flickable: mainFlickable
 
     property Transfer transferView: Transfer {
-        onPaymentClicked: root.paymentClicked(address, paymentId, amount, mixinCount, priority, description)
+        onPaymentClicked: root.paymentClicked(recipients, paymentId, mixinCount, priority, description)
         onSweepUnmixableClicked: root.sweepUnmixableClicked()
     }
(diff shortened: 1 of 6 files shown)


rating89us:
I think we should display somewhere a destinations count. It could be a numbered list on the destination addresses, like this:
1. Monero Address - 0.01 XMR
2. Monero Address - 0.01 XMR
3. John Doe - 0.02 XMR

rating89us:
Destinations address field is using a different font

xiphon:
> I think we should display somewhere a destinations count. It could be a numbered list on the destination addresses, like this:
> 
>     1. Monero Address - 0.01 XMR
> 
>     2. Monero Address - 0.01 XMR
> 
>     3. John Doe - 0.02 XMR

Updated.

> Destinations address field is using a different font

Yep, using Roboto monospace font. That's intended.

rating89us:
I think it is also important to add fiat conversion after each XMR amount in To: field

Example: 1.00 XMR (150 USD)

xiphon:
> 
> 
> I think it is also important to add fiat conversion after each XMR amount in To: field
> 
> Example: 1.00 XMR (150 USD)

Updated

xiphon:
Rebased

selsta (approved):

Issue #740, opened 2017-05-24
Send to multiple addresses in the gui


fresheneesz:
The GUI should be able to send to multiple addresses. Is it possible to send with multiple addresses as inputs? If so, that should be added too

medusadigital:
sounds like a reasonable request to me. also simply becasue its one of the basic features in the cli wallet.

solving this UX wise seems like another task. might have effect on:

- send tab
- confirmation pop up
- transaction history details

opinions welcome

chpio:
any news on this one?

ankurss:
would like to see this implemented too, please check how its implemented in electrum or other btc gui clients

i send payments daily to 20+ address using btc with single transaction, would switch to xmr if this is implemented

ankurss:
@medusadigital @Jaqueeee can you implement this in coming version ?

binaryFate:
Would be nice to have that by now.

SamsungGalaxyPlayer:
It's feasibly possible, but I would like to see some UI mockups before recommending this feature's inclusion. I'm concerned this will clutter the UI.

fresheneesz:
Here's a UI mock for ya:

![monero multi-address ui](https://user-images.githubusercontent.com/149531/52931292-b4550c80-3300-11e9-9b6d-b991b186411f.PNG)

binaryFate:
Looks good but not sufficient: their amounts might be different and should be displayed too. A drop-down type of display could display more.

fresheneesz:
Great point. Here's another try:

![monero multi-recipient transactions ui](https://user-images.githubusercontent.com/149531/54497964-3879c000-48be-11e9-9379-337276306cf0.PNG)

SamsungGalaxyPlayer:
I'd love to see some UI mockups for this in the latest versions. The simple mode should not support this feature imo, but the advanced/expert mode should :)

selsta:
This could be helpful for this issue: https://doc.qt.io/qt-5/qtqml-javascript-dynamicobjectcreation.html

selsta:
I’m currently working on this: https://streamable.com/bzw7p

@GKBS I’d need your design help on this :)

selsta:
I’m currently working on this: https://streamable.com/bzw7p

@GBKS I’d need your design help on this :)

GBKS:
[...]

PR #3332, opened 2021-02-05
Transfer: implement sending to multiple recipients 

xiphon:
Design is inspired by https://github.com/monero-project/monero-gui/issues/740 discussion. Thanks to all the participants.

Closes https://github.com/monero-project/monero-gui/issues/740.

Requires https://github.com/monero-project/monero-gui/pull/3325.

Only the last commit is a subject for review.

monero-send-to-multiple-recipients

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -26,6 +26,7 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import QtQml.Models 2.2
 import QtQuick 2.9
 import QtQuick.Controls 1.4
 import QtQuick.Layouts 1.1
@@ -60,19 +61,18 @@ Rectangle {
         }
 
         // There are sufficient unlocked funds available
-        if (walletManager.amountFromString(amountLine.text) > appWindow.getUnlockedBalance()) {
+        if (recipientModel.getAmountTotal() > appWindow.getUnlockedBalance()) {
             return qsTr("Amount is more than unlocked balance.") + translationManager.emptyString;
         }
 
-        if (addressLine.text)
-        {
+        if (!recipientModel.hasEmptyAddress()) {
             // Address is valid
-            if (!TxUtils.checkAddress(addressLine.text, appWindow.persistentSettings.nettype)) {
+            if (recipientModel.hasInvalidAddress()) {
                 return qsTr("Address is invalid.") + translationManager.emptyString;
             }
 
             // Amount is nonzero
-            if (!amountLine.text || parseFloat(amountLine.text) <= 0) {
+            if (recipientModel.hasEmptyAmount()) {
                 return qsTr("Enter an amount.") + translationManager.emptyString;
             }
         }
@@ -93,10 +93,16 @@ Rectangle {
     }
 
     function fillPaymentDetails(address, payment_id, amount, tx_description, recipient_name) {
-        addressLine.text = address
-        setPaymentId(payment_id);
-        amountLine.text = amount
-        setDescription((recipient_name ? recipient_name + " " : "") + tx_description);
+        if (recipientModel.count > 0) {
+            const last = recipientModel.count - 1;
+            if (recipientModel.get(recipientModel.count - 1).address == "") {
+                recipientModel.remove(last);
+            }
+        }
+
+        recipientModel.newRecipient(address, Utils.removeTrailingZeros(amount || ""));
+        setPaymentId(payment_id || "");
+        setDescription((recipient_name ? recipient_name + " " : "") + (tx_description || ""));
     }
 
     function updateFromQrCode(address, payment_id, amount, tx_description, recipient_name) {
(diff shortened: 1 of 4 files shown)


rating89us:
Very nice. Some feedback:
1. Fee estimative should receive a "Fee" title in white text (the same used in Fee priority). Fee value should be aligned below total amount.

image

2. Both inline buttons should stay inside the address field, and should be displayed only when it's empty (to save space for the address).

image

3. I miss fiat estimation of each destination, even though it is displayed in send confirmation dialog). But I understand we lack space here. Maybe we could list all amounts at the end (see below). Example:
```
Total 
         0.047 XMR
         (0.01 + 0.03 + 0.007 XMR)
         7.22 USD
         (1.54 + 4.61 + 1.07 USD)
```
If there are more than 3, just break the line

```
Total 
         0.077 XMR
         (0.01 + 0.03 + 0.007
         0.01 + 0.01 + 0.01 XMR)
         11.84 USD
         (1.54 + 4.61 + 1.07
          1.54 + 1.54 + 1.54 USD)
```

selsta:
> I miss fiat estimation of each destination, even though it is displayed in send confirmation dialog. 

I think that would be a bit overkill, there is already a lot going on here with multiple recipients.

pages/Transfer.qml
```diff
@@ -167,184 +163,448 @@ Rectangle {
...
+                                }
+                                validator: RegExpValidator {
+                                    regExp: /^\s*(\d{1,8})?([\.,]\d{1,12})?\s*$/
+                                }
+                            }
+
+                            MoneroComponents.TextPlain {
+                                Layout.preferredWidth: recipientLayout.thirdRowWidth
+                                horizontalAlignment: Text.AlignHCenter
+                                font.family: MoneroComponents.Style.fontRegular.name
```
selsta, line 424:
```suggestion
                                font.family: FontAwesome.fontFamilySolid
                                font.styleName: "Solid"
```

selsta (commented):
I did not get OpenAlias to work yet.

GBKS:
Looks really good.

The inline scrolling of recipients in the modal is not that great. Maybe the default could be to just red "3 recipients", and clicking that would expand the full list (which would then also expand the modal)? That would keep the initial modal size small and still allow users to more conveniently review every single address without the cramped scroll interaction. Just an idea.

xiphon:
@GBKS 

> Maybe the default could be to just red "3 recipients"

Current implementation displays 2 recipients with no scrolling.

> clicking that would expand the full list (which would then also expand the modal)?

It is ~1000px height list in 16 recipients case. I'm concerned it will look good.

pages/Transfer.qml
xiphon: Fixed

xiphon:
> I did not get OpenAlias to work yet.

Fixed.

GBKS:
> It is ~1000px height list in 16 recipients case. I'm concerned it will look good.

It is a tricky one. Easy to come up with some overly complex solutions for this, but let's just stick with your approach then.

xiphon:
>     1. Fee estimative should receive a "Fee" title in white text and should be aligned below total amount.

Will give it a try in a while.

xiphon:
Updated estimated fee layout.

monero-gui-send-multiple-recipients-fee

xiphon:
>     1. Both inline buttons should stay inside the address field, and should be displayed only when it's empty (to save space for the address).

Considered this option as well. I don't like hiding some controls based on the user's input. IMO the current implementation looks better.

I'd like to hear other guys opinions on this.

rating89us:
image
Good, but Fee lines have too much vertical spacing

rating89us:
> IMO the current implementation looks better.

I always like to think about the function of each button, imagining a label explaining its action. Currently it's not easy for a user to know what these two buttons are going to do.

Currently it seems that these two buttons have two functions:
- "Add an address" (it adds an address when last field is empty)
- "Add a new recipient" (it creates a new recipient when the last field is not empty)

When inside the address field, both have a single function:
- "Add an address to this field" (it adds an address when this field is empty)

xiphon:
> Good, but Fee lines have too much vertical spacing

image

selsta:
Needs rebase.

xiphon:
Rebased

selsta:
> https://user-images.githubusercontent.com/26060097/107953419-046c1580-6f93-11eb-8ef1-6d4347c2d533.png

Minor comment, while it is technically correctly aligned with the text on the left side, it looks wrong in my eyes, especially on a large window. @GBKS any opinions on spacing here?

GBKS:
Not an easy layout issue. The fee and total texts are all a bit floaty. How about bringing in a little more structure with these containers (see below)? That way there's a clear relationship what context each bit of information belongs to. Not totally clean, TBH.

monero-send-multiple-recipients-gbks-210323

xiphon:
Looks like we're risking to end up in endless discussion about minor and not that important PR part. 

Reverted all fee layout changes i made during the review back to the original state.

Once we merge the PR, anyone is free to experiment with fee layout, open the discussion on topic, propose the changes or even get 'em implemented in a separate PR later.

selsta (approved):
ok, any further design changes can be done later

selsta:
> How about bringing in a little more structure with these containers (see below)?

If we add fee again later it probably should be done this way, looks cleaner.

Issue #2087, opened 2019-04-16
Support database pruning


cdiv1e12:
Refer *existing* users (those with a non-pruned blockchain) to the `prune-blockchain` tool. Add some information inside the GUI on how to use this tool. New users would prune their blockchain by default, which means: add `--prune-blockchain` to the daemon startup flags.

To get pruned status a RPC call is required: `prune_blockchain` with `check` argument set to `true` will return a bool `pruned`.

SamsungGalaxyPlayer:
I support automatically pruning the blockchain by default. It's the usability tradeoff that is appropriate for most people.

maogo:
I objected strongly  automatically pruning . As for me，  I  dont want to pruning.

maogo:
and  should  encourage  users  do that, if they have the conditions .

dEBRUYNE-1:
Automatic pruning by default would only be enabled on `Simple mode (bootstrap mode)` as far as I know.

ITwrx:
While i agree that most people would choose pruning, i don't prefer it were enabled by default. What i would like to see, is the option given during the setup "wizard", and it would be nice if one could get info about all available flags (in the gui where the daemon flag input is). The fact that i had to do a github issues search, after a failing web search, and not finding it in the pdf manual included with the download, is an aggravating user experience. Advanced mode shouldn't mean the info is not available in the GUI at all, IMO. It should be that it's just more low level stuff most people won't want to deal with, at least at first.

erciccione:
@ITwrx 

> What i would like to see, is the option given during the setup "wizard"

That sounds like a good option. We could make the user check "pruned" or "normal", maybe leaving "pruned" checked by default.

> would be nice if one could get info about all available flags

I guess the best way to do it would be to redirect to existing documentation with these info. I wouldn't put a long list of flags on that page.

> not finding it in the pdf manual included with the download

FWIW, there is an old open issue about adding that: https://github.com/monero-ecosystem/monero-GUI-guide/issues/149

ITwrx:
> I guess the best way to do it would be to redirect to existing documentation with these info. I wouldn't put a long list of flags on that page.
yes, i like tooltips and/or links

selsta:
Simple mode (bootstrap) now pruned automatically.

It is also possible to select pruning during wizard if a new chain is setup.

PR #3345, opened 2021-03-01
Blockchain pruning support for new databases

benevanoff:
Addresses #2087

Wallets set up in simple mode will use a pruned node by default going forward. Pruning is optional and turned off by default in advanced mode.

### DIFF
--- a/wizard/WizardDaemonSettings.qml
+++ b/wizard/WizardDaemonSettings.qml
@@ -94,6 +94,28 @@ ColumnLayout {
             }
         }
 
+        RowLayout {
+            id: pruningOptionRow
+            MoneroComponents.CheckBox {
+                id: pruneBlockchainCheckBox
+                checked: !existingDbWarning.visible ? persistentSettings.pruneBlockchain : false
+                enabled: !existingDbWarning.visible
+                onClicked: {
+                    persistentSettings.pruneBlockchain =  !persistentSettings.pruneBlockchain
+                    this.checked = persistentSettings.pruneBlockchain
+                }
+                text: qsTr("Prune blockchain") + translationManager.emptyString
+            }
+
+            Text {
+                id: existingDbWarning
+                text: "A blockchain database already exists here. Select a new location to start a pruned node"
+                visible: daemonManager ? daemonManager.checkLmdbExists(blockchainFolder.text) : false
+                color: MoneroComponents.Style.defaultFontColor
+                font.family: MoneroComponents.Style.fontRegular.name
+            }
+        }
+
         ColumnLayout{
             Layout.topMargin: 6
             spacing: 0
--- a/wizard/WizardModeSelection.qml
+++ b/wizard/WizardModeSelection.qml
@@ -119,6 +119,7 @@ Rectangle {
 
                 onMenuClicked: {
                     if(appWindow.persistentSettings.nettype == 0){
+                        appWindow.persistentSettings.pruneBlockchain = true;
                         applyWalletMode(1, 'wizardModeBootstrap');
                     }
                 }
@@ -139,6 +140,7 @@ Rectangle {
                 imageIcon: "qrc:///images/local-node-full.png"
 
                 onMenuClicked: {
+                    appWindow.persistentSettings.pruneBlockchain = false; // can be toggled on next page
                     applyWalletMode(2, 'wizardHome');
                 }
             }
--- a/main.qml
+++ b/main.qml
@@ -724,7 +724,7 @@ ApplicationWindow {
 
         const noSync = appWindow.walletMode === 0;
         const bootstrapNodeAddress = persistentSettings.walletMode < 2 ? "auto" : persistentSettings.bootstrapNodeAddress
-        daemonManager.start(flags, persistentSettings.nettype, persistentSettings.blockchainDataDir, bootstrapNodeAddress, noSync);
+        daemonManager.start(flags, persistentSettings.nettype, persistentSettings.blockchainDataDir, bootstrapNodeAddress, noSync, persistentSettings.pruneBlockchain);
     }
 
     function stopDaemon(callback, splash){
@@ -1410,6 +1410,7 @@ ApplicationWindow {
         property bool checkForUpdates: true
         property bool autosave: true
         property int autosaveMinutes: 10
+        property bool pruneBlockchain: false
 
         property bool fiatPriceEnabled: false
         property bool fiatPriceToggle: false
--- a/src/daemon/DaemonManager.cpp
+++ b/src/daemon/DaemonManager.cpp
@@ -27,6 +27,7 @@
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "DaemonManager.h"
+#include "common/util.h"
 #include <QElapsedTimer>
 #include <QFile>
 #include <QMutexLocker>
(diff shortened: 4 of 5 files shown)


selsta:
There is a problem here. If someone has a local daemon and starts simple mode the daemon will silently wait ages during start to prune the existing blockchain.

benevanoff:
@selsta I've changed this PR to first check if the currently selected DB folder already exists before using the `--prune-blockchain` flag which should prevent this issue.

wizard/WizardDaemonSettings.qml
```diff
@@ -94,6 +94,13 @@ ColumnLayout {
             }
         }
 
+        MoneroComponents.CheckBox {
+            id: pruneBlockchainCheckBox
+            checked: persistentSettings.pruneBlockchain
+            onClicked: persistentSettings.pruneBlockchain = true
```
selsta, line 100: Is it possible to uncheck it again? Seems like this checkbox can only be set to true.

selsta (commented):
If you check prune blockchain during the wizard nothing will happen if a chain already exists. There should be a warning there else it might be confusing.

wizard/WizardDaemonSettings.qml
benevanoff: Oops, thanks

benevanoff:
@selsta thanks, updated accordingly

wizard/WizardDaemonSettings.qml
```diff
@@ -92,6 +92,35 @@ ColumnLayout {
...
+            }
+        }
+
+        RowLayout {
+            id: pruningOptionRow
+            MoneroComponents.CheckBox {
+                id: pruneBlockchainCheckBox
+                checked: persistentSettings.pruneBlockchain
+                onClicked: {
+                    if (daemonManager.validateDataDir(persistentSettings.blockchainDataDir).lmdbExists) {
```
selsta, line 106: As far as I can see persistentSettings.blockchainDataDir is only set if a custom blockchain data dir is used, so if someone has a blockchain in the default path this does not work.

benevanoff: gotcha

wizard/WizardDaemonSettings.qml
```diff
@@ -92,6 +92,35 @@ ColumnLayout {
...
+                    persistentSettings.pruneBlockchain =  !persistentSettings.pruneBlockchain
+                    this.checked = persistentSettings.pruneBlockchain
+                }
+                text: qsTr("Prune blockchain") + translationManager.emptyString
+            }
+
+            Text {
+                id: existingDbWarning
+                text: "A blockchain database already exists here. Select a new location to start a pruned node"
+                visible: daemonManager.checkLmdbExists(persistentSettings.blockchainDataDir)
```
selsta, line 120: daemonManager can be undefined / null. Same change at other places daemonManager is used.

```suggestion
                visible: daemonManager ? daemonManager.checkLmdbExists(persistentSettings.blockchainDataDir) : false
```

wizard/WizardDaemonSettings.qml
```diff
@@ -92,6 +92,35 @@ ColumnLayout {
...
+            onTextUpdated: {
+                existingDbWarning.visible = daemonManager.checkLmdbExists(persistentSettings.blockchainDataDir)
+            }
+        }
+
+        RowLayout {
+            id: pruningOptionRow
+            MoneroComponents.CheckBox {
+                id: pruneBlockchainCheckBox
+                checked: persistentSettings.pruneBlockchain
```
selsta, line 104:
```suggestion
                checked: persistentSettings.pruneBlockchain
                enabled: !existingDbWarning.visible
```

selsta (commented):
This does not work on my Mac yet, did not figure out why (will investigate later). Left some comments.

benevanoff:
Strange, I went back and double checked with `check_blockchain_pruning` after spinning up a new DB from the create new wallet wizard in a custom set folder on my Windows machine and it works fine. You're sure you're on the most up to date revision? Some of those commits in between are broken (oops...)

wizard/WizardModeSelection.qml
```diff
@@ -91,6 +91,7 @@ Rectangle {
 
                 onMenuClicked: {
                     if(appWindow.persistentSettings.nettype == 0){
+                        appWindow.persistentSettings.pruneBlockchain = true;
```
xiphon, line 94: ?

wizard/WizardDaemonSettings.qml
```diff
@@ -92,6 +92,34 @@ ColumnLayout {
...
+                    persistentSettings.pruneBlockchain =  !persistentSettings.pruneBlockchain
+                    this.checked = persistentSettings.pruneBlockchain
+                }
+                text: qsTr("Prune blockchain") + translationManager.emptyString
+            }
+
+            Text {
+                id: existingDbWarning
+                text: "A blockchain database already exists here. Select a new location to start a pruned node"
+                visible: daemonManager ? daemonManager.checkLmdbExists(persistentSettings.blockchainDataDir) : false
```
xiphon, line 119: Should use `blockchainFolder.text`, not `persistentSettings.blockchainDataDir`

wizard/WizardModeSelection.qml
benevanoff: Prune by default for mainnet simple mode

selsta: Simple mode does not sync a chain so it does not make sense to prune here.

benevanoff: Ah you’re right

selsta: So only necessary for simple mode (bootstrap)

wizard/WizardModeSelection.qml
```diff
@@ -119,6 +119,7 @@ Rectangle {
...
```
xiphon, line 122: Switch from `Advanced mode` to `Simple mode (bootstrap)` and immediately switch back to `Advanced mode`. You have pruning enabled.

wizard/WizardDaemonSettings.qml
```diff
@@ -92,6 +92,34 @@ ColumnLayout {
...
+                text: qsTr("Prune blockchain") + translationManager.emptyString
+            }
+
+            Text {
+                id: existingDbWarning
+                text: "A blockchain database already exists here. Select a new location to start a pruned node"
+                visible: daemonManager ? daemonManager.checkLmdbExists(blockchainFolder.text) : false
+                color: MoneroComponents.Style.defaultFontColor
+                font.family: MoneroComponents.Style.fontRegular.name
+            }
```
xiphon, line 122: Showing the warning using `appWindow.showStatusMessage` on `pruneBlockchainCheckBox` click will probably look better.

benevanoff: mmmm i don't think so. I tried that and it honestly looked horrible. plus then you have to enable the check box even if you dont really want the user to click it

wizard/WizardModeSelection.qml
benevanoff: got it

xiphon (approved):

wizard/WizardDaemonSettings.qml
```diff
@@ -92,6 +92,34 @@ ColumnLayout {
                     blockchainFolder.focus = true;
                 }
             }
+            onTextUpdated: {
+                var dbExists = daemonManager ? daemonManager.checkLmdbExists(blockchainFolder.text) : false
+                existingDbWarning.visible = dbExists
+                if (dbExists)
+                    pruneBlockchainCheckBox.checked = false;
+            }
```
selsta, line 100:
```suggestion
```

Seems redundant, works without this code here too.

existingDbWarning.visible should get updated once blockchainFolder.text changes, and pruneBlockchainCheckBox.checked also gets updated once existingDbWarning.visible changes.

selsta:
Ok, don’t know if it was one of the recent changes but I got it to work on Mac too.

selsta (approved):

PR #3347, opened 2021-03-02
DaemonManagerDialog use preferred color scheme

benevanoff:
The white window with black text looks out of place when using the dark theme. Instead use the default colors for the currently selected theme.


selsta (approved):
Thank you.

PR #3362, opened 2021-03-20
Added support for more fiat currencies.

### LINKED ISSUE #3351: Displaying currency in GBP?
Hello,

The mockup at https://github.com/monero-project/monero-gui/issues/1484 shows the currency being translated into GBP via CoinMarketCap, however in my wallet (`0.17.1.9-3ca5f10f (Qt 5.15.2)`) I only have the option of USD or EUR from one of the following sources:

   * Kraken
   * CryptoCompare
   * Coingecko

Given that at least Coingeko supports XMR/GBP, is there anything in particular I need to do to get this showing in my UI?

0xc1f1:
Added support for more fiat currencies and removed Kraken as it had too few currencies that where supported. A scrollbar will still need to be added since the current drop down list is too long to display all the currency alternatives.

Sorry for doing another PR. I am new to git and I made some mistakes.

Fixes #3351

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1281,7 +1281,8 @@ ApplicationWindow {
             }
             return resp["monero"][key];
         } else if(url.startsWith("https://min-api.cryptocompare.com/data/")){
-            var key = currency === "xmreur" ? "EUR" : "USD";
+			var key = currency.substring(3);
+
             if(!resp.hasOwnProperty(key)){
                 appWindow.fiatApiError("cryptocompare API has error(s)");
                 return;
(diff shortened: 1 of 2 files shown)


selsta (commented):
Hi, thanks for the PR. Two issues, first this is a lot of code duplication, which can theoretically be solved in a simpler way.

Also I don't like that Kraken is being removed, I personally use Kraken as my exchange so I also want it for fiat conversion.

I will see if I can find a solution for this to add a bit more currencies without huge code duplication.

selsta:
Thanks again for the PR, https://github.com/monero-project/monero-gui/pull/3852 seems to have similar functionality with less code duplication so I will close this here.

PR #3363, opened 2021-03-20
Renamed Norwegian translation file to the correct name.

### LINKED ISSUE #3353: Norwegian language is not working
A user reported they can select the language on the GUI, but all the content appears in English. [Norwegian is almost completely done](https://github.com/monero-project/monero-gui/blob/master/translations/monero-core_nb_NO.ts) and should show up correctly.

fsrw42:
It seems like the name was incorrect since the Norwegian translation in the program did not function with that file name.

Fixes #3353

### DIFF
--- a/translations/monero-core_nb.ts
+++ b/translations/monero-core_nb.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="nb_NO">
+<TS version="2.1" language="nb">
 <context>
     <name>Account</name>
     <message>


erciccione:
Thanks for the PR. Have you tested it locally to make sure that that's indeed the problem?

fsrw42:
> Thanks for the PR. Have you tested it locally to make sure that that's indeed the problem?

Yes I have. It seems to work just as intended.

selsta (approved):
Thank you.

erciccione (approved):
For the records, this language files has a different name because was generated automatically by Weblate. I'll make sure to generate new languages manually next ime.

PR #3364, opened 2021-03-21
Makefile: Add ppc64le target

JeremyRand:
The `README.md` build instructions for Linux don't work as advertised on ppc64le systems, because `x86-64` is hardcoded in the Makefile target that the instructions say to use.  This PR adds a new target that uses `ppc64le` instead, and amends the build instructions to clarify that the `release` target is specific to `x86-64`.

I can confirm that this fixes the build for me on my Talos II Workstation.

Thanks to whoever maintains the Void PowerPC packages of Monero for helping me figure out what was wrong and how to fix it.


selsta (approved):

JeremyRand:
Is anything blocking this from a merge?

selsta:
No, I forgot about it. Added it to merge list now.

Issue #3226, opened 2020-11-11
[feature] Add "offline" option for "Node" tab.


jonathancross:
Currently, users must specify one of 2 options "Local node" or "Remote node":

![xmr-node-screenshot](https://user-images.githubusercontent.com/5115470/98841978-2ceeeb80-2449-11eb-98c3-305166d64267.png)

Unfortunatly neither option is appropriate for an offline system used for air-gapped signing.

Users who are offline are currently prompted with various error messages and attempts to start the daemon.

Request:
1. Make an "Offline" option available with explanation that it will have not attempt to make network connections.
2. Prevent wallet from making network connections.
3. Bottom left should indicate that user is in "Offline" mode if explicitly selected.

Related: #3225

PR #3378, opened 2021-04-05
daemon manager: add remote node and offline options

benevanoff:
I thought this seemed like a good way to add an offline mode as suggested in #3226.

daemon_options

Depends on [these changes](https://github.com/benevanoff/monero/commit/25e82545f35b5bfe8beb17220c232fb89411d4da) to core but I'm blocked by the filter rn :/


selsta:
My worry with this is that it will confuse people if suddenly three buttons show up with a timer.

benevanoff:
Maybe scrap the timer and just design a new startup page or popup with options and information?

PR #3385, opened 2021-04-09
workflows: cache android docker build

selsta:
As far as I can see both Linux and Windows cache are less than 1GB. Github allows 5GB cache per repo so it should be fine to cache android too.


xiphon (approved):

selsta:
Did not get it to work :/

PR #3393, opened 2021-04-14
main: drop unused settings

selsta:
v0.17.2.1 is out


main.qml
```diff
@@ -1376,22 +1376,18 @@ ApplicationWindow {
...
         property bool transferShowAdvanced: false
         property bool receiveShowAdvanced: false
         property bool historyShowAdvanced: false
         property bool historyHumanDates: true
         property string blockchainDataDir: ""
         property bool useRemoteNode: false
-        property string remoteNodeAddress: "" // TODO: drop after v0.17.2.0 release
         property string remoteNodesSerialized: JSON.stringify({
                 selected: 0,
                 nodes: remoteNodeAddress != ""
```
xiphon, line 1393: Have to fix this as well. Just store empty array by default.

selsta: Updated.

selsta:
I guess it might be a good idea to wait a while before including this in a release, not everyone does every single update.

xiphon (approved):
> I guess it might be a good idea to wait a while before including this in a release, not everyone does every single update.

Yep

PR #3400, opened 2021-04-14
cmake: fix ccache detection on mac

selsta:
Seeing that FindCcache gets imported in `add_subdirectory(monero)`, I don't even know if it is necessary for this file to be duplicated.


selsta:
Will remove it on GUI side instead.

PR #3408, opened 2021-04-17
WizardModeSelection: update modes descriptions (+pruned blockchain)

rating89us:


selsta (commented):
In general I'm ok with the text changes but I feel like the text is a bit overly detailed.

A simple mode user doesn't have to know what pruning is.

For advanced mode I would also remove the info about pruning.

PR #3409, opened 2021-04-17
SettingsLayout: add "Tor" to "Socks5 proxy"

rating89us:
Not everyone knows that Socks5 proxy is used by Tor


selsta:
I'm not sure about this. Socks5 itself has nothing to do with Tor. Tor is one use case for this. The GUI guide is better for this.

plowsof:
Agree, should close this one. I2p is another one for example.

Issue #3243, opened 2020-11-20
Remove "Copy" Button Next to QR Code


livinginformation:
After some discussion with Selsta in #monero-dev (and with an affected user), I propose that we remove the "Copy to Clipboard" button that appears next to the scannable QR code, due to it being both redundant and confusing.

![gui_copy](https://user-images.githubusercontent.com/3836500/99842258-592efa00-2b24-11eb-9ceb-14a53ca067d8.png)

The first button copies the wallet address as expected (437KEQM2VQYjbkbCZk1UWpBo5RsdpawhudDVSzsYK31rSxYnTdQWq6dTno7wrejs2p8yN1hChGxHbEcBUTKZ16GbDedGa2w), whereas the second button copies the  QR-ified version of the address (monero:437KEQM2VQYjbkbCZk1UWpBo5RsdpawhudDVSzsYK31rSxYnTdQWq6dTno7wrejs2p8yN1hChGxHbEcBUTKZ16GbDedGa2w). 

The QR-ified version of the wallet address apparently causes errors when entered into exchanges which don't sanity-check wallet addresses in the browser (bitfinex specifically in this case). It is not immediately apparent to the user that the copied address contains an extra prefix (that isn't even useful if you aren't scanning a QR), which is a UX concern.

livinginformation:
Alternatively, removing the prefix would be acceptable as well. The buttons are still kind of redundant, but at least we won't have users with stuck exchange withdrawals.

lucas-wade:
having the copy button right next to the qr code is really nice for new users. its very direct. the little clipboard icon is not clear whats going on. the problem is just that the copy to clipboard button copies in the "monero:" text. I think the copy button should remain next to the qr image and the "monero:" text should be removed from the copied text.

selsta:
@lucas-wade see #3415

PR #3415, opened 2021-04-19
Receive: bring back large copy button (without uri)

selsta:
Suggested here: https://github.com/monero-project/monero-gui/issues/3243#issuecomment-781109208

I kinda agree that having a large copy button is more user friendly, especially on the receive page. Any other opinions?

### DIFF
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -301,6 +301,18 @@ Rectangle {
             MoneroComponents.StandardButton {
                 Layout.preferredWidth: 220
                 small: true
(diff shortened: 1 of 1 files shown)


xiphon:
I wouldn't add it back. Bloats the UI and the page will have two controls doing exactly the same: per address copy button and this one.

Copy button next to QR code image should either copy image data to clipboard or the QR code contents (Payment URL).

If a user wants to copy an address, there are per address copy buttons.

selsta:
I don't have any strong opinions on either. Closing if you don't like it.

PR #3437, opened 2021-04-24
main: remove minor switch/case code duplication

selsta:


main.qml
```diff
@@ -534,12 +534,6 @@ ApplicationWindow {
             // try to resolve common wallet cache errors automatically
             switch (wallet.errorString) {
                 case "basic_string::_M_replace_aux":
```
xiphon, line 536: Would be nice to check whether we still need them at all.

xiphon (approved):

PR #3441, opened 2021-04-25
Revert "Animation slow down for debugging"

selsta:
This reverts commit baff7f213f07b7ae134062cd54b22c691f48fbe4.

This is broken without private Qt headers. With qmake this was easy to add (`gui-private`), I did not figure out how to do it with cmake so I will revert this instead of having broken debug build.


selsta:
#3442 can be used alternatively if we want to keep animation slowdown in.

xiphon (approved):
Yep, let's just strip the code.

PR #3449, opened 2021-04-30
workflows: update cache action, invalidate cache

selsta:
It might be a good idea to invalidate the cache every time we update the workflows file. `-1-` can be bumped to `-2-` next time.


selsta:
Alternatively we can also add `ENV version 1`  into the docker file and update that every time we want to invalidate all caches.

Which do you prefer @xiphon ?

PR #3450, opened 2021-04-30
workflows: checkout latest commit instead of master

selsta:
Seeing that we upload the resulting Mac binary it probably is a good idea to checkout a known commit instead of master.


xiphon (approved):

selsta:
#3452

PR #3452, opened 2021-05-01
workflows: replace qt download script for macos-bundle

selsta:
Advantages over the currently used script:

- script now automatically retries on download / connection failure (this has happened multiple times in the past now)
- script verifies sha256sum of qt download

Second one is important as we upload the binaries.

I can move the script to an external repo but inside the .github folder it shouldn't annoy anyone.


perfect-daemon:
correct

perfect-daemon (approved):
correct

PR #3483, opened 2021-05-17
WizardModeSelection: disable simple mode on Tails

selsta:


selsta:
@mj-xmr please review

mj-xmr (changes requested):
Could you rather do:

```
if (isTails) // in future: || other such OS || (...)
    "Not available on tails"
else
{
   if (net == 0)
      "Easy"
   else 
       "Only on mainnet"
}
```
?

This would make the logic more cohesive.

Other than that, a good idea.

selsta:
@mj-xmr updated, also added a variable for `simpleModeAvailable`

mj-xmr (approved):

PR #3490, opened 2021-05-19
Implement tooltips in multiple pages and components

rating89us:


selsta:
Really nice, thanks :) Will review soon.

rating89us:
ok, it's ready for review

selsta:
We could also add a delay + opacity effect.

```patch
diff --git a/components/Tooltip.qml b/components/Tooltip.qml
index f2e71def..b60b007e 100644
--- a/components/Tooltip.qml
+++ b/components/Tooltip.qml
@@ -67,7 +67,7 @@ Rectangle {
         }
     }
 
-    Popup {
+    ToolTip {
         id: popup
 
         background: Rectangle {
@@ -84,7 +84,14 @@ Rectangle {
         y: tooltipBottom
             ? (tooltipIconVisible ? icon.y + height : parent.y + parent.height + 2)
             : (tooltipIconVisible ? icon.y - height : parent.y - tooltip.height - 20)
+        enter: Transition {
+            NumberAnimation { property: "opacity"; from: 0.0; to: 1.0; duration: 150 }
+        }
 
+        exit: Transition {
+            NumberAnimation { property: "opacity"; from: 1.0; to: 0.0; duration: 150 }
+        }
+        delay: 200
         RowLayout {
             Text {
                 id: tooltip
```

Note that this patch breaks Transfer -> Advanced Options tooltips, I did not figure out why yet.

handwavingdrivendevelopment:
Will these Tooltips also work with taps on a tablet/mobile device or do they only work on mouseenter/leave events?

rating89us:
> Will these Tooltips also work with taps on a tablet/mobile device 

Currently they are working if you tap and hold on Android, but I don't think they are really useful on mobile, because sometimes you can't read the tooltip since your finger stays on top of it

rating89us:
> We could also add a delay + opacity effect.

Done. 

> Note that this patch breaks Transfer -> Advanced Options tooltips, I did not figure out why yet.
Actually all tooltips where affected. It was just a matter of setting a height to the ToolTip component.

components/AdvancedOptionsItem.qml
```diff
@@ -22,6 +22,7 @@ RowLayout {
         MoneroComponents.Label {
             id: title
             fontSize: 14
+            tooltipIconVisible: true
```
selsta, line 25: What is this used for? I wasn't able to figure it out.

pages/Transfer.qml
```diff
@@ -483,12 +486,16 @@ Rectangle {
                                 horizontalAlignment: Text.AlignHCenter
                                 opacity: mouseArea.containsMouse ? 1 : 0.85
                                 text: recipientModel.count == 1 ? FontAwesome.infinity : FontAwesome.times
+                                tooltip: recipientModel.count == 1 ? qsTr("Send all unlocked balance of this account") :  qsTr("Remove recipient")  + translationManager.emptyString
```
selsta, line 489:
```suggestion
                                tooltip: recipientModel.count == 1 ? qsTr("Send all unlocked balance of this account") : qsTr("Remove recipient")  + translationManager.emptyString
```

components/AdvancedOptionsItem.qml
rating89us: When tooltipIconVisible = true we display the question mark icon that when hovered turns tooltip visible:
image

When tooltipIconVisible = false (default), we display only the tooltip box.

selsta: Ah I looked on the wrong page :)

selsta:
Please squash and then I can approve it.

selsta:
Where is the checkbox tooltip used?

rating89us:
I'm going to use it in WizardCreateDevice1.qml of PR #3508. Decided to include here in this PR because it's implementing tooltips in multiple components.

selsta:
ok please squash

selsta (approved):

PR #3500, opened 2021-05-24
Some fixes to Android UI

rating89us:
- add upper bar
- fix spacing in transfer page
- menu on LeftPanel closes when an item is selected
-


handwavingdrivendevelopment:
Nice work!

Do you mind posting screenshots of:
- landscape mode default 
- hamburger menu open in portrait as well as landscape mode 

Why is there a large space between 'receive' and 'transactions'? Is this just the width of the underlying buttons?

pages/Transfer.qml
```diff
@@ -366,6 +366,9 @@ Rectangle {
                                 spacing: 0
                                 wrapMode: Text.WrapAnywhere
                                 placeholderText: {
+                                    if(isAndroid) {
+                                      return "4.. / 8..";
+                                    }
```
selsta, line 371: Should this be inside the next if? Otherwise it will always display `4.. / 8..` even on testnet / stagenet.

pages/SharedRingDB.qml
```diff
@@ -293,81 +293,56 @@ Rectangle {
...
-                            }
-                        }
-                    }
-                }
+            MoneroComponents.LineEdit {
+                id: getRingLine
+                Layout.fillWidth: true
+                fontSize: 16
+                labelFontSize: 14
+                labelText: qsTr("Get ring") + ":" + translationManager.emptyString
```
selsta, line 305: Can we decide on one way to spell it? Label has "ring", button has "Ring".

UpperPanel.qml
```diff
@@ -0,0 +1,274 @@
...
+
+import "./components" as MoneroComponents
+import "./components/effects/" as MoneroEffects
+
+import moneroComponents.Clipboard 1.0
+import moneroComponents.Wallet 1.0
+import moneroComponents.WalletManager 1.0
+import moneroComponents.TransactionHistory 1.0
+import moneroComponents.TransactionHistoryModel 1.0
+import "../js/TxUtils.js" as TxUtils
```
selsta, line 44: A lot of these imports can be removed.

UpperPanel.qml
```diff
@@ -0,0 +1,274 @@
...
+
+
+
+Rectangle {
+    id: upperPanel
+    color: MoneroComponents.Style.blackTheme ? "#222222" : "#EEEEEE"
+    property var previousPage
```
selsta, line 51:
```suggestion
```

Unless I did not figure out what this is used for.

selsta (commented):
It might be a good idea to use isMobile instead of isAndroid because it is also possible to compile for iOS.

This will also allow us to differentiate between mobile and tablet.

handwavingdrivendevelopment:
On the isAndroid vs isMobile thing, it may be good at some point (if this doesn't already happen somewhere) to specifically define the screen sizes into views like: isSmall, isMedium, isLarge or isMobile, isTablet, isDesktop, based on some design system standard (not sure if Monero GUI uses one) https://material.io/design/layout/responsive-layout-grid.html#breakpoints for example, and use those for the regular visual changes that happen at certain breakpoints, and reserve isAndroid, or isIOS for platform specific things.

MoroccanMalinois (commented):
Tested on phone with Android 11, this significantly improves UI/UX

PR #3508, opened 2021-05-26
Display instructions when can't connect to hardware wallet

### LINKED ISSUE #3507: Ledger Nanox : handle locked device gracefully
Error message from "Please enter password for: [wallet name]" screen:
> Couldn't open wallet: Wrong Device Status: 0x6e00 (SW_CLA_NOT_SUPPORTED), EXPECTED 0x9000 (SW_OK), MASK 0xffff

Monero app version: `1.7.6` (latest Firmware)

Related: #2762

2021-05-26_nanox_locked_error_gui

Ideally we would just ask the user to unlock the device using their PIN and ensure the Monero app is open.

rating89us:
Fixes #3507

I tested only on Trezor.


wizard/WizardCreateDevice1.qml
```diff
@@ -201,10 +201,94 @@ Rectangle {
...
+            if(restoreHeight.text){
+                // Parse date string or restore height as integer
+                if(restoreHeight.text.indexOf('-') === 4 && restoreHeight.text.length === 10){
+                    _restoreHeight = Wizard.getApproximateBlockchainHeight(restoreHeight.text, Utils.netTypeToString());
+                } else {
+                    _restoreHeight = parseInt(restoreHeight.text)
+                }
+                wizardController.walletOptionsRestoreHeight = _restoreHeight;
+            }
+            wizardController.walletCreatedFromDevice.connect(onCreateWalletFromDeviceCompleted);
```
selsta, line 277:
```suggestion
```

This code duplication seems unnecessary.

wizard/WizardCreateDevice1.qml
```diff
@@ -201,10 +201,94 @@ Rectangle {
...
+    }
+
     Component.onCompleted: {
         errorMsg.text = "";
     }
 
+    function showHardwareWalletAsk() {
+        if (typeof message != 'undefined') {
+            splash.messageText = message
+        }
```
selsta, line 230:
```suggestion
```

main.qml
```diff
@@ -66,6 +66,8 @@ ApplicationWindow {
     property bool ctrlPressed: false
     property alias persistentSettings : persistentSettings
     property string accountsDir: !persistentSettings.portable ? moneroAccountsDir : persistentSettings.portableFolderName + "/wallets"
+    property bool firstHardwareWalletConnection: true
```
selsta, line 69: This will always be true on every start. Is this intended?

main.qml
```diff
@@ -529,23 +531,37 @@ ApplicationWindow {
...
-                    appWindow.showStatusMessage(qsTr("Repairing incompatible wallet cache. Resyncing wallet."),6);
-                    return;
-                default:
-                    // opening with password but password doesn't match
-                    console.error("Error opening wallet with password: ", wallet.errorString);
-                    passwordDialog.showError(qsTr("Couldn't open wallet: ") + wallet.errorString);
-                    console.log("closing wallet async : " + wallet.address)
-                    closeWallet();
-                    return;
+            if(wallet.errorString == "Could not connect to the device Trezor" || wallet.errorString == "Wrong Device Status: 0x6e00 (SW_CLA_NOT_SUPPORTED), EXPECTED 0x9000 (SW_OK), MASK 0xffff") {
```
selsta, line 534: 0x6e00 is not the only possible Ledger error status.

main.qml
```diff
@@ -1090,6 +1106,52 @@ ApplicationWindow {
         }
     }
 
+    function showHardwareWalletDialog(path, hardwareWalletType) {
+        if (typeof message != 'undefined') {
+            splash.messageText = message
+        }
```
selsta, line 1112:
```suggestion
```

wizard/WizardCreateDevice1.qml
```diff
@@ -201,10 +201,94 @@ Rectangle {
...
+            hardwareWalletAsk.imgSource = "";
+            if (hardwareWalletType == "Ledger") {
+                hardwareWalletAsk.messageInstructions1 = qsTr("Check if your Ledger has the Monero app installed (available in Ledger Live) and is running it.") + translationManager.emptyString;
+                hardwareWalletAsk.messageInstructions2 = qsTr("Check if your Ledger has the most recent firmware version (available on Manager section of Ledger Live).") + translationManager.emptyString;
+            } else {
+                hardwareWalletAsk.messageInstructions1 = qsTr("Check if the USB cable is connected. Push the cable into your Trezor until you hear/feel a click.") + translationManager.emptyString;
+                hardwareWalletAsk.messageInstructions2 = qsTr("Check if your Trezor has firmware v%1 or higher (available on Trezor Wallet or Trezor Suite).").arg(trezorFirmwareVersion) + translationManager.emptyString;
+            }
+            hardwareWalletAsk.okButtonText = qsTr("Try again") + translationManager.emptyString;
+        }
```
selsta, line 259: Seems like a lot of code duplication here that can be avoided.

wizard/WizardCreateDevice1.qml
rating89us: This is a copy of onNextClicked of WizardNav in Line 171. I first tried to call it, but it returns a "not a function" error.

main.qml
rating89us: Yes. When `true`, this means the user has not yet seen the first message ("please connect device and unlock it"). After the user sees this message for the first time, this becomes `false` and the second message (troubleshooting) is displayed the next time connection fails.

main.qml
rating89us: Can you inform me what are the error codes when 1) ledger is locked and 2) ledger app is not open?

main.qml
selsta: This is not a persistent variable. Every time the GUI gets restarted this will display.

rating89us: Sure, and that's intended. A better name for this variable would be `firstHardwareWalletConnectionOnThisSession`.

selsta: Not sure it is necessary to display instructions during every session but I did not test the PR yet.

rating89us: The idea here is to display a nice message every time the user opens a wallet file of a hardware wallet, but trezor/ledger device is locked. 

I usually forget to unlock my trezor before clicking on the wallet in "open wallet" page, and then I get this error message "Could not connect to the device Trezor" that doesn't inform that the wallet is locked.

Ideally we should know before opening the wallet file that it is a hardware wallet file, and then we display this nice reminder that the user should unlock the hardware wallet first. However, it is not possible to see if the wallet file is from a hardware wallet before opening and decrypting it. 

So it works this way:
Open wallet file -> can't connect to hardware wallet -> display nice message to remind user to unlock it (on first try) -> `firstHardwareWalletConnection ` set to `false ` -> can't connect again -> display troubleshooting message

selsta: Ok, I might have misunderstood it, like I said, I did not test it. So the message only shows up if you fail. This is fine then.

PR #3512, opened 2021-05-27
main: hide/show popups before/after inactivity screen lock

### LINKED ISSUE #3511: Inactivity screen lock covers Transaction confirmation dialog
Test case:

1. Perform transaction
2. Stay at confirmation dialog
3. Have screen lock engage

Observed:
1. The lock screen fades the main screen
2. The confirmation dialog stays readable
3. The screen lock components are drawn on top of confirmation dialog
4. Both the confirmation components and screen lock components can be interacted with, e.g. buttons can be pressed.

Expected (I suppose):
1. The screen lock covers and blurs everything.
2. Only the screen lock can be interacted with.
3. When unlocked, the user is returned to the confirmation dialog.

image

rating89us:
fixes #3511

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -2212,6 +2212,8 @@ ApplicationWindow {
         if(!persistentSettings.lockOnUserInActivity) return;
         if(passwordDialog.visible) return;
         var inputDialogVisible = inputDialog && inputDialog.visible
+        var successfulTxPopupVisible = successfulTxPopup && successfulTxPopup.visible
+        var informationPopupVisible = informationPopup && informationPopup.visible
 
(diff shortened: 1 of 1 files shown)


components/SuccessfulTxDialog.qml
```diff
@@ -55,6 +55,7 @@ Rectangle {
     Clipboard { id: clipboard }
 
     property var transactionID;
+    property alias transactionID: root.transactionID
```
selsta, line 58:
```suggestion
```

selsta (approved):

PR #3519, opened 2021-06-02
Transfer: display contact name when address is stored in address book

rating89us:
image


selsta (commented):
I like this design wise. Two bugs I noticed:

Importing from address book adds a third row instead of entering the address in this case.

Screenshot 2021-06-02 at 16 52 37

Backtab does not work in this situation:

Screenshot 2021-06-02 at 16 52 01

rating89us:
> Importing from address book adds a third row instead of entering the address in this case. (EDIT: this is unrelated to this PR and exists in current code too)

Import from address book button always try to import to the last recipient line and ignores blank spaces on previous recipient lines. If it's not possible, it will insert a new line (I didn't change this behavior of address book button in PR #3498).

> Backtab does not work in this situation

Currently backtab ignores the address field when it contains an address book contact. I prefer it this way. You shouldn't be able to edit the text inserted in address field.

selsta:
> Currently backtab ignores the address field when it contains an address book contact. I prefer it this way. You shouldn't be able to edit the text inserted in address field.

Did you test the situation from the screenshot?

rating89us:
Yes. I get this result, and I think it's ok:
backtab

rating89us:
Do you think the contact name should be selected?

selsta:
> Do you think the contact name should be selected?

No it's correct in your GIF, backtab didn't work at all for me but now I can't reproduce it. Will have to play around to see if I can get into this state again.

selsta:
Now I was able to trigger it again, see how it skips one of the amount fields during backtab:

loop

rating89us:
It's not skipping for me in Ubuntu, with the same fields. Not sure why it's triggered on your computer.

selsta:
It also does not skip for me on Mac if I restart the GUI. Will write if I figure out how to trigger it.

rating89us:
You could also try something like this to see if the problem is fixed:
```
MoneroComponents.LineEditMulti {
    id: addressInput
    KeyNavigation.backtab: index > 0 ? recipientRepeater.itemAt(index - 1).amountInput : sendButton
```

jonathancross:
@rating89us  This looks like a great feature, is it still being worked on?

PR #3520, opened 2021-06-02
History: clean search bar inline button

rating89us:
image

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -154,10 +154,10 @@ Rectangle {
                 Layout.fillWidth: true
                 input.topPadding: 6
                 input.bottomPadding: 6
-                fontSize: 16
+                fontSize: 15
                 labelFontSize: 14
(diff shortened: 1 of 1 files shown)


pages/History.qml
```diff
@@ -170,6 +170,21 @@ Rectangle {
                         root.updateFilter();
                     }
                 }
+
+                MoneroComponents.InlineButton {
+                    Layout.topMargin: -8
+                    Layout.rightMargin: -8
+                    Layout.leftMargin: 8
```
selsta, line 177:
```suggestion
                    Layout.leftMargin: -2
```

rating89us: I think it's necessary enough blank space after the last character, so the user can click at the end of the search bar content with the mouse. By using a small left margin it's easy to accidentally click the clean search button.

selsta: But even if you click on the right side it does not put you to the end of the text.

rating89us: ok, no problem

selsta:
Please squash

selsta (approved):

PR #3522, opened 2021-06-02
History: move cursor to end of search input field after importing add…

rating89us:
…ress


selsta:
Would it make more sense to move to amount field?

rating89us:
This is in transactions page, there is no amount field.

selsta:
Oops, was testing the "address book" PR so I was confused.

selsta (approved):

PR #3532, opened 2021-06-06
DatePicker: highlight selected date/today; hide invalid days

rating89us:
`styleData.valid = true` means the date is between minimum and maximum dates of calendar component. Since we're not setting minimum and maximum dates yet (will be implemented in another PR), `styleData.valid` will always return `true`.

### DIFF
--- a/components/DatePicker.qml
+++ b/components/DatePicker.qml
@@ -305,11 +305,6 @@ Item {
                             id: dayRect
                             anchors.fill: parent
                             radius: parent.implicitHeight / 2
-                            color: {
-                                if(dayArea.pressed && styleData.visibleMonth)
-                                    return MoneroComponents.Style.blackTheme ? "#20FFFFFF" : "#10000000"
-                                return "transparent";
-                            }
                         }
 
                         MoneroComponents.TextPlain {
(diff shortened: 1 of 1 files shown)


components/DatePicker.qml
```diff
@@ -40,6 +40,7 @@ Item {
     id: datePicker
     readonly property alias expanded: popup.visible
     property date currentDate
+    property var today: new Date();
```
selsta, line 43:
```suggestion
```

components/DatePicker.qml
```diff
@@ -327,19 +323,32 @@ Item {
...
+                                  }
+                              } else {
+                                  if (dayArea.containsMouse) {
+                                      dayRect.color = MoneroComponents.Style.blackTheme ? "#20FFFFFF" : "#10000000"
+                                  } else {
+                                      dayRect.color = "transparent";
+                                  }
+                              }
+                              if(!styleData.valid) return "transparent"
+                              if(styleData.date.toDateString() === today.toDateString()) return "#FFFF00"
```
selsta, line 340:
```suggestion
                              if(styleData.date.toDateString() === (new Date()).toDateString()) return "#FFFF00"
```

selsta (approved):

PR #3533, opened 2021-06-06
DatePicker: support mouse wheel to navigate months

rating89us:
Go to previous/next month with mouse wheel (a lot faster than clicking on left/right arrows)

### DIFF
--- a/components/DatePicker.qml
+++ b/components/DatePicker.qml
@@ -271,6 +271,11 @@ Item {
 
             MouseArea {
                 anchors.fill: parent
+                scrollGestureEnabled: false
+                onWheel: {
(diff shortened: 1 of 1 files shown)


components/DatePicker.qml
```diff
@@ -271,6 +271,10 @@ Item {
 
             MouseArea {
                 anchors.fill: parent
+                onWheel: {
```
selsta, line 274:
```suggestion
                scrollGestureEnabled: false
                onWheel: {
```

It does not work well on trackpad so best to disable it for trackpads.

selsta (approved):

PR #3542, opened 2021-06-07
SettingsLog: ScrollBar always visible

rating89us:


selsta:
Why? Other scrollbars are also not always on. On macOS at least an always visible scrollbar looks out of place.

rating89us:
Depending on the log text, it may seem that there isn't any additional text to the top: 
image

selsta:
> It may seem that there isn't any additional text to the top

But isn't that always the case with hidden scrollbars? Please disable it on Mac, there are no always visible scrollbars on Mac.

rating89us:
About policy property in https://doc.qt.io/qt-5/qml-qtquick-controls2-scrollbar.html#policy-prop:
> This property was introduced in QtQuick.Controls 2.2 (Qt 5.9).

Strange, because we are importing QtQuick.Controls 2.0 in SettingsLog.qml and it's working fine for me.

selsta:
You can still set it to 2.2, we support 5.9 as minimum.

selsta (approved):

PR #3543, opened 2021-06-07
SettingsLog: fix text color bug on theme switch; new warning/green text colors

### LINKED ISSUE #3475: Theme Switch Bug
_**Settings**_ 🠖 **_Log_**: The text of the Daemon log does not change color when changing from Light to Dark Theme and vice versa.

image

rating89us:
Closes #3475

Changes:
- Fix text color bug on theme switch (#3475)
- New warning text color (orange darker than previous orange, for better contrast on light theme)
- New green text color (dark green, for better contrast on light theme)

### DIFF
--- a/pages/settings/SettingsLog.qml
+++ b/pages/settings/SettingsLog.qml
@@ -132,6 +132,18 @@ Rectangle {
             font.pixelSize: 18
             font.family: MoneroComponents.Style.fontRegular.name
             text: qsTr("Daemon log") + translationManager.emptyString
+            themeTransition: false
+            onColorChanged: {
+                var flickableContentYBefore = flickable.contentY
+                var daemonLogText = consoleArea.text
+                consoleArea.clear();
+                if (MoneroComponents.Style.blackTheme) {
+                    consoleArea.append(daemonLogText.replace(/#000000/g, '#ffffff').replace(/#008000/g, '#00ff00'));
+                } else {
+                    consoleArea.append(daemonLogText.replace(/#ffffff/g, '#000000').replace(/#00ff00/g, '#008000'));
+                }
+                flickable.contentY = flickableContentYBefore
+            }
         }
 
         Item {
@@ -164,7 +176,7 @@ Rectangle {
                     wrapMode: TextEdit.Wrap
                     readOnly: true
                     function logCommand(msg){
-                        msg = log_color(msg, "lime");
+                        msg = log_color(msg, MoneroComponents.Style.blackTheme ? "lime" : "green");
                         consoleArea.append(msg);
                     }
                     function logMessage(msg){
@@ -173,7 +185,7 @@ Rectangle {
                         if(msg.toLowerCase().indexOf('error') >= 0){
                             color = MoneroComponents.Style.errorColor;
                         } else if (msg.toLowerCase().indexOf('warning') >= 0){
-                            color = MoneroComponents.Style.warningColor;
+                            color = "#fa6800"
                         }
 
                         // format multi-lines


selsta:
> New green text color (dark green, for better contrast on light theme)

This has noticeably worse contrast in dark theme.

pages/settings/SettingsLog.qml
```diff
@@ -132,6 +132,20 @@ Rectangle {
...
+                var flickableContentYBefore = flickable.contentY
+                var daemonLogText = consoleArea.text
+                consoleArea.clear();
+                if (MoneroComponents.Style.blackTheme) {
+                    var daemonLogTextSwitchedTextColor = daemonLogText.replace(/#000000/g, '#ffffff');
+                } else {
+                    var daemonLogTextSwitchedTextColor = daemonLogText.replace(/#ffffff/g, '#000000');
+                }
+                consoleArea.append(daemonLogTextSwitchedTextColor);
+                flickable.contentY = flickableContentYBefore
```
selsta, line 147:
```suggestion
                if (MoneroComponents.Style.blackTheme) {                       
                    consoleArea.text = consoleArea.text.replace(/#000000/g, '#ffffff');   
                } else {                                                            
                    consoleArea.text = consoleArea.text.replace(/#ffffff/g, '#000000');   
                }                                                              
```

rating89us: I tried this approach before, but it doesn't work. After switching theme, text area scrolls up (scrollbar position doesn't persist) and every time you type a command (like 'status'), it scrolls up again. That's why I used .clean() and then .append() on consoleArea and reconfigured    flickable.contentY.

selsta: Oh, I didn't test this. Will recheck.

rating89us:
> This has noticeably worse contrast in dark theme. Maybe two different colors depending on the theme?

fixed

pages/settings/SettingsLog.qml
rating89us: this fixes the scroll bar, but this bug is still present: every time you type a command (like 'status'), it scrolls up again.

rating89us: read this
https://doc.qt.io/qt-5/qml-qtquick-textedit.html#text-prop

> "The append(), insert() and remove() methods provide more fine-grained control and remarkably better performance for modifying especially large rich text content."

selsta: This code is in `onColorChanged` which will rarely get called so performance does not matter here. Apart from that append, insert, remove being faster in not valid if you clear it first.

selsta: > every time you type a command (like 'status'), it scrolls up again.

It does not for me. Did you test my exact code?

loop

rating89us: Yes, I used your code. Did you switch theme before typing status?

rating89us: logtext
This is what happens when I type status, after switching theme.

selsta: Yes, can reproduce now. Don't know a better solution than your code.

pages/settings/SettingsLog.qml
```diff
@@ -132,6 +132,20 @@ Rectangle {
...
             font.family: MoneroComponents.Style.fontRegular.name
             text: qsTr("Daemon log") + translationManager.emptyString
+            themeTransition: false
+            onColorChanged: {
+                console.log("color changed")
+                var flickableContentYBefore = flickable.contentY
+                var daemonLogText = consoleArea.text
+                consoleArea.clear();
+                if (MoneroComponents.Style.blackTheme) {
+                    var daemonLogTextSwitchedTextColor = daemonLogText.replace(/#000000/g, '#ffffff');
```
selsta, line 142: If we want to do it perfectly we would also have to fix the different lime / green color on change.

pages/settings/SettingsLog.qml
```diff
@@ -164,7 +178,11 @@ Rectangle {
                     wrapMode: TextEdit.Wrap
                     readOnly: true
                     function logCommand(msg){
-                        msg = log_color(msg, "lime");
```
selsta, line 167: Ternary operator would be more readable here IMO and more compact, but that's just personal style.

```
msg = log_color(msg, MoneroComponents.Style.blackTheme ? "lime" : "green")
consoleArea.append(msg);
```

or even in one line but that's pushing readability :D

```
consoleArea.append(log_color(msg, MoneroComponents.Style.blackTheme ? "lime" : "green"));
```

pages/settings/SettingsLog.qml
rating89us: yeah, I spent a long time trying to fix this, but this was the only solution I found.

selsta (approved):

PR #2637, opened 2019-12-19
AddressBook, Transfer: disallow line break in description field

tobtoht:
closes #2613

PR #3545, opened 2021-06-09
LineEdit: use ColumnLayout instead of Item

rating89us:
Required for #2637

Currently LineEdit uses an Item{} component while LineEditMulti uses a ColumnLayout{}. 

This PR modifies LineEdit component to use ColumnLayout{}. This is useful because we can set its properties (like margins) with the same values of a neighbor LineEditMulti component.

This modification requires some changes to LineEdit components in Transfer and WizardWalletInput.

Other changes:
- Removing unused inlineIcon
- inputHeight changed to 39 (the same as LineEditMulti)
- Add paste button

### DIFF
--- a/components/LineEdit.qml
+++ b/components/LineEdit.qml
@@ -33,13 +33,20 @@ import QtQuick.Layouts 1.1
 
 import "../components" as MoneroComponents
 
-Item {
+ColumnLayout {
     id: item
+    Layout.fillWidth: true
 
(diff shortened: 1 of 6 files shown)


rating89us:
Updated. I also imported changes from #3553 here, but removed a line (Layout.topMargin: -8).

selsta (approved):

PR #3547, opened 2021-06-09
History: display description; change amount colors; remove dots; display fiat conversion

rating89us:
Changes:
- Change order of row items: Date + Address/Description + Amount. Most cryptocurrency wallets display in this order.
- Remove green and orange dots on the left side
- "Receive" text in green color
- "Send" text in orange color
- Removed "To:" and "From:" on address field
- Description below address/recipient. It is a really important feature in Monero and should always be displayed.
- Incoming amount: "+" and green color
- Outgoing amount: "-" and orange color
- Display fiat conversion (USD/EUR) below XMR amount (when enabled)

image


selsta:
> Display fiat conversion (USD/EUR) below XMR amount (when enabled)

This might be confusing that users expect historical price data here. Not sure.

rating89us:
> users expect historical price data here

Yes, I think we will have to import kraken historical data in the wallet. I'll do it in a future PR. For now, I'll remove fiat price.

PR #3551, opened 2021-06-10
DatePicker: minimum and maximum dates

### LINKED ISSUE #2468: Transactions page: "Date to" should always be later than "Date from"
image

rating89us:
Closes #2468

Main changes:
- "Date from" minimum date = Date of first transaction or date of Monero birth (2014-04-18) if wallet has 0 transactions
- "Date from" maximum date = "Date to" selected date
- "Date to" minimum date = "Date from" selected date
- "Date to" maximum date = Today
- Add "First transaction" and "Today" links on calendar bottom
- Mouse wheel considers minimum and maximum dates


plowsof:
This is nice, the mouse wheel scroll respects the date limits imposed as promised. Today / First transaction dates work correctly.

The only weird thing i spotted (as shown in the 'Today' image) today is the 10th , but the highlighted day appears to be 11 (which is out of bounds , and not being  drawn, this may be a timezone issue as its currently 2am on the 10th.. will check again tomorrow if its a bug or not)

Also, the orange text is going against the standard for clickable items, which are 'white' and change to orange when hovered over, i think we should keep with that.


date_1
date_2

PR #3558, opened 2021-06-12
History: advanced options hidden when transaction count = 0

rating89us:
Both advanced options  ("Human readable date format" and "Export all history") are useful only when transaction count >0.


selsta:
This was previously discussed in https://github.com/monero-project/monero-gui/pull/2607

selsta:
Closing as this was rejected in 2607.

PR #3563, opened 2021-06-14
main: fix bug not accepting new restore height

### LINKED ISSUE #2463: Can't change wallet restore height after restoring a wallet
I was recovering a wallet from seed and forgot to add my wallet restore height.

So I tried to change it in Settings > Info during the sync, but it didn't work, because the wallet kept returning its value to 0.

I'm using GUI 0.15.0.1 in Windows.

Steps to reproduce:
1. Import a wallet from seed
2. Don't set a restore height
3. Select remote node
4. Wallet will start sync from block 0
5. Go to Settings > Info and change _Wallet restore height_ to a higher height (I set 1819000)
6. Confirm the prompt about the wallet cache
7. Wallet automatically returns to restore height = 0 and start sync from the block 0 again.

rating89us:
Fixes  #2463

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -359,6 +359,7 @@ ApplicationWindow {
         middlePanel.getProofClicked.connect(handleGetProof);
         middlePanel.checkProofClicked.connect(handleCheckProof);
 
+        persistentSettings.restore_height = currentWallet.walletCreationHeight;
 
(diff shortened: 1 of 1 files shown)


selsta (approved):
Good catch. I don't really understand why this issue happened in the first place but this seems to fix it.

PR #3564, opened 2021-06-15
SettingsInfo, WizardRestoreWallet1, WizardCreateDevice1: Correct date typed in a wrong format

### LINKED ISSUE #3207: Restore: date inadvertently entered as YYYYMMDD is accepted as restore blockheight
I've already seen two users reporting that they had problems restoring a wallet after entering YYYYMMDD instead of the expected YYYY-MM-DD (one is here: https://github.com/monero-project/monero-gui/issues/3206#issuecomment-720559061). The other user reported that he typed YYYYMMDD because his keyboard was not typing the "-" separator.

Monero GUI accepts the YYYYMMDD number as a restore blockheight that is ~10x higher that current blockheight:
- Current blockheight: 2221943
- Restore blockheight: 20201102 (~10x higher; user should have entered 2020-11-02)

As a temporary workaround, I'd suggest the restore wallet page should convert the entered number to a date (YYYY-MM-DD) if:
1. entered number is too high to be a restore blockheight (>current blockheight)
2. entered number starts with 4 digits that seems to represent years (2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021).

rating89us:
Fixes #3207

Probably the best solution is to use two fields, one for date and another for restore height, but let's use this PR as a temporary fix.

### DIFF
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -115,3 +115,18 @@ function capitalize(s){
 function removeTrailingZeros(value) {
     return (value + '').replace(/(\.\d*?)0+$/, '$1').replace(/\.$/, '');
 }
+
+function parseDateStringOrRestoreHeightAsInteger(value) {
+    // Parse date string or restore height as integer
+    var restoreHeight = 0;
+    if (value.indexOf('-') === 4 && value.length === 10) {
+        restoreHeight = Wizard.getApproximateBlockchainHeight(value, Utils.netTypeToString());
+    } else if (parseInt(value.substring(0, 4)) >= 2014 && parseInt(value.substring(0, 4)) <= 2025 && value.length === 8) {
+        // Correct date typed in a wrong format (20201225 instead of 2020-12-25)
+        var restoreHeightHyphenated = value.substring(0, 4) + "-" + value.substring(4, 6) + "-" + value.substring(6, 8);
+        restoreHeight = Wizard.getApproximateBlockchainHeight(restoreHeightHyphenated, Utils.netTypeToString());
+    } else {
+        restoreHeight = parseInt(value);
+    }
+    return restoreHeight;
+}
--- a/wizard/WizardCreateDevice1.qml
+++ b/wizard/WizardCreateDevice1.qml
@@ -183,15 +183,8 @@ Rectangle {
                     wizardController.walletOptionsDeviceName = wizardCreateDevice1.deviceName;
                     if(lookahead.text)
                         wizardController.walletOptionsSubaddressLookahead = lookahead.text;
-                    var _restoreHeight = 0;
                     if(restoreHeight.text){
-                        // Parse date string or restore height as integer
-                        if(restoreHeight.text.indexOf('-') === 4 && restoreHeight.text.length === 10){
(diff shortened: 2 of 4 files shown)


selsta:
Now would be a good time to move the whole parse part into an external function (e.g. js/Utils.js) so that we don't have to write the same code 3 times.

wizard/WizardCreateDevice1.qml
```diff
@@ -185,12 +185,7 @@ Rectangle {
                         wizardController.walletOptionsSubaddressLookahead = lookahead.text;
                     var _restoreHeight = 0;
                     if(restoreHeight.text){
-                        // Parse date string or restore height as integer
-                        if(restoreHeight.text.indexOf('-') === 4 && restoreHeight.text.length === 10){
-                            _restoreHeight = Wizard.getApproximateBlockchainHeight(restoreHeight.text, Utils.netTypeToString());
-                        } else {
-                            _restoreHeight = parseInt(restoreHeight.text)
-                        }
+                        _restoreHeight = Utils.parseDateStringOrRestoreHeightAsInteger(restoreHeight.text);
```
selsta, line 188:
```
wizardController.walletOptionsRestoreHeight = Utils.parseDateStringOrRestoreHeightAsInteger(restoreHeight.text);
```

Just set it directly, no need for an extra variable.

Same in wizard/WizardRestoreWallet1.qml.

js/Utils.js
```diff
@@ -115,3 +115,18 @@ function capitalize(s){
 function removeTrailingZeros(value) {
     return (value + '').replace(/(\.\d*?)0+$/, '$1').replace(/\.$/, '');
 }
+
+function parseDateStringOrRestoreHeightAsInteger(value) {
+    // Parse date string or restore height as integer
+    var _restoreHeight = 0;
```
selsta, line 121: Also while we are at it, you can remove the `_` prefix here, it does not match our other coding style.

selsta:
Looks good now. Please squash.

js/Utils.js
```diff
@@ -115,3 +115,18 @@ function capitalize(s){
 function removeTrailingZeros(value) {
     return (value + '').replace(/(\.\d*?)0+$/, '$1').replace(/\.$/, '');
 }
+
+function parseDateStringOrRestoreHeightAsInteger(value) {
+    // Parse date string or restore height as integer
+    var restoreHeight = 0;
+    if (value.indexOf('-') === 4 && value.length === 10) {
+        restoreHeight = Wizard.getApproximateBlockchainHeight(value, Utils.netTypeToString());
+    } else if (value.substring(0, 4) >= "2014" && value.substring(0, 4) <= "2025" && value.length === 8) {
```
selsta, line 124:
```suggestion
    } else if (parseInt(value.substring(0, 4)) >= 2014 && parseInt(value.substring(0, 4)) <= 2025 && value.length === 8) {
```

This isn't 100% necessary but I would add this to avoid bugs like this https://stackoverflow.com/questions/5630123/javascript-string-integer-comparisons

I know in this specific code this can't happen but let's be on the safe side.

selsta (approved):

PR #3573, opened 2021-06-18
Utils, SettingsInfo, WizardCreateDevice1: use minimum value for restore height of Ledger and Trezor

rating89us:
Use minimum restore heights values for each device model, according to when they started supporting Monero (see monero-project/monero#6797). Now users setting a too low restore height will not sync unnecessary blocks (from 2014 until 2018-11 or 2019-03).

Ledger Nano S support: 
- height: 1695508 (2018-11-01)

Trezor Model T support: 
- height: 1781908 (2019-03-01)


selsta (changes requested):
- As far as I can see this ignores Testnet / Stagnet, we would only want this on Mainnet
- IMO if someone enters a specific block height we should also set this block height. Only if someone keeps the restore height empty then we should use this minimum value for Ledger / Trezor.

rating89us:
> we would only want this on Mainnet

Fixed

> IMO if someone enters a specific block height we should also set this block height. Only if someone keeps the restore height empty then we should use this minimum value for Ledger / Trezor instead of 0.

I disagree. Why should we let users sync unnecessary blocks from 2014-04 to 2018-03/2018-11? Just for the sake of respecting the restore height entered by the user? Why would a user want to sync 4 years of unnecessary blocks?

selsta:
> Just for the sake of respecting the restore height entered by the user?

I personally dislike software that overwrites my entry.

> Why would a user want to sync 4 years of unnecessary blocks?

I can imagine that someone converts their existing paper wallet monero seed into a Ledger seed. In this case it is possible that they have to sync earlier than "2018-03-04". Also some used Ledger before it was publicly available, including me. I don't know if my first transaction was before `2018-03-04`, but I definitely used it before release. This are just 2 quick idea where it could be necessary.

I will also ask @dEBRUYNE-1 as he had this idea initially.

dEBRUYNE-1:
The coded restore height should arguably only be used if the user enters `0` or `1` as value or if no value is set. There are a few scenarios where a user would desire to use a height that is in advance of the suggested height, hence my recommendation to only use the coded height in aforementioned scenario.

rating89us:
Made the following changes:
- Default value of Restore height field on WizardCreateDevice is now blank (it was "0")
- If restore height field is blank & Trezor: use restore height 1697500
- If restore height field is blank & Ledger: use restore height 1522000
- Restore height field on SettingsInfo now accepts blank value ("") as restore height

js/Utils.js
```diff
@@ -125,8 +125,19 @@ function parseDateStringOrRestoreHeightAsInteger(value) {
...
+        var trezorMinRestoreHeight = 1697500;
+        console.log("Restore height undefined. Changing restore height to %1.".arg(device == "Ledger" ? ledgerMinRestoreHeight : trezorMinRestoreHeight));
+        restoreHeight = (device == "Ledger" ? ledgerMinRestoreHeight : trezorMinRestoreHeight);
     } else {
-        restoreHeight = parseInt(value);
+        if (value == "") {
+            restoreHeight = 0;
+        } else {
+            restoreHeight = parseInt(value);
+        }
```
selsta, line 140:
```
} else if (value != "") {
    restoreHeight = parseInt(value);
}
```
`restoreHeight` is already set to 0 above so no need to set it twice.

js/Utils.js
```diff
@@ -125,8 +125,19 @@ function parseDateStringOrRestoreHeightAsInteger(value) {
         // Correct date typed in a wrong format (20201225 instead of 2020-12-25)
         var restoreHeightHyphenated = value.substring(0, 4) + "-" + value.substring(4, 6) + "-" + value.substring(6, 8);
         restoreHeight = Wizard.getApproximateBlockchainHeight(restoreHeightHyphenated, Utils.netTypeToString());
+    } else if ((device == "Ledger" || device == "Trezor") && value == "" && Utils.netTypeToString() == "Mainnet") {
+        // 1522000 = 2018-03-04 (Ledger Nano S support)
+        var ledgerMinRestoreHeight = 1522000;
+        // 1697500 = 2018-11-04 (Trezor Model T support)
+        var trezorMinRestoreHeight = 1697500;
+        console.log("Restore height undefined. Changing restore height to %1.".arg(device == "Ledger" ? ledgerMinRestoreHeight : trezorMinRestoreHeight));
+        restoreHeight = (device == "Ledger" ? ledgerMinRestoreHeight : trezorMinRestoreHeight);
```
selsta, line 134:
```suggestion
        restoreHeight = (device == "Ledger" ? ledgerMinRestoreHeight : trezorMinRestoreHeight);
        console.log("Restore height undefined. Changing restore height to %1.".arg(restoreHeight);
```

Makes the code a bit simpler.

wizard/WizardCreateDevice1.qml
```diff
@@ -183,10 +182,8 @@ Rectangle {
                     wizardController.walletOptionsDeviceName = wizardCreateDevice1.deviceName;
                     if(lookahead.text)
                         wizardController.walletOptionsSubaddressLookahead = lookahead.text;
-                    if(restoreHeight.text){
-                        wizardController.walletOptionsRestoreHeight = Utils.parseDateStringOrRestoreHeightAsInteger(restoreHeight.text);
-                    }
 
+                    wizardController.walletOptionsRestoreHeight = Utils.parseDateStringOrRestoreHeightAsInteger(restoreHeight.text, wizardCreateDevice1.deviceName);
```
selsta, line 186:
```suggestion
                    if (wizardController.walletOptionsDeviceIsRestore) {
                        wizardController.walletOptionsRestoreHeight = Utils.parseDateStringOrRestoreHeightAsInteger(restoreHeight.text, wizardCreateDevice1.deviceName);
                    }
```

I did not test this but I'm quite sure this has to be added else it breaks the "Create new wallet from device" option.

selsta (changes requested):

rating89us:
There is still a bug to be fixed (I think it was already present before this PR): if you set restore height = 0 on  WizardCreateDevice1.qml (I tested on Trezor only), the restore height is set to current restore height.

selsta:
I think that's intentional. Will have to look at the code again but we want to set it to 1 if someone enters 0.

rating89us:
but if you enter restore height = 0 on SettingsInfo, it respects the user input

selsta:
That's most likely a bug then.

rating89us:
I observed this behavior only when you restore a hardware wallet (Trezor). If you restore a normal wallet (on Wizard) or change the restore height on SettingsInfo, it respects the restore height = 0.

selsta:
Yes, that's intentional. Restoring height 0 with hw device sets it to approximately current height.

rating89us:
But it's not consistent. And it's really strange, because "0" was the default value of restore height field.

Now that the default value of restore height is blank, don't you think we should change it to respect the "0" entered by the user?

PR #3574, opened 2021-06-19
StandardDropdown: use ColumnLayout; include dropdownLabel

### LINKED ISSUE #2205: MoneroComponents.StandardDropdown should include labelText
QML API for the following 2 components: `lineEdit` and `lineEditMulti` both have a `labelText` property that functions as the label text, displayed above the aforementioned components.

`MoneroComponents.StandardDropdown` however, does not have a `labelText` property. If you want text above a `StandardDropdown`, it is necessary to include it within a `Layout.ColumnLayout` and use a `Text {}` field there.

Which results in PRs like this: https://github.com/monero-project/monero-gui/pull/2198

On-going effort for me is 'standardization' of our existing components in order to create an uniform API.

In addition, I believe our current dropdown component has some layout issues as observed here: https://github.com/monero-project/monero-gui/blob/68c7cf7276b8192ef800e3ab8fcc51815fb9c1cc/pages/settings/SettingsLayout.qml#L193

The layout of the language picker dropdown box is wrong.

Another example below.

https://github.com/monero-project/monero-gui/blob/68c7cf7276b8192ef800e3ab8fcc51815fb9c1cc/pages/Transfer.qml#L184-L220

Which, ideally, would be rewritten to:

```QML
         MoneroComponents.StandardDropdown { 
             Layout.fillWidth: true 
             labelText: qsTr("Transaction priority") + translationManager.emptyString
             id: priorityDropdown 
             Layout.topMargin: 5 
             currentIndex: 0 
         } 
```

rating89us:
closes #2205

Main changes:
- use ColumnLayout instead of Item (consistent with LineEdit component)
- use same height of LineEdit component (39 pixels)
- include dropdownLabel on StandardDropdown
- set maximum width for dropdowns on WizardHome and SettingsLayout pages

### DIFF
--- a/wizard/WizardHome.qml
+++ b/wizard/WizardHome.qml
@@ -181,43 +181,36 @@ Rectangle {
                 columns: 4
                 columnSpacing: 20
                 Layout.fillWidth: true
+                Layout.topMargin: 10
 
-                ColumnLayout {
-                    Layout.topMargin: 4
+                MoneroComponents.StandardDropdown {
+                    id: networkTypeDropdown
+                    currentIndex: persistentSettings.nettype
+                    dataModel: networkTypeModel
+                    Layout.maximumWidth: 180
+                    labelText: qsTr("Network") + ":" + translationManager.emptyString
+                    labelFontSize: 14
 
-                    MoneroComponents.Label {
-                        text: qsTr("Change Network:") + translationManager.emptyString
-                        fontSize: 14
-                    }
-
-                    MoneroComponents.StandardDropdown {
-                        id: networkTypeDropdown
-                        currentIndex: persistentSettings.nettype
-                        dataModel: networkTypeModel
-                        Layout.fillWidth: true
-                        Layout.maximumWidth: 180
-                        Layout.topMargin: 5
-
-                        onChanged: {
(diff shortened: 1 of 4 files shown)


wizard/WizardHome.qml
```diff
@@ -181,43 +181,36 @@ Rectangle {
...
                     }
                 }
 
                 MoneroComponents.LineEdit {
                     id: kdfRoundsText
-                    Layout.fillWidth: true
+                    Layout.maximumWidth: 180
 
-                    labelText: qsTr("Number of KDF rounds:") + translationManager.emptyString
+                    labelText: qsTr("Number of KDF rounds") + ":" + translationManager.emptyString
```
selsta, line 211: This means this string has to get retranslated in all languages. We try to avoid this. PR looks good otherwise.

rating89us: And what if we change it to `Number of rounds for the key derivation function (KDF rounds)`
Not everyone knows what KDF means.

selsta: We don't want that anyone changes this unless they 100% know what they are doing. If someone knows what this does, they also know what KDF rounds are.

selsta (approved):

PR #3576, opened 2021-06-19
Transfer: don't overlap inside borders; fix margins

rating89us:
Changes:

Don't overlap inside borders: 
Before:
image

After: 
image

Height is the same as Transaction Priority dropdown (42 pixels)
image


selsta:
Screenshot 2021-06-19 at 22 30 17

I don't know if related to this PR but the margin on the left and right is not equal.

selsta (approved):
My previous comment can be improved (if possible) in a separate PR.

PR #3580, opened 2021-06-21
Receive: selected address details on top; click on QR code to ampliate; click on description to edit; display full address; use standard button for create new address

### LINKED ISSUE #2610: Full address is never displayed
Even when there is enough space, full address is never displayed in Account, Receive and Address book pages.

add

rating89us:
closes  #2610

Changes: 
- Selected address details on top (QR code, address #, description, full address).
- Left click on QR code: increase/decrease QR code size
- Right click on QR code: display "Save as Image" option
- Click on description: edit description
- Display full address
- Click on full address: copy address to clipboard
- Removed "Save as image" button
- ~~Display last created address first on addresses list~~
- Create new address: use standard (orange) button; move to the top.

image

### DIFF
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -72,17 +72,194 @@ Rectangle {
         anchors.top: parent.top
         anchors.right: parent.right
 
-        spacing: 20
+        spacing: 15
+
+        ColumnLayout {
+            id: selectedAddressDetailsColumn
+            Layout.alignment: Qt.AlignHCenter
+            spacing: 0
+            property int qrSize: 220
+
+            Rectangle {
+                id: qrContainer
+                color: MoneroComponents.Style.blackTheme ? "white" : "transparent"
+                Layout.fillWidth: true
+                Layout.alignment: Qt.AlignHCenter
+                Layout.maximumWidth: parent.qrSize
+                Layout.preferredHeight: width
+                radius: 4
+
+                Image {
+                    id: qrCode
+                    anchors.fill: parent
+                    anchors.margins: 1
+                    smooth: false
+                    fillMode: Image.PreserveAspectFit
+                    source: "image://qrcode/" + TxUtils.makeQRCodeString(appWindow.current_address)
+
+                    MouseArea {
+                        anchors.fill: parent
+                        hoverEnabled: true
+                        cursorShape: Qt.PointingHandCursor
+                        acceptedButtons: Qt.LeftButton | Qt.RightButton
+                        onClicked: {
+                            if (mouse.button == Qt.LeftButton){
+                                selectedAddressDetailsColumn.qrSize = selectedAddressDetailsColumn.qrSize == 220 ? 300 : 220;
+                            } else if (mouse.button == Qt.RightButton){
+                                qrMenu.x = this.mouseX;
+                                qrMenu.y = this.mouseY;
+                                qrMenu.open()
+                            }
+                        }
+                    }
+                }
+
+                Menu {
+                    id: qrMenu
+                    title: "QrCode"
+
+                    MenuItem {
+                        text: qsTr("Save as Image") + translationManager.emptyString;
+                        onTriggered: qrFileDialog.open()
+                    }
+                }
+            }
+
+            MoneroComponents.TextPlain {
+                id: selectedaddressIndex
+                Layout.alignment: Qt.AlignHCenter
+                Layout.preferredWidth: 220
+                Layout.maximumWidth: 220
+                Layout.topMargin: 15
+                horizontalAlignment: Text.AlignHCenter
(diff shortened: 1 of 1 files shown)


selsta:
> Removed "Save as image" button

Why? This can be useful. At least add it to the context menu when right clicking on the QR code.

rating89us:
`At least add it to the context menu when right clicking on the QR code.`

Done

selsta:
One thing I'm not sure yet is that opening the page always selects the latest subaddress and also that the sort order is reverse (newest to oldest).

I kinda makes sense but I found it a bit confusing at first.

rating89us:
> opening the page always selects the latest subaddress

Instead of displaying a QR code + address when opening the page, we could display a question:
- Receiving from someone you received before? Select an address from the list below
- Receiving from a new person? Click on Create new address button

rating89us:
Here is an example of what I mean:
reusevscreatenew

selsta:
I don't know yet, I don't like the idea of having a separate page and something when clicking on "Receive", it feels a bit clunky. Though I don't have a different suggestion currently. Maybe for this PR we can keep undo

> One thing I'm not sure yet is that opening the page always selects the latest subaddress and also that the sort order is reverse (newest to oldest).

and then the other changes can be done later when this is merged so that we don't keep this PR stuck.

selsta:
Screenshot 2021-06-25 at 02 10 48

Intentional? I think it looks better centered.

rating89us:
No, I'm trying to fix the width of these two lines (address and label) for Windows, but removed the center alignment.

pages/Receive.qml
```diff
@@ -75,14 +75,182 @@ Rectangle {
...
+                        text: qsTr("Save as Image") + translationManager.emptyString;
+                        onTriggered: qrFileDialog.open()
+                    }
+                }
+            }
+
+            MoneroComponents.TextPlain {
+                id: selectedaddressIndex
+                Layout.alignment: Qt.AlignHCenter
+                Layout.maximumWidth: 220
```
selsta, line 131: Can you add a Layout.topMargin here? The margin between QR code and address isn't the same as between address and description.

pages/Receive.qml
```diff
@@ -75,14 +75,182 @@ Rectangle {
...
+                    anchors.fill: parent
+                    cursorShape: Qt.PointingHandCursor
+                    onClicked: {
+                        clipboard.setText(appWindow.current_address);
+                        var selectedAddressLabel = appWindow.currentWallet.getSubaddressLabel(appWindow.currentWallet.currentSubaddressAccount, subaddressListView.currentIndex);
+                        if (selectedAddressLabel == "") {
+                            appWindow.showStatusMessage(qsTr("Address copied to clipboard") + translationManager.emptyString, 3);
+                        } else {
+                            appWindow.showStatusMessage(qsTr("Address (%1) copied to clipboard").arg(selectedAddressLabel) + translationManager.emptyString, 3);
+                        }
```
selsta, line 188:
```suggestion
                        appWindow.showStatusMessage(qsTr("Address copied to clipboard") + translationManager.emptyString, 3);
```

Just my personal opinion. The label is directly above the address you are clicking anyway and the text is only visible 3 seconds so we want to keep the information as concise as possible.

pages/Receive.qml
```diff
@@ -362,6 +462,7 @@ Rectangle {
         if (appWindow.currentWallet) {
             appWindow.current_address = appWindow.currentWallet.address(appWindow.currentWallet.currentSubaddressAccount, 0)
             appWindow.currentWallet.subaddress.refresh(appWindow.currentWallet.currentSubaddressAccount)
+            subaddressListView.currentIndex = 0
```
selsta, line 465: Any reason for always resetting to index 0?

rating89us: when the pages open for the first time, subaddressListView.currentIndex is undefined (= -1). 
Now it will only change to zero when it's undefined.

rating89us:
fixed your comments and added tooltips to label and address

selsta:
Please squash.

rating89us:
I'm trying to fix the center alignment on windows, but could't do it. Do you have any ideas?
image

selsta:
Which Qt version?

rating89us:
windows 7 GUI 0.17.2.2-83-g3acc24ea (Qt 5.15.2)

selsta:
Weird, I don't have this on Mac with the same Qt version.

selsta:
@rating89us try https://doc.qt.io/qt-5/qml-qtquick-text.html#horizontalAlignment-prop

also play around with wrap mode, layout.fillWidth: true, etc.

rating89us:
fixed

selsta (approved):

PR #3581, opened 2021-06-21
Include hardware wallet support on unrooted android devices

Malinero:
This requires an [unmerged branch in libusb](https://github.com/libusb/libusb/pull/874) and https://github.com/monero-project/monero/pull/7758


plowsof:
the libusb PR has moved to here and looks active https://github.com/libusb/libusb/pull/1164

PR #3586, opened 2021-06-23
RemoteNodeList: display shield icon if trusted daemon

rating89us:
image


selsta:
This could have the unwanted side effect of thinking that the GUI is connecting to the node with the checkmark. At least that's my first thought when seeing the picture.

selsta:
Maybe there is a different "secure" icon?

rating89us:
changed to shield icon

selsta (approved):

PR #3611, opened 2021-07-05
WizardOpenWallet1: navigate with keyboard; enable accessibility

rating89us:


selsta (changes requested):
The "Open wallet from file" header does not show as selected, otherwise looks good.

selsta (approved):

PR #3628, opened 2021-07-18
SettingsNode: implement offline mode

rating89us:
Requires https://github.com/monero-project/monero/pull/7648 & https://github.com/monero-project/monero/pull/7113

image

To do:
The wallet should not keep trying to connect to daemon when offline mode is on.


src/libwalletqt/Wallet.cpp
```diff
@@ -301,7 +308,14 @@ void Wallet::initAsync(
     });
     if (future.first)
     {
-        setConnectionStatus(Wallet::ConnectionStatus_Connecting);
+        if (m_walletImpl->isOffline())
+        {
+            setConnectionStatus(Wallet::ConnectionStatus_Initializing);
```
selsta, line 313: Could you explain this? Why ConnectionStatus_Initializing when offline?

rating89us: When the user opens the wallet in offline mode, it will display "Initializing wallet" on Network status during initAsync, instead of "Connecting", which is the default. It would be contradictory to display a "Connecting" message when the wallet is in offline mode.

selsta: Could you directly set it to `ConnectionStatus_OfflineMode`?

rating89us: fixed

PR #3629, opened 2021-07-19
History: fix filtering of transactions by date/time

rating89us:
Some recent transactions were not being displayed in Transactions page, specially transactions done near the beginning of the day (for example, at 1:25).

This occured because we were not multiplying `item.timestamp` * 1000, and the toDate value was also incorrect. Both item.timestamp and toDate resulted in dates in 1970, which sometimes results in a incorrect filtering

This PR simplifies how we set these values, and it also sets the fromDate hour to the beginning of the day (0:00) and the toDate hour to the end of the day (23:59:59).

To test this PR, I suggest changing the clock to 1:25, sending a transaction and then checking if it appears on Transactions page.

### DIFF
--- a/pages/History.qml
+++ b/pages/History.qml
@@ -1439,9 +1439,8 @@ Rectangle {
         // applying filters
         root.txData = JSON.parse(JSON.stringify(root.txModelData)); // deepcopy
 
-        const timezoneOffset = new Date().getTimezoneOffset() * 60;
-        var fromDate = Math.floor(fromDatePicker.currentDate.getTime() / 86400000) * 86400 + timezoneOffset;
-        var toDate = (Math.floor(toDatePicker.currentDate.getTime() / 86400000) + 1) * 86400 + timezoneOffset;
+        var fromDate = fromDatePicker.currentDate.setHours(0, 0, 0, 0);
+        var toDate = toDatePicker.currentDate.setHours(23, 59, 59, 999);
 
         var txs = [];
         for (var i = 0; i < root.txData.length; i++){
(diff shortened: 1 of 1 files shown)


selsta (approved):
This date code kinda goes over my head. Did you test all cases now by changing your date on the computer?

rating89us:
I tested a lot of cases, not all possible cases, but I think the most important ones.
The problem resides on the first hours of the day, when the timestamps are near the beginning of unix epoch.  For instance:
A transaction was sent at Jul 19 01:21:10 **2021** while Date picker on Transactions page was Jul 19 01:21:12 2021.
However, `updateFilter()` considers the `item.timestamp` Jan 19 20:50:50 **1970** and `toDate` Jan 19 20:50:45 1970, which means item.timestamp > toDate, but in reality the transaction was sent before toDate.

rating89us:
I'll do some more tests before and after midnight. There have been some reports of users with amount turning zero after leaving computer turned on during the night. Probably the same cause.

selsta:
Thanks

selsta:
#3644 fixed it for me, this PR didn't. Will still try to review it.

selsta (commented):
I don't see any difference before / after this PR when setting the time to after midnight. I would prefer not changing this part of the code, unless I can reproduce what this is fixing.

rating89us:
> I don't see any difference before / after this PR when setting the time to after midnight.

Did you try sending transactions after midnight?

selsta:
> Did you try sending transactions after midnight?

Yes, that's how I tested it.

PR #3630, opened 2021-07-19
RemoteNodeList: truncate long addresses

rating89us:


selsta (changes requested):
- Enter 1.onion:18081 as an example, it will display incorrectly.
- Domains can contain `.onion` in a different ways, see test.onion.test2.com

I would prefer if we use the elide property: https://doc.qt.io/qt-5/qml-qtquick-text.html#elide-prop

rating89us:
> Domains can contain .onion in a different ways, see test.onion.test2.com

fixed

> Enter 1.onion:18081 as an example, it will display incorrectly.

I agree, but this is an edge case. If you look at any [.onion remote node list](https://monero.fail/?onion=true), this PR will work for 100% of nodes.

rating89us:
Also, V2 onion addresses are 16 characters long and V3 are 56 characters long.

selsta:
Screenshot 2021-07-20 at 20 52 04

```diff
diff --git a/components/RemoteNodeList.qml b/components/RemoteNodeList.qml
index ffbce57d..5d3b14fa 100644
--- a/components/RemoteNodeList.qml
+++ b/components/RemoteNodeList.qml
@@ -84,6 +84,7 @@ ColumnLayout {
 
                     MoneroComponents.TextPlain {
                         id: addressText
+                        width: parent.width - trustedDaemonCheckMark.width
                         color: index === remoteNodesModel.selected ? MoneroComponents.Style.defaultFontColor : MoneroComponents.Style.dimmedFontColor
                         anchors.verticalCenter: parent.verticalCenter
                         anchors.left: parent.left
@@ -91,6 +92,7 @@ ColumnLayout {
                         font.pixelSize: 16
                         text: address
                         themeTransition: false
+                        elide: Text.ElideMiddle
                     }
 
                     MoneroComponents.Label {
```

Simpler code and works for all long urls, not only .onion

selsta (changes requested):
Please undo the margin changes. They are unrelated and look wrong to me.

selsta (approved):

PR #3634, opened 2021-07-22
README: TOC, update building titles

rating89us:


selsta:
Github has a TOC button, we removed it intentionally.

rating89us:
This button is really hidden, I've never seen it before.

I think it's better to have this additional TOC, like https://github.com/monero-project/monero has.

selsta (approved):

PR #3635, opened 2021-07-22
Transfer: fix button colors, font size, bold; use secondary button for advanced options

rating89us:
Changes:
- AdvancedOptions: use secondary buttons (only the send button is primary/orange in Transfer page); remove trailing whitespace from source code
- Tooltip: display tooltip icon in default text color (not orange)
- InlineButton: fix background/hover color (sometimes it didn't change after clicking switch theme button)
- StandardButton: smaller font size for small button; remove bold text from secondary button; add transparency to disabled button (keeps original color)
- Send button: remove _rightIconInactive_; add transparency to _rightIcon_
- Style: add variables for black/white themes


components/StandardButton.qml
```diff
@@ -102,7 +99,14 @@ Item {
                 when: !button.enabled
                 PropertyChanges {
                     target: buttonRect
-                    color: MoneroComponents.Style.buttonBackgroundColorDisabled
+                    opacity: 0.5
```
selsta, line 102: Would the opacity automatically go to 1 again if it jumps from `disabled` to `active`? It seems like it never gets reset.

rating89us: Yes, when going from disabled to active it resets to the default value (opacity = 1).

selsta (approved):

PR #3636, opened 2021-07-22
Transfer: send button 

rating89us:
Changes:
- Send button: disable if amount = 0
- Send button: rename to "Review and Send" (user still will be able to review transaction details before sending)
- Send button: increase width and centralize
- Send button warning box: move below send button (so that when warning box appears, it will now change send button position)
- Send button warning box: don't display when amount = 0 (it is always appearing after typing the address)

### DIFF
--- a/pages/Transfer.qml
+++ b/pages/Transfer.qml
@@ -70,11 +70,6 @@ Rectangle {
             if (recipientModel.hasInvalidAddress()) {
                 return qsTr("Address is invalid.") + translationManager.emptyString;
             }
-
-            // Amount is nonzero
-            if (recipientModel.hasEmptyAmount()) {
-                return qsTr("Enter an amount.") + translationManager.emptyString;
-            }
         }
 
         return "";
@@ -214,7 +209,7 @@ Rectangle {
 
             function hasEmptyAmount() {
                 for (var index = 0; index < recipientModel.count; ++index) {
-                    if (recipientModel.get(index).amount === "") {
+                    if (recipientModel.get(index).amount === "" || recipientModel.get(index).amount == 0) {
                         return true;
                     }
                 }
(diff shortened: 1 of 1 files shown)


selsta:
> Send button: disable if amount = 0

Ok

> Send button: rename to "Review and Send" (user still will be able to review transaction details before sending)

This also removes the icon from the button?

> Send button warning box: move below send button (so that when warning box appears, it will now change send button position)

Don't know about that yet :/ Everything else is left aligned so it looks out of place and the button looks too wide IMO. Changing it to "Review and send" will make it wider anyway.

> Send button warning box: move below send button (so that when warning box appears, it will now change send button position)

Should be ok

> Send button warning box: don't display when amount = 0 (it is always appearing after typing the address)

Also ok

rating89us:
> > Send button: rename to "Review and Send" (user still will be able to review transaction details before sending)
> 
> This also removes the icon from the button?

Yes, I updated the description now. "- Send button: remove right arrow icon"

> Don't know about that yet :/ Everything else is left aligned so it looks out of place and the button looks too wide IMO. Changing it to "Review and send" will make it wider anyway.

At least the button should have a minimum width, because in some languages the text has a smaller width.

selsta:
Any reason for removing the icon? It makes it stand out a bit which is something we want.

rating89us:
About the send button alignment, I think that centering and/or making it full width is something that gives importance to the button and make it look that it will "finish" the transfer page. 

While right aligning it would imply that there is a next page/step on middle panel, but there isn't (we will open a tx confirmation popup). 

Left aligning is the worst option in my opinion.

rating89us:
The right arrow icon seem unnecessary to me, now that we have a big bright orange button.

PR #3638, opened 2021-07-23
AddressBook: add Grab QR code inline button

rating89us:
- Add "Grab QR code" inline button, which will import address and recipient name. This is specially useful for QR codes asking donations (see QR code of https://grapheneos.org/donate).
- Import only address and recipient name when scanning QR code with camera (removed importing tx description).

### DIFF
--- a/pages/AddressBook.qml
+++ b/pages/AddressBook.qml
@@ -346,6 +346,29 @@ Rectangle {
                 onReturnPressed: addButton.enabled ? addButton.clicked() : ""
 
                 MoneroComponents.InlineButton {
+                    fontFamily: FontAwesome.fontFamilySolid
+                    fontStyleName: "Solid"
+                    fontPixelSize: 18
+                    text: FontAwesome.desktop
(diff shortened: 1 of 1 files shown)


pages/AddressBook.qml
```diff
@@ -345,6 +345,26 @@ Rectangle {
...
+                    onClicked: {
+                        clearFields();
+                        const codes = oshelper.grabQrCodesFromScreen();
+                        for (var index = 0; index < codes.length; ++index) {
+                            const parsed = walletManager.parse_uri_to_object(codes[index]);
+                            if (!parsed.error) {
+                                addressLine.text = parsed.address
+                                descriptionLine.text = parsed.recipient_name
+                                break;
+                            }
```
selsta, line 363:
```suggestion
                            if (!parsed.error) {
                                addressLine.text = parsed.address
                                descriptionLine.text = parsed.recipient_name
                                break;
                            } else if (walletManager.addressValid(codes[index], appWindow.persistentSettings.nettype)) {
                                addressLine.text = codes[index];
                                break;
                            }
```

selsta (approved):

PR #3649, opened 2021-07-29
Mining: use "local node" instead of "daemon"; remove "<br>" from error strings

rating89us:


erciccione (approved):

rating89us:
@selsta please review

PR #3655, opened 2021-08-01
Windows.js: minimize frameless window when clicking on taskbar icon

### LINKED ISSUE #3653: [Windows] GUI Does Not Minimize On Taskbar Icon Click
To reproduce, make sure the GUI is in the foreground and that custom decorations are used.

Next click the Monero GUI icon in the taskbar.
Expected result is for the GUI to minimize to taskbar.
Actual result is no result (as if the program is trying to become the foreground application it already is).

I also noticed that when the minimize icon is clicked, the window just disappears instead of the minimizing to taskbar animation one would expect.

rating89us:
Fixes #3653

### DIFF
--- a/js/Windows.js
+++ b/js/Windows.js
@@ -1,4 +1,4 @@
-var flagsCustomDecorations = (Qt.FramelessWindowHint | Qt.CustomizeWindowHint | Qt.WindowSystemMenuHint | Qt.Window);
+var flagsCustomDecorations = (Qt.FramelessWindowHint | Qt.CustomizeWindowHint | Qt.WindowMinimizeButtonHint | Qt.WindowSystemMenuHint | Qt.Window);
 var flags = (Qt.WindowSystemMenuHint | Qt.Window | Qt.WindowMinimizeButtonHint | Qt.WindowCloseButtonHint | Qt.WindowTitleHint | Qt.WindowMaximizeButtonHint | Qt.WindowFullscreenButtonHint);
 
 /**


selsta (commented):
I have found this online, we have to confirm that this flag doesn't break frameless window.

https://github.com/Jorgen-VikingGod/Qt-Frameless-Window-DarkStyle/issues/4

https://github.com/Jorgen-VikingGod/Qt-Frameless-Window-DarkStyle/commit/2a02c0e803d07fe1fee8abe332d4260a81683b58

elibroftw:
@selsta, can't you just test the artifacts on Linux then to see if frameless windows brake on Linux? If it does break, then you can do the below.

From https://forum.qt.io/topic/5739/platform-specific-part-of-qml-file/13
> Just a quick update to the thread: It seems that now you can check the platform directly in your QML code:
> 
> @ if (Qt.platform.os === "linux" || Qt.platform.os === "windows" || Qt.platform.os === "osx" || Qt.platform.os === "unix") {
> //..
> }
> @
> 
> This doesn't seem to be documented anywhere else though :S (so not sure if this can change, without notice).
> 
> Got the reference from:
> http://qt-project.org/forums/viewthread/39956
> http://qt-project.org/doc/qt-5/qtmultimedia-video-qmlvideofx-qml-qmlvideofx-main-qml.html
> 

```js
var flagsCustomDecorations = (Qt.FramelessWindowHint | Qt.CustomizeWindowHint | Qt.WindowSystemMenuHint | Qt.Window);
if (Qt.platform.os === "windows") {
    flagsCustomDecorations = flagsCustomDecorations  | Qt.WindowMinimizeButtonHint;
}
```

selsta:
Continue in #3905

PR #3675, opened 2021-08-21
main: alternative wording for "Create Desktop Shortcut" window

### LINKED ISSUE #3673: Awkward/non-intuitive wording in Desktop Entry window (Translation / EN)
I just installed the GUI for the first time in a long time and it took me half a second to parse the first window and prompt that I saw. I'd like to suggest an alternative wording. Just looking to minimize the barriers to entry...

Current:

Window Title: Desktop Entry
Prompt: Would you like to register Monero GUI Desktop entry?

Suggested:

Window Title: Create Desktop Shortcut
Prompt: Would you like to create a shortcut to your Monero wallet on your desktop?

There it is, the most trivial "Issue" you'll see all day, hopefully!

rating89us:
Fixes #3673


selsta:
https://wiki.archlinux.org/title/desktop_entries

Desktop entry is how it's called on Linux.

rating89us:
I think most Linux begginers wouldn't understand "desktop entry", but everyone (both beginners and advanced Linux users) would understand "desktop shorcut".

selsta:
My problem is I'm not sure if desktop shortcut and desktop entry is the same in Linux.

PR #3677, opened 2021-08-23
wizard: move mnemonic seed into a separate screen

### LINKED ISSUE #2043: [UX] Move seed into a separate screen of the onboarding flow
During [user testing](https://paper.dropbox.com/doc/Monero-GUI-user-testing--AaPItIR28uMAAkEXUdQZnYARAg-YAWmy01OJa5vkmdQvt6wg), testers seemed to miss the seed section or not understand it. It just seems to get a bit lost because it is below the wallet name and location, and because the term "Mnemonic seed" isn't clear. Writing down the seed and storing it securely is probably the most important thing a user needs to do during wallet creation, so I suggest moving it into a separate screen.

Here's a suggestion for what this screen could look like, all feedback is welcome. The title is very direct about what users should be doing, and there's a checkbox to confirm that they wrote it down.

monero-seed-gbks-190329

And here's the print template referenced in the screen. Without this, users may just use a post-it or a scrap of paper they happen to have on their desks (or just copy/paste the seed into a file). A proper template should hopefully make it easier for users to "do the right thing".

[Monero recovery phrase template.pdf](https://github.com/monero-project/monero-gui/files/3022055/Monero.recovery.phrase.template.pdf)

rating89us:
Closes #2043

This PR introduces a dedicated page to the mnemonic seed + a pdf template to be printed (with 25 blank fields). When the user clicks on Next button, 5 random words are hidden and the user is requested to type them to confirm that the mnemonic seed was written down.

Main changes:
- Create dedicated seed page on "Create a new wallet" flow
- Add description explaining what is the mnemonic seed, its importance and detailed instructions on how to proceed
- Display seed in 25 numbered fields. The seed is generated when the user clicks on "Create a new wallet" menu item. If the user is in seed page and click on previous button, the wallet file name and location can be changed and the seed will remain the same.
- Wallet creation date + Wallet restore height
- Add tooltips
- "Create new seed" button: creates a new seed on the same page (keeps wallet name and location defined on the previous page)
- "Copy to clipboard" button: copies the mnemonic seed to the clipboard (visible on advanced mode only because copying the seed to the clipboard is something potentially dangerous)
- "Print a template" button: displays a pdf file with blank numbered fields, one for each word (always visible)
- Fix width of wizard navigation buttons (180)
- Use secondary color for Previous (back) button
- Accessibility

Writing down seed and verifying it:
aaa123

Choosing 5 new random words to hide (on in each column) every time verification page is opened
aaa1234

Resizing from 5 columns (desktop) to 2 columns (mobile)
aaa12345


rating89us:
@GBKS, could you please create a new version of the pdf template with the following text: 
- change "Block height" to "Wallet restore height"
- general instructions: "The 25 words above are your recovery phrase (mnemonic seed). They are the only thing needed to access your funds and restore your Monero wallet, so no one except you should be able to access it. It's strongly NOT recommended to store your recovery phrase digitally (in an email, online service, screenshot, photo or any other type of computer file)"
- warning with large and bold font: "**Print this paper, fill it out, and keep it in a safe location! NEVER SHARE YOUR MNEMONIC SEED WITH ANYBODY, ESPECIALLY WITH STRANGERS OFFERING TECHNICAL SUPPORT!**"
- instructions on how to restore wallet: "For instructions on how to restore this wallet, visit www.getmonero.org and go to Resources > User Guides > "How to restore a wallet from mnemonic seed". Use only Monero wallets that are trusted and recommended by the Monero community (see a list of them in www.getmonero.org/downloads)."

GBKS:
Hi there, here's an update to the template. I am not a fan of screaming at users with huge capitalized text and exclamation marks, so I put the instructions (and warning notice) at the top so it's clearly noticeable. Let me know what you think.

[monero-onboarding-recovery-phrase-template-gbks-210906.pdf](https://github.com/monero-project/monero-gui/files/7114703/monero-onboarding-recovery-phrase-template-gbks-210906.pdf)
monero-onboarding-recovery-phrase-template-gbks-210906

rating89us:
@GBKS It would also be nice to have a dedicated icon for the verification part, something like this: 
image

selsta:
Did you get opening a PDF from qrc to work yet? I doesn't seem to work on Mac. As far as I know you have to write it to disk first before you can open it.

wizard/WizardCreateWallet2.qml
```diff
@@ -47,39 +102,332 @@ Rectangle {
...
+                    text: qsTr("Copy to clipboard") + translationManager.emptyString
+                    tooltip: qsTr("Copy the mnemonic seed to the clipboard") + translationManager.emptyString
+                    onClicked: {
+                        clipboard.setText(wizardController.walletOptionsSeed);
+                        appWindow.showStatusMessage(qsTr("Recovery phrase copied to clipboard"),3);
+                    }
+                    Accessible.role: Accessible.Button
+                    Accessible.name: qsTr("Copy to clipboard") + translationManager.emptyString
+                    KeyNavigation.up: wizardCreateWallet2.seedListGrid ? seedListGridColumn.children[0].children[24] : recoveryPhraseLabel
+                    KeyNavigation.backtab: wizardCreateWallet2.seedListGrid ? seedListGridColumn.children[0].children[24] : recoveryPhraseLabel
```
selsta, line 287:
```suggestion
                    KeyNavigation.up: (wizardCreateWallet2.seedListGrid && seedListGridColumn.children[0]) ? seedListGridColumn.children[0].children[24] : recoveryPhraseLabel
                    KeyNavigation.backtab: (wizardCreateWallet2.seedListGrid && seedListGridColumn.children[0]) ? seedListGridColumn.children[0].children[24] : recoveryPhraseLabel
```

```
2021-10-05 22:36:49.032	W qrc:/wizard/WizardCreateWallet2.qml:286: TypeError: Cannot read property 'children' of undefined
2021-10-05 22:36:49.032	W qrc:/wizard/WizardCreateWallet2.qml:287: TypeError: Cannot read property 'children' of undefined
```

wizard/WizardCreateWallet2.qml
```diff
@@ -47,39 +102,332 @@ Rectangle {
...
-            spacing: 0
+            visible: !mobileDialog.visible
+            spacing: 15
+
+            WizardHeader {
+                id: header
+                imageIcon: wizardController.layoutScale != 4 ? (MoneroComponents.Style.blackTheme ? "qrc:///images/write-down.png" : "qrc:///images/write-down-white.png") : ""
+                title: qsTr("Write down your recovery phrase") + translationManager.emptyString
+                subtitleVisible: wizardController.layoutScale != 4
+                subtitle: qsTr("These words are a backup of your wallet. WRITE THESE WORDS DOWN NOW ON A PIECE OF PAPER in the same order displayed. Keep this paper in a safe place and do not disclose it to anybody! DO NOT store these words digitally, always use a paper!") + translationManager.emptyString
```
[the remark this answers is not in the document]
rating89us: better just use normal text format.

rating89us:
> Did you get opening a PDF from qrc to work yet? I doesn't seem to work on Mac. As far as I know you have to write it to disk first before you can open it.

It's working on Windows and Ubuntu.  I haven't tested on Mac yet.

selsta:
So the only thing remaining here is getting the PDF to work on Mac?

rating89us:
> So the only thing remaining here is getting the PDF to work on Mac?
Yes

wizard/WizardCreateWallet2.qml
```diff
@@ -47,39 +98,318 @@ Rectangle {
...
+                    MoneroComponents.TextPlain {
+                        id: walletCreationDateValue
+                        property var locale: Qt.locale()
+                        property date currentDate: new Date()
+                        font.family: MoneroComponents.Style.fontRegular.name
+                        font.pixelSize: 16
+                        font.bold: true
+                        textFormat: Text.RichText
+                        color: MoneroComponents.Style.defaultFontColor
+                        text: currentDate.toLocaleDateString(locale, Locale.ShortFormat)
```
selsta, line 331: What do you think about specifying it as `YYYY-MM-DD`? That's the format that is required for the restore height textbox.

rating89us: I prefer the default LocaleDateString. Later I want to display a calendar during wallet restore instead of asking for YYYY-MM-DD, which is not familiar to users from multiple countries.

rating89us:
Added "create a new seed" button

selsta (commented):
Couple minor comments. Did you test the binaries that were created with CI to check if the PDF works in them?

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+                    }
+                }
+
+                MoneroComponents.StandardButton {
+                    id: createNewSeedButton
+                    visible: appWindow.walletMode >= 2
+                    small: true
+                    primary: false
+                    text: qsTr("Create new seed") + translationManager.emptyString
+                    tooltip: qsTr("Generate a new seed") + translationManager.emptyString
```
selsta, line 282: I don't see a point for this tooltip. It's redundant.

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+                    onClicked: {
+                        wizardController.restart(true);
+                        wizardController.createWallet();
+                        wizardCreateWallet2.seedArray = wizardController.walletOptionsSeed.split(" ")
+                        wizardCreateWallet2.seedListGrid.destroy();
+                        checkSeedListGridDestruction.start();
+                    }
+                    Accessible.role: Accessible.Button
+                    Accessible.name: qsTr("Generate a new seed") + translationManager.emptyString
```
selsta, line 291: Would change it to "Create new seed" so that we don't have to two similar strings for no apparent reason.

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+                    KeyNavigation.tab: copyToClipboardButton
+                }
+
+                MoneroComponents.StandardButton {
+                    id: copyToClipboardButton
+                    visible: appWindow.walletMode >= 2
+                    small: true
+                    primary: false
+                    text: qsTr("Copy to clipboard") + translationManager.emptyString
+                    tooltip: qsTr("Copy the mnemonic seed to the clipboard") + translationManager.emptyString
```
selsta, line 304: Also redundant, would remove.

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+                    color: mobileImage.focus ? MoneroComponents.Style.titleBarButtonHoverColor : "transparent"
+                }
+            }
+
+            Text {
+                id: mobileText
+                Layout.fillWidth: true
+                Layout.alignment: Qt.AlignHCenter
+                color: MoneroComponents.Style.dimmedFontColor
+                text: qsTr("The next page will display your recovery phrase, also known as mnemonic seed. ") + qsTr("These words are a backup of your wallet. Write these words down now on a piece of paper in the same order displayed. Keep this paper in a safe place and do not disclose it to anybody! Do not store these words digitally, always use a paper!") + translationManager.emptyString
```
selsta, line 154: Would remove the " " out of the translatable string

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+
+                font.family: MoneroComponents.Style.fontRegular.name
+                font.pixelSize: 14
+                wrapMode: Text.WordWrap
+                leftPadding: 0
+                topPadding: 0
+                Accessible.role: Accessible.StaticText
+                Accessible.name: qsTr("The next page will display your recovery phrase, also known as mnemonic seed. ") + qsTr("These words are a backup of your wallet. Write these words down now on a piece of paper in the same order displayed. Keep this paper in a safe place and do not disclose it to anybody! Do not store these words digitally, always use a paper!") + translationManager.emptyString
```
selsta, line 162: Same here

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
-                        wizardStateView.state = "wizardCreateWallet1";
-                    }
+                    wizardStateView.state = "wizardCreateWallet1";
+                    mobileDialog.visible = Qt.binding(function() { return wizardController.layoutScale == 4 })
                 }
+                btnPrevKeyNavigationBackTab: wizardCreateWallet2.state == "default" ? walletRestoreHeight
+                                                                                    : seedListGridColumn.children[0].children[wizardCreateWallet2.hiddenWords[4]].lineEdit.visible ? seedListGridColumn.children[0].children[wizardCreateWallet2.hiddenWords[4]].lineEdit
+                                                                                                                                                                                   : seedListGridColumn.children[0].children[24]
+                btnNextKeyNavigationTab: mobileDialog.visible ? mobileHeader : header
+                btnNext.enabled: walletCreationDate.opacity == 1 || appWindow.ctrlPressed ? true
```
selsta, line 423: Would use here `wizardCreateWallet2.state == "default"` instead of `walletCreationDate.opacity == 1`

I assume they should be equivalent.

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+                        font.pixelSize: 15
+                        font.bold: false
+                        textFormat: Text.RichText
+                        color: MoneroComponents.Style.dimmedFontColor
+                        text: qsTr("Restore height") + ":" + translationManager.emptyString
+                        tooltip: wizardController.layoutScale != 4 ? qsTr("Enter this number when restoring the wallet to make your initial wallet synchronization faster.") : "" + translationManager.emptyString
+                        tooltipIconVisible: true
+                        themeTransition: false
+                        Accessible.role: Accessible.StaticText
+                        Accessible.name: qsTr("Wallet restore height") + " " + Utils.roundDownToNearestThousand(wizardController.m_wallet ? wizardController.m_wallet.walletCreationHeight : 0) + translationManager.emptyString
```
selsta, line 392: Would have Accessible.name and text the same.

wizard/WizardCreateWallet2.qml
```diff
@@ -49,39 +100,353 @@ Rectangle {
...
+                    small: true
+                    primary: false
+                    visible: !isMac
+                    text: qsTr("Print a template") + translationManager.emptyString
+                    tooltip: qsTr("Print a template to write down your seed") + translationManager.emptyString
+                    onClicked: {
+                        oshelper.openFile("wizard/template.pdf")
+                    }
+                    Accessible.role: Accessible.Button
+                    Accessible.name: qsTr("Print a blank template to write down your seed") + translationManager.emptyString
```
selsta, line 328: "Print a template to write down your seed"

wizard/WizardCreateWallet2.qml
rating89us: Doesn't work.

wizard/WizardCreateWallet2.qml
plowsof: ALOT of beginners forget the "-" or just type the date incorrectly. This would save alot of people pain if a calendar was implemented :+1:

selsta (commented):
There is a merge conflict, a rebase is required.

```patch
diff --git a/src/main/oshelper.cpp b/src/main/oshelper.cpp
index a910dae8..ddf3d9ed 100644
--- a/src/main/oshelper.cpp
+++ b/src/main/oshelper.cpp
@@ -292,3 +292,9 @@ bool OSHelper::installed() const
     return false;
 #endif
 }
+
+void OSHelper::openSeedTemplate() const
+{
+    QFile::copy(":/wizard/template.pdf", QDir::tempPath() + "/seed_template.pdf");
+    openFile(QDir::tempPath() + "/seed_template.pdf");
+}
diff --git a/src/main/oshelper.h b/src/main/oshelper.h
index 2d8d41f0..d171fc57 100644
--- a/src/main/oshelper.h
+++ b/src/main/oshelper.h
@@ -53,6 +53,7 @@ public:
     Q_INVOKABLE QString temporaryPath() const;
     Q_INVOKABLE bool removeTemporaryWallet(const QString &walletName) const;
     Q_INVOKABLE bool isCapsLock() const;
+    Q_INVOKABLE void openSeedTemplate() const;
 
 private:
     bool installed() const;
diff --git a/wizard/WizardCreateWallet2.qml b/wizard/WizardCreateWallet2.qml
index 475e722f..04c2bf99 100644
--- a/wizard/WizardCreateWallet2.qml
+++ b/wizard/WizardCreateWallet2.qml
@@ -316,11 +316,10 @@ Rectangle {
                     id: printPDFTemplate
                     small: true
                     primary: false
-                    visible: !isMac
                     text: qsTr("Print a template") + translationManager.emptyString
                     tooltip: qsTr("Print a template to write down your seed") + translationManager.emptyString
                     onClicked: {
-                        oshelper.openFile("wizard/template.pdf")
+                        oshelper.openSeedTemplate();
                     }
                     Accessible.role: Accessible.Button
                     Accessible.name: qsTr("Print a template to write down your seed") + translationManager.emptyString
```

This solves saving the PDF on macOS. I did not test it on other OS.

selsta:
It saves the PDF to the temp directory and then it gets opened. The PDF inside the temp directory gets deleted after the next computer reboot.

selsta:
@rating89us did you time to update this so that we can merge it? :)

selsta:
Let's continue here: #3878

PR #3680, opened 2021-08-31
Makefile: fix release-linux-armv8

### LINKED ISSUE #3642: Could we update the ARM build instructions, please?
I'm on the latest Ubuntu and first tried the "whonix repo method" to get the binary, even though I see the ARM parameter somewhere, it only managed to get me x86 binaries itself.

I also followed the standard Ubuntu recipe with make release, and it broke with the -maes unknown error, which has already been discussed in the RandomX context a long time ago, if not in this repo as well. As -maes simply should be removed for ARM, I looked into the Makefile only to find a release-linux-armv8 target. So I guessed that should do the job, except that it does not behave like make release, has some hardcoded dependency to ../.. for whatever reason. I did try the standard cmake recipe but it breaks almost immediately. 

Finally, I created bin/arm , copied the Makefile there, and now the stars seem to have aligned and the project gets built correctly in monero-gui/bin/arm/bin . Perhaps someone more competent than me knows what to fix here or there to make the build more streamlined.

selsta:
Resolves #3642


mj-xmr (approved):
No regressions detected.

PR #3690, opened 2021-09-05
Transfer, AddressBook: paste QR code image from clipboard into text input

rating89us:
This PR includes #3687 and should be merged after it.

If the clipboard has a QR code image and the user paste (Ctrl+V) inside the text input, the QR code URI will be parsed as text. This feature is useful if someone sends you a payment request as a QR code image and you can't scan it (using mobile, for instance).


selsta:
Can you move the function to `src/main/oshelper.cpp` like the other function?

rating89us:
Monero GUI is crashing if I have some images on clipboard that are not QR codes. I got the following errors:

```
corrupted size vs. prev_size
Aborted (core dumped)
```

```
free(): corrupted unsorted chunks
Aborted (core dumped)
```

```
double free or corruption (out)
Aborted (core dumped)
```

Any idea?

src/main/oshelper.cpp
```diff
@@ -159,6 +160,26 @@ QList<QString> OSHelper::grabQrCodesFromScreen() const
...
 }
 
+QList<QString> OSHelper::grabQrCodesFromClipboard() const
+{
+    QList<QString> codes;
+
+    try
+    {
+        QClipboard *clipboard = QGuiApplication::clipboard();
+        const QImage clipboardImage = clipboard->image();
```
selsta, line 170:
```suggestion
        const QImage clipboardImage = clipboard->image();
        if (clipboardImage.isNull()) {
            return codes;
        }
```

Untested but this should be added.

PR #3706, opened 2021-09-20
AddressBook: add payment request URI prefix to address field placeholder

rating89us:


selsta (approved):

selsta (commented):
Testnet and Stagenet shouldn't advertise OpenAlias, same as on the Transfer page.

selsta (approved):

PR #3719, opened 2021-10-27
cmake: don't print verbose make

selsta:
If someone wants verbose make they can set the env var.
This is now in line with monero's behaviour.


mj-xmr (approved):
Agreed. Errors and warnings are important. Everything else should be explicit, even in Debug.

PR #3734, opened 2021-11-15
Settings: option to skip stop local node screen

### LINKED ISSUE #3732: Allow the option to disable the "Do you want to stop local node" screen.
Hello,

I am currently using monero-gui and running a local node. I run this local node via systemd because I found it easy to setup and it let's me easy run it always.

A problem I have encounterd is that when I set "Local Node" in my Monero GUI it detects the systemd node... so far so good. Now when I exit it it asks me if I want to stop the node, ofcourse I don't I run it on systemd for a reason.

Right now I have found that just setting remote node and then setting 127.0.0.1 as IP works, but I would rather just be able to select local node.

So I want to ask if it would be possible to add an option when selecting your desired option (So for me, don't stop node) and then save that with a "Remember" option so that in the future it just leaves it running.

If this isn't smth which get's added that's fine I hope you have a fantastic day and thanks for your project :D

selsta:
Closes #3732

Needs good testing as I had issues with settings resetting after running this patch.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1368,6 +1368,7 @@ ApplicationWindow {
         }
 
         property bool askDesktopShortcut: isLinux
+        property bool askStopLocalNode: true
         property string language: 'English (US)'
         property string language_wallet: 'English'
         property string locale: 'en_US'
@@ -2123,7 +2124,7 @@ ApplicationWindow {
             showProcessingSplash(qsTr("Checking local node status..."));
             const handler = function(running) {
                 hideProcessingSplash();
-                if (running) {
+                if (running && persistentSettings.askStopLocalNode) {
                     showDaemonIsRunningDialog(closeAccepted);
                 } else {
                     closeAccepted();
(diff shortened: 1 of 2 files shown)


rating89us:
It seems that it's not possible to save any value to persistentSettings after wallet "close event" is called.

That is the reason why I decided to implement a dropdown on Settings > Node page in #2898, and that's why I wasn't able to implement a "Don't ask me again" checkbox on the dialog.

Please check #2898.

selsta:
> It seems that it's not possible to save any value to persistentSettings after wallet "close event" is called.

I don't change any value after the "close event" is called is called.

> That is the reason why I decided to implement a dropdown on Settings > Node page in #2898, and that's why I wasn't able to implement a "Don't ask me again" checkbox on the dialog.

I basically did the same, just on the Settings -> Interface page where it fits better IMO.

plowsof:
The new checkbox in settings -> interface highlighted below:
dprompt
works as intended, ive tried combinations of setting / un-setting / closing / opening wallet and all changes persisted / acted correctly. 

The only improvement i can think of would be to perhaps add something along the lines of ' Disable this prompt in settings -> interface ' here: (a new string to avoid translation nightmares? if this is causing headaches let me know) 
https://github.com/monero-project/monero-gui/blob/b26f38d37ed81e84e51199425a633bc517f7bec2/main.qml#L2090

LGTM

plowsof (approved):

PR #3747, opened 2021-11-23
RadioButton: use native RadioButton Qt Quick control; add tooltip; WizardCreateDevice1: radio buttons in a row + update text + tooltips

rating89us:


plowsof:
ex
ex2

PR #3748, opened 2021-11-23
WizardRestoreWallet1: disable Enter/Return and double space on seed text input area

rating89us:
Recently an user on reddit reported that he/she wasn't able to restore a wallet using mnemonic seed because he/she was inadvertently typing Enter at the end of each line. 

This PR disables Enter and Return keys as well as double space on the seed text input area.


selsta:
I've opened #3958 as a replacement, now it will ignore all whitespaces and newlines. As far as I can see your patch only addresses when entering the seed from the keyboard and not when copy / pasting it.

PR #3749, opened 2021-11-24
LeftPanel: use different color for each account; account card navigation; display account in page headers

rating89us:
[Some users are having trouble](https://www.reddit.com/r/monerosupport/comments/qjww4j/severe_ux_bug/) noticing that a different account was selected, and that selecting an account changes the remaining pages.

Main changes of this PR
- use a different color for each account
- account card: color + mouse scrolling (to navigate accounts) + page indicator
- color for list item indicators on Account and Receive pages
- Receive, Send and Transactions pages header: display "Account #" + account label if wallet contains >1 account

accountcard3


selsta:
That's a cool idea :D One thing I found is that if you have too many accounts the page indicator won't work correctly. Opinion on disabling the page indicator if you have too many accounts? Otherwise it looks like this, first account selected and the page indicator doesn't show anything.

Screenshot 2021-11-25 at 15 52 57

rating89us:
Fixed. Page indicator will not be visible if wallet has more than 40 accounts.

rating89us:
ready for review

Malinero (approved):
Nice feature. Would have preferred with stg like https://github.com/monero-project/monero-gui/commit/3dc33d515cdd6945f431c4fb853ee806de467029, but it can be changed afterwards.

Malinero:
@selsta I think this one can be closed as its rebase has been merged https://github.com/monero-project/monero-gui/pull/4040

PR #3751, opened 2021-11-25
LeftPanel: separate Advanced and settings buttons from account buttons

rating89us:
image


selsta:
Screenshot 2021-11-26 at 19 22 09

I don't know if it makes sense to group Advanced Options and Settings together with the network status, instead of the other pages.

PR #3754, opened 2021-11-27
AdvancedOptionsItem: fix title width

rating89us:


selsta:
I wonder if this is fixable without hardcoding widths for specific languages.

rating89us:
I tried a lot of possibilities, but it was the best solution I could find.

If you use a fixed large width for all languages, many languages will have a too large space between text and buttons.

If you set a warpMode (line break) for the title, the tooltip icon is displayed way too much to the right, and it can't be fixed because the contentWidth doesn't include only the text width ( it includes the blank space past the width).

components/AdvancedOptionsItem.qml
```diff
@@ -14,14 +14,14 @@ RowLayout {
 
     RowLayout {
         id: titlecolumn
-        Layout.alignment: Qt.AlignTop | Qt.AlignLeft
-        Layout.preferredWidth: 195
-        Layout.maximumWidth: 195
+        Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
+        property var language: persistentSettings.language
+        Layout.preferredWidth: language == "Deutsch" ? 280 : language == "Español" ? 250 : language == "Français" ? 235 : 220
+        Layout.maximumWidth: language == "Deutsch" ? 280 : language == "Español" ? 250 : language == "Français" ? 235 : 220
```
mj-xmr, line 20: A nitpick, but since it's equivalent to line 19, how about:

`Layout.maximumWidth: Layout.preferredWidth` ?

mj-xmr (changes requested):

PR #3763, opened 2021-11-30
Wizard: display selected network (testnet/stagenet) on "Create a new wallet" header

rating89us:


rbrunner7 (approved):

selsta:
Opinion on adding the network type afterwards? "Create a new wallet (testnet)" ? Then we have to do less translations of the same string.

rbrunner7:
Sounds good to me. The majority of people probably won't see those slightly less elegant strings with "(testnet)" or "(stagenet)" appended anyway because they won't create such wallets.

rating89us:
Updated

selsta (approved):

PR #3774, opened 2021-12-03
WizardWalletInput: disallow special characters in wallet name

rating89us:
This PR disallows invalid characters/names for filename in Windows + Linux, as well as filename containing "." (Monero GUI doesn't recognize files containing ".").

### DIFF
--- a/wizard/WizardWalletInput.qml
+++ b/wizard/WizardWalletInput.qml
@@ -70,7 +70,7 @@ GridLayout {
                 errorMessageWalletName.text = qsTr("Wallet name is empty") + translationManager.emptyString;
                 return false;
             }
-            if (/[\\\/]/.test(walletName.text)) {
+            if (/^(con|prn|aux|nul|com[0-9]|lpt[0-9])$|([.<>:"\/\\|?*])|(\.|\s)$/ig.test(walletName.text)) {
                 errorMessageWalletName.text = qsTr("Wallet name is invalid") + translationManager.emptyString;
                 return false;
             }


selsta:
What's the source of this regex?

rating89us:
https://stackoverflow.com/a/53635003. 

I added "." to the regex because Monero GUI is not recognizing files containing "."

rbrunner7:
Not sure what you mean with the GUI wallet not being able to work with filenames that contain dots ".".

I just tried on Linux with a wallet file name containing 2 dots and had no problems to create the wallet and later open it again, selected from the filesystem.

And even if there was a problem somewhere with such "dotted" filenames, wouldn't it be better to find that problem and correct it?

I use an extension of ".wallet" for all my wallets. After your PR, would such an extension not be supported anymore? Or more in general, I doubt we want to forbid GUI wallet users the use of extensions.

rating89us:
> Not sure what you mean with the GUI wallet not being able to work with filenames that contain dots ".".

If you create a wallet using a name containing "." (for instance, "my.wallet" or "user@email.com"), the "Open a wallet from file" page will not display it (user has to click on Browse filesystem button to find the wallet).

> wouldn't it be better to find that problem and correct it?

I agree, but I don't know how to solve it yet. This PR would be a temporary "fix" so that users don't have problems of missing wallets.

> I use an extension of ".wallet" for all my wallets. After your PR, would such an extension not be supported anymore?

Extensions still work. Monero GUI only sets the wallet name, it doesn't let the user change the file extension.

rbrunner7:
I made an attempt to find out where the listing of wallets with dots in their name might go wrong but quickly had to acknowledge that I simply do not know enough right now about the structure of the codebase to have a good chance.

So your PR would only influence the creation of new wallets, right? It will still be possible to open existing such "dotted" wallets using the standard Qt file selector? After all it would be unfortunate to make access to those wallets with the GUI wallet app impossible, however few there may be.

selsta:
Closing as the `.` bug was fixed in #3968 and I don't think we need all this Windows specific code.

PR #3783, opened 2021-12-07
Settings, Advanced: mirror stack transition when clicking on a button on the left

rating89us:
mirror

### DIFF
--- a/pages/settings/Settings.qml
+++ b/pages/settings/Settings.qml
@@ -49,6 +49,7 @@ ColumnLayout {
     property alias settingsStateViewState: settingsStateView.state
 
     MoneroComponents.Navbar {
+        id: navbarId
         Layout.alignment: Qt.AlignHCenter
         Layout.topMargin: height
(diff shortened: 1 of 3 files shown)


pages/settings/Settings.qml
```diff
@@ -101,6 +101,14 @@ ColumnLayout {
                     previousView.onPageClosed();
                 }
             }
+            if ((previousView == settingsLayoutView && (currentView == settingsWalletView))  ||
+                (previousView == settingsNodeView   && (currentView == settingsLayoutView || currentView == settingsWalletView)) ||
+                (previousView == settingsLogView    && (currentView != settingsInfoView)) ||
+                (previousView == settingsInfoView)) {
```
selsta, line 107: I've been trying to find a better solution for this, ideally with a position variable, but didn't find a nice solution yet.

rating89us: Using previousIndex and currentIndex now

PR #3785, opened 2021-12-09
[WIP] LeftPanel: sync progress in Windows taskbar button

rating89us:


LeftPanel.qml
```diff
@@ -550,6 +573,15 @@ Rectangle {
...
             connected: Wallet.ConnectionStatus_Disconnected
+            onConnectedChanged: {
+                if (networkStatus.connected == Wallet.ConnectionStatus_Disconnected ||
+                    networkStatus.connected == Wallet.ConnectionStatus_Connecting ||
+                    networkStatus.connected == Wallet.ConnectionStatus_WrongVersion) {
+                    //taskbarButton.stop()
+                } else {
+                    //taskbarButton.resume()
+                }
+            }
```
rating89us, line 584: @selsta 
Could you please check how to implement this in C++?
https://doc.qt.io/qt-5/qwintaskbarprogress.html#stop
https://doc.qt.io/qt-5/qwintaskbarprogress.html#resume

There is no function available in QML (https://doc.qt.io/qt-5/qml-qtwinextras-taskbarbutton.html#progress.stopped-prop)

selsta: What happens when you stop the taskbarButton? What's the difference to setting visible to false?

rating89us: When you stop, the progress bar in taskbar button turns red, indicating an error (in this case, a disconnection, but we could also add other errors in future, like not enough space in disk).

image

When visible is false, the progress bar doesn't appear on taskbar button.

selsta: See e.g. https://code.woboq.org/qt5/qtwinextras/examples/winextras/quickplayer/qml/main.qml.html#91

rating89us: taskbarButton.stopped and taskbarButton.paused don't work:
` qrc:/LeftPanel.qml:580: Error: Cannot assign to non-existent property "stopped"`
See: https://www.qtcentre.org/threads/62358-QtWinExtras-TaskbarButton-Progress-not-turning-red

selsta: Did you try `taskbarButton.progress.stop()`?

rating89us: You're right, it works. I was using taskbarButton.stop().

selsta (commented):
QtWinExtras only exists on Windows, so compilation fails on Mac and Windows.

Something like this will be required: https://stackoverflow.com/a/52434062

See also https://stackoverflow.com/questions/29378959/how-to-use-qtwinextras-in-cross-platform-application

rating89us:
Fixed. Tested in Linux and Windows.

selsta:
https://github.com/monero-project/monero-gui/blob/master/CMakeLists.txt#L149-L151

Can you try adding

```
if(WIN32)
    list(APPEND QT5_LIBRARIES Qt5WinExtras)
endif()
```

It seems to work without it for some reason but it seems wrong.

rating89us:
> Can you try adding
> 
> ```
> if(WIN32)
>     list(APPEND QT5_LIBRARIES Qt5WinExtras)
> endif()
> ```
> 
> It seems to work without it for some reason but it seems wrong.

I added these lines and it stopped building docker-windows-static:

```
CMake Error at CMakeLists.txt:183 (find_package):
  By not providing "FindQt5WinExtras.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "Qt5WinExtras", but CMake did not find one.

  Could not find a package configuration file provided by "Qt5WinExtras" with
  any of the following names:

    Qt5WinExtrasConfig.cmake
    qt5winextras-config.cmake

  Add the installation prefix of "Qt5WinExtras" to CMAKE_PREFIX_PATH or set
  "Qt5WinExtras_DIR" to a directory containing one of the above files.  If
  "Qt5WinExtras" provides a separate development package or SDK, be sure it
  has been installed.


-- Configuring incomplete, errors occurred!
See also "/monero-gui/build/x86_64-w64-mingw32/release/CMakeFiles/CMakeOutput.log".
See also "/monero-gui/build/x86_64-w64-mingw32/release/CMakeFiles/CMakeError.log".
make: *** [Makefile:38: depends] Error 1
Error: Process completed with exit code 2.
```

PR #3788, opened 2021-12-10
main: check if currentWallet is loaded before calling it

rating89us:
When daemon is started/stopped while the wallet is being closed, the following error appears:
`W qrc:/main.qml:733: TypeError: Cannot call method 'connected' of undefined`


selsta (approved):
Never saw that message but patch looks good.

rating89us:
Open a wallet and rapidly click on "close wallet" button. If the daemon finishes startup while wallet is closing, the error appears.

PR #3792, opened 2021-12-11
LeftPanel: auto refresh Transactions page during wallet scan

### LINKED ISSUE #3134: History: transaction history page is not refreshed during wallet sync
While the wallet is syncing, the user must leave Transactions page (for example, enter in Receive page) and open it again in order to refresh the recently synced received/spent transactions.

rating89us:
Closes #3134

Transactions page will auto refresh every time balance is updated.

### DIFF
--- a/LeftPanel.qml
+++ b/LeftPanel.qml
@@ -63,6 +63,13 @@ Rectangle {
     signal addressBookClicked()
     signal accountClicked()
 
+    onBalanceStringChanged: {
+        if (middlePanel.state == "History") {
+            middlePanel.historyView.onPageClosed()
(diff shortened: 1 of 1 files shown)


plowsof:
I had a look around in the source and found that by removing the "do not signal on received tx if wallet is not syncronized completely" functions (received/unconf/spent) , the Monero GUI will update the balance/history list "live" without the need for us to leave / return 

https://github.com/monero-project/monero/blob/66184f30859796f3c7c22f9497e41b15b5a4a7c9/src/wallet/api/wallet.cpp#L168

PR #3793, opened 2021-12-11
LeftPanel: display notification in account card when balance changes

rating89us:


LeftPanel.qml
```diff
@@ -33,6 +33,7 @@ import moneroComponents.Wallet 1.0
 import moneroComponents.NetworkType 1.0
 import moneroComponents.Clipboard 1.0
 import FontAwesome 1.0
+import QtMultimedia 5.15
```
selsta, line 36:
```suggestion
import QtMultimedia 5.12
```

AFAIK Ubuntu CI still uses Qt 5.12

PR #3795, opened 2021-12-11
wizard: clear password field after wallet creation

### LINKED ISSUE #3794: GUI Not Clearing Password After Wallet Creation
When using the GUI to create a set of wallets, the password used for the initial wallet was still filled in on the 'Give your wallet a password' screen when creating a second wallet. This is repeatable, in that if a second wallet is created it's password will still be entered when creating a third.
This was observed using the latest release (0.17.3.0) on Windows.

selsta:
Closes #3794

### DIFF
--- a/wizard/WizardAskPassword.qml
+++ b/wizard/WizardAskPassword.qml
@@ -104,7 +104,6 @@ ColumnLayout {
             labelFontSize: 14
             password: true
             labelText: qsTr("Password") + translationManager.emptyString
-            text: walletOptionsPassword
         }
 
         ColumnLayout {
@@ -170,7 +169,6 @@ ColumnLayout {
             labelFontSize: 14
             passwordLinked: passwordInput
             labelText: qsTr("Password (confirm)") + translationManager.emptyString
-            text: walletOptionsPassword
             onTextChanged:{
                 if (passwordInputConfirm.text.length == passwordInput.text.length) {
                     firstUserInput = false;
--- a/wizard/WizardCreateWallet1.qml
+++ b/wizard/WizardCreateWallet1.qml
@@ -194,6 +194,8 @@ Rectangle {
                 btnNext.enabled: walletInput.verify();
                 btnPrev.text: qsTr("Back to menu") + translationManager.emptyString
                 onPrevClicked: {
+                    wizardController.wizardStateView.wizardCreateWallet2View.pwField = "";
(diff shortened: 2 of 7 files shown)


rating89us:
See #3796

rating89us:
Two issues:
- Bar color rapidly changes to red before the next page is displayed:
barcolor_

- If the user clicks on Previous button on the next page (Daemon settings), password fields will be blank, and the user will have to type the long password again. IMO the password fields should remain filled when the user goes back, so I suggest cleaning the fields only on the last page (when the "Create wallet" button is clicked).

selsta:
> Bar color rapidly changes to red before the next page is displayed

Same happens when pressing on the back button. But IMO that's not too important now, I fixed it for the forward button.

> when the "Create wallet" button is clicked

Done (Yes the code turned out a bit ugly)

rating89us:
I haven't tested it yet, but we should probably clean password fields when the user is on the first page and clicks on back button returning to the main menu. Wizard controller qml file only resets the password variable (walletOptionsPassword), it doesn't clean the password text fields.

selsta:
> I haven't tested it yet, but we should probably clean password fields when the user is on the first page and clicks on back button returning to the main menu.

I don't see a case where there can be anything in the password field in this scenario. Do you have anything specific in mind?

rating89us:
User A types password in page 2, returns to page 1 to change wallet name or review seed, but gives up creating a wallet and goes back to main menu.

User A or B enters in create a new wallet wizard, and password fields will still be filled.

selsta:
We do clear the password field when going from page 2 to page 1.

rating89us:
Sorry, I misread the code. Will test it soon.

selsta:
I think you added that a couple weeks ago. I didn't touch it anymore, but from my testing it works fine.

rating89us (approved):

selsta:
@rating89us I could change it so that the password clears from going to page 1 to main menu instead of page 2 to page 1. Opinion?

That would also solve "Bar color rapidly changes to red before the next page is displayed"

rating89us:
I think it's ok. It seems a better idea if we let the user navigate between wizard pages without having to rewrite password or daemon configurations, and we only reset the fields when leaving the wizard (either going back to main menu on page 1 or creating a wallet on last page).

selsta:
@rating89us updated

rating89us (approved):

PR #3796, opened 2021-12-11
WizardCreateWallet2, WizardRestoreWallet2: clean password fields when clicking on back button

### LINKED ISSUE #3794: GUI Not Clearing Password After Wallet Creation

rating89us:
Closes #3794

This bug was introduced in #3755 (property binding was lost when setting password fields to blank).

### DIFF
--- a/wizard/WizardCreateWallet2.qml
+++ b/wizard/WizardCreateWallet2.qml
@@ -39,6 +39,11 @@ Rectangle {
     property alias pageHeight: pageRoot.height
     property string viewName: "wizardCreateWallet2"
 
+    function onPageCompleted() {
+        passwordFields.password = Qt.binding(function() { return wizardController.walletOptionsPassword })
+        passwordFields.passwordConfirm = Qt.binding(function() { return wizardController.walletOptionsPassword })
+    }
+
     ColumnLayout {
         id: pageRoot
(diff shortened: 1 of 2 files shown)


selsta (commented):
I don't fully understand this. Where is the property binding in the first place?

rating89us:
Lines 107 and 173 of WizardAskPassword.qml

selsta:
What's the point of the property binding in the first place? These are textboxes so why would they be binded to something else?

rating89us:
I think it's because WizardController has a restart function that cleans wizardController.walletOptionsPassword and other variables, it doesn't touch text fields.

rating89us:
I mean I don't know exactly why it was implemented this way. My PR just fixes the bug, reverts the code to how it was working before (with property binding) as well as does what I wanted to do initially (clean password fields when clicking on back button).

selsta:
I have updated #3795 to remove the property binding. Your solution is insofar correct that it restores previous behaviour but in my opinion the property binding was incorrect in the first place.

PR #3798, opened 2021-12-12
SettingsWallet, SettingsListItem: disable "Rescan wallet balance" button if wallet is not synced

rating89us:
Monero GUI freezes if "Rescan wallet balance" button is clicked while wallet is being synced.

### DIFF
--- a/pages/settings/SettingsWallet.qml
+++ b/pages/settings/SettingsWallet.qml
@@ -91,6 +91,7 @@ Rectangle {
         }
 
         MoneroComponents.SettingsListItem {
+            enabled: leftPanel.progressBar.fillLevel == 100
             iconText: FontAwesome.repeat
             description: qsTr("Use this feature if you think the shown balance is not accurate.") + translationManager.emptyString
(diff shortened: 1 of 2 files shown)


pages/settings/SettingsWallet.qml
```diff
@@ -91,6 +91,7 @@ Rectangle {
         }
 
         MoneroComponents.SettingsListItem {
+            enabled: leftPanel.progressBar.fillLevel == 100
```
selsta, line 94:
```suggestion
            enabled: appWindow.walletSynced
```

did you try that?

rating89us: Yes, but `appWindow.walletSynced` is not always correct. It is `true `when you change wallet restore height and the wallet is synchronizing again.

selsta (approved):

pages/settings/SettingsWallet.qml
selsta: I guess ideally that would be fixed, but we can merge this for now.

PR #3802, opened 2021-12-13
SettingsWallet: improve and translate trusted remote node error message

rating89us:


selsta (approved):
I don't really like this method (comparing strings) but there is no error code returned so I guess it's fine for frequent errors.

PR #3815, opened 2022-01-07
LeftPanel: use correct shortcut modifier for Mac

selsta:


rbrunner7:
So that "^" modifier key already works on the Mac without further changes, the only job left to do is making the on-screen labelling correct? Or is that change included in another PR?

selsta:
correct

rbrunner7 (approved):

PR #3823, opened 2022-01-16
WizardController: recommend user to export Ledger view key

rating89us:


selsta (approved):

selsta (commented):
Screenshot 2022-02-01 at 04 10 51

It doesn't fit by default

wizard/WizardController.qml
```diff
@@ -434,7 +434,7 @@ Rectangle {
 
     function creatingWalletDeviceSplash(){
         var splashMsg = qsTr("Creating wallet from device...");
-        splashMsg += wizardController.walletOptionsDeviceName === "Ledger" ? qsTr("\n\nPlease check your hardware wallet –\nyour input may be required.") : "";
+        splashMsg += wizardController.walletOptionsDeviceName === "Ledger" ? qsTr("\n\nPlease check your hardware wallet: –\nit is strongly recommended to export the view key.") : "";
```
selsta, line 437:
```suggestion
        splashMsg += wizardController.walletOptionsDeviceName === "Ledger" ? qsTr("\n\nPlease check your hardware wallet –\nand export the view key.") : "";
```

As far as I know it's broken without exporting the view key so I would just use this. Also the `:` you added seems wrong.

PR #3828, opened 2022-01-21
Advanced: ReserveProof: Add support for reserve proof

### LINKED ISSUE #3826: [Feature Request] Implement Reserve Proofs
Looks like this got dropped during review of PR. 1508 was moved to 1927, last comment 09/25/2020. I would like to take a shot at it if seems like something worth implementing. 

Reference:
#1508
#1927

reemuru:
Closes #3826 

This change adds the ability to prove and check a reserve proof
to this existing "Prove/check" advanced menu. If the amount line has
a valid amount less than that of the current account index this input
will drive the logic for generating a reserve proof. Checking a reserve
proof is accomplished by validating the address and verifying that
signature.indexOf("ReserveProofV") === 0. The result displays the total
and spent amount of the proof.

reserve_proof_layout

reserve_proof_test

check_reserve_proof_layout

check_reserve_proof

### DIFF
--- a/pages/TxKey.qml
+++ b/pages/TxKey.qml
@@ -60,13 +60,13 @@ Rectangle {
             MoneroComponents.Label {
                 id: soloTitleLabel
                 fontSize: 24
-                text: qsTr("Prove Transaction / Reserve") + translationManager.emptyString
+                text: qsTr("Prove Transaction") + " / " + qsTr("Reserve") + translationManager.emptyString
             }
 
             MoneroComponents.TextPlain {
                 Layout.fillWidth: true
                 text: qsTr("Generate a proof of your incoming/outgoing payment by supplying the transaction ID, the recipient address and an optional message. \n" +
-                           "For the case of outgoing payments, you can get a 'Spend Proof' that proves the authorship of a transaction. In this case, you don't need to specify the recipient address. \nFor reserve proofs you don't need to specify tx id or address.") + translationManager.emptyString
+                           "For the case of outgoing payments, you can get a 'Spend Proof' that proves the authorship of a transaction. In this case, you don't need to specify the recipient address.") + qsTr("\nFor reserve proofs you don't need to specify tx id or address.") + translationManager.emptyString
                 wrapMode: Text.Wrap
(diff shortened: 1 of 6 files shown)


pages/TxKey.qml
```diff
@@ -133,12 +166,12 @@ Rectangle {
...
                 id: soloTitleLabel2
                 fontSize: 24
-                text: qsTr("Check Transaction") + translationManager.emptyString
+                text: qsTr("Check Transaction / Reserve") + translationManager.emptyString
             }
 
             MoneroComponents.TextPlain {
                 text: qsTr("Verify that funds were paid to an address by supplying the transaction ID, the recipient address, the message used for signing and the signature.\n" +
-                           "For the case with Spend Proof, you don't need to specify the recipient address.") + translationManager.emptyString
+                           "For the case with Spend Proof, you don't need to specify the recipient address.\nTransaction is not needed for reserve proof.") + translationManager.emptyString
```
selsta, line 174: Could you add this as a separate `qsTr()` so that we don't lose translations on the old string? Same with other strings, wherever possible.

pages/TxKey.qml
```diff
@@ -222,7 +255,7 @@ Rectangle {
                 font.family: MoneroComponents.Style.fontRegular.name
                 font.pixelSize: 14
                 color: MoneroComponents.Style.defaultFontColor
-            }
+            }
```
selsta, line 258: looks unintended

selsta (commented):
Looks good, thank you. Left two minor comments.

pages/TxKey.qml
reemuru: Thanks for reviewing @selsta! I have updated on commit 2bf0dd840f907b406271863540c8efe002b2cdc3 so that previous translations are not broken. Also removed the erroneous whitespace.

pages/TxKey.qml
reemuru: Oops. Not sure what happened there. Fixed on 2bf0dd840f907b406271863540c8efe002b2cdc3

selsta (approved):

PR #3831, opened 2022-02-01
bug: reset wallet path on network change 

### LINKED ISSUE #3818: Error when opening mainnet wallet when Monero GUI is starting in stagenet mode
Steps to reproduce:
- Open a mainnet wallet
- Close the wallet and return to main menu
- In advanced options, choose stagenet in network type
- Close Monero GUI
- Reopen Monero GUI
- Monero GUI will try to open the last opened wallet (which is a mainnet wallet) while stagenet is selected, displaying the error below:
image

reemuru:
Closes #3818

When changing network type without changing
wallet path the gui will attempt to open the
most recently used wallet path and network type
even if they don't match. This commit forces the
user back to wallet setup under this condition.
As long as network type and wallet path are both
updated before closing the issue will no longer occur.

### DIFF
--- a/wizard/WizardHome.qml
+++ b/wizard/WizardHome.qml
@@ -213,6 +213,7 @@ Rectangle {
                         }
                         appWindow.disconnectRemoteNode()
                         networkTypeDropdown.currentIndex = Qt.binding(function() { return persistentSettings.nettype });
+                        persistentSettings.wallet_path = ""
                     }
                 }
 


mj-xmr (approved):

selsta (approved):
It's not perfect because now the GUI resets to the home screen that should only display on first start. But using a persistent first start variable also isn't a nice solution. I guess changing network mode is something that isn't too frequent so this should be fine.

PR #3838, opened 2022-02-13
settings: bind return key after password change

### LINKED ISSUE #3836: Password Changed Successfully OK with Keyboard
Pressing enter on this screen should be binded to OK but its not. It forces users to use their mouse even though they just used their keyboard on the previous screen.
image

reemuru:
Closes #3836 

The enter / return key is not bound to the information
popup that appears after a successful password change.
The functionality exists on the previous menu. Make it
possible to use the return key on this popup as well.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1692,6 +1692,9 @@ ApplicationWindow {
             informationPopup.open();
         }
         onRejectedNewPassword: {}
+        Keys.enabled: !passwordDialog.visible && informationPopup.visible
+        Keys.onEnterPressed: informationPopup.close()
+        Keys.onReturnPressed: informationPopup.close()
     }
 
     DevicePassphraseDialog {


main.qml
```diff
@@ -1692,6 +1692,9 @@ ApplicationWindow {
             informationPopup.open();
         }
         onRejectedNewPassword: {}
+        Keys.enabled: !passwordDialog.visible
```
selsta, line 1695:
```suggestion
        Keys.enabled: !passwordDialog.visible && informationPopup.visible
```

I'm not sure how this Keys.enabled things works honestly, but does it make sense to add this? So that .close() can't be called if the popup isn't open.

reemuru: @selsta I was trying to figure out how it works as well. This makes more sense to me. Let me update and retest.

reemuru:
> Do you know how to squash the commits?
@selsta 
Yeah I can squash them. I've never had this co-authored message. Should  I keep that in the commit message?

reemuru:
> It doesn't matter.

@selsta squashed to [ddd95f7](https://github.com/monero-project/monero-gui/pull/3838/commits/ddd95f73b4e6570239d2ce4647db58dd6700f54d)

selsta (approved):
Thanks

PR #3846, opened 2022-02-28
Wizard: fix stagenet approx blockheight

selsta:


selsta:
@mj-xmr please review, it's basically the equivalent to https://github.com/monero-project/monero/pull/8195

mj-xmr (approved):
What's that unrelated failed Windows build about? Does the build behave differently on various machines?

selsta:
The CI run is from 1.5 months ago, which currently had broken Windows build. It's unrelated to the patch.

PR #3853, opened 2022-03-07
Update languages.xml

whirlwind110:
Modifying the Chinese brackets symbol and area description and the flag correctly.

### DIFF
--- a/lang/languages.xml
+++ b/lang/languages.xml
@@ -37,8 +37,8 @@ Lojban
         <language display_name="Português (PT)" locale="pt-pt_PT" wallet_language="Português" flag="/lang/flags/pt.png" qs="none"/>
         <language display_name="Português (BR)" locale="pt-br_BR" wallet_language="Português" flag="/lang/flags/br.png" qs="none"/>
         <language display_name="Русский язык" locale="ru_RU" wallet_language="русский язык" flag="/lang/flags/ru.png" qs="none"/>
-        <language display_name="简体中文 (中国)" locale="zh-cn_CN" wallet_language="简体中文 (中国)" flag="/lang/flags/cn.png" qs="none"/>
-        <language display_name="繁體中文 (台灣)" locale="zh-tw_CN" wallet_language="English" flag="/lang/flags/tw.png" qs="none"/>
+        <language display_name="简体中文（中国）" locale="zh-cn_CN" wallet_language="简体中文（中国）" flag="/lang/flags/cn.png" qs="none"/>
+        <language display_name="繁體中文（中國港澳台地區）" locale="zh-tw_CN" wallet_language="繁體中文（中國港澳台地區）" flag="/lang/flags/cn.png" qs="none"/>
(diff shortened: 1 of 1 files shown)


selsta:
Hello, can you explain your changes in more detail? Why the flag change?

whirlwind110:
> 

Traditional Chinese characters are common use in Taiwan, Hong Kong, and Macau. They are different area of China.  We should use a country flag to instead.

selsta (commented):
We will likely remove all flags in the future to avoid any ambiguity. Regarding your text changes, I will have to get this confirmed first by a native speaker.

whirlwind110:
> We will likely remove all flags in the future to avoid any ambiguity. Regarding your text changes, I will have to get this confirmed first by a native speaker.

Sure, I am also a Chinese native speaker in the UK. Removing all flags is a sensible way because Taiwan is a province rather than a country. Monero should avoid being implicated in political conflicts.

CryptoGodfatherVA23 (approved):

CryptoGodfatherVA23 (approved):

mj-xmr (approved):
I know a little bit of the history of the region and I'd tend to agree with the change. Keeping it like it was could lead to Monero software being targeted for political reasons in Taiwan.

The change translates to:
"China Hong Kong, Macao and Taiwan regions" ([via Google Translate](https://translate.google.com/?sl=auto&tl=en&text=%E4%B8%AD%E5%9C%8B%E6%B8%AF%E6%BE%B3%E5%8F%B0%E5%9C%B0%E5%8D%80&op=translate))

which treats each provinces equally.

selsta:
Definitely not going to merge this unless confirmed by a native speaker.

mj-xmr:
I'm perfectly fine with not merging it, as long as the final state of this (unintentionally) controversial area of Monero ends up in removing all the flags and territorial claims completely. I don't mind if this doesn't happen by tomorrow morning.

I'm a pacifist and against all wars or insurrections but all that I can really do against it, is improving Monero.

erciccione (changes requested):
NACK

Merging this PR will cause a political debate about the political status of Taiwan. Something we should avoid like the plague.

There is no way to resolve this matter without pissing off one faction or the other. I suggest to not merge this PR, but instead focus on removing flags entirely. The issue would not be resolved completely, because the 2 factions refer to the island in two different ways, but getting rid of flags would be a great improvement that will definitely limit future issues.

mj-xmr:
Exactly

whirlwind110:
> NACK
> 
> Merging this PR will cause a political debate about the political status of Taiwan. Something we should avoid like the plague.
> 
> There is no way to resolve this matter without pissing off one faction or the other. I suggest to not merge this PR, but instead focus on removing flags entirely. The issue would not be resolved completely, because the 2 factions refer to the island in two different ways, but getting rid of flags would be a great improvement that will definitely limit future issues.

Many thanks!

PR #3859, opened 2022-03-13
SettingsWallet: lock wallet on demand

### LINKED ISSUE #3857: [Feature] Lock Wallet & Bind it with Ctrl + L
image
This Wallet tab under Settings should allow the user to lock the wallet. This is my workflow:
I log into my computer, the Monero-GUI is automatically opened (I have a shortcut in the Startup folder), and then I log in to start sync. I then create a new desktop to work. The first desktop has the GUI unlocked since there's no easy way to lock it. I'd have to set the timeout to a really low number and there are cases where a low timeout hinders UX. Therefore, a way to lock the wallet on demand would be really nice.

reemuru:
Closes #3857 
This commit adds the ability to lock the wallet on demand instead
of waiting for the user inactivity time out. It is binded to the
hot keys Ctrl+L. Pressing cancel or the esc key will send the user
back to the wizard home for wallet selection. Incorrect password
returns the error message. Correct password will remove the
PasswordDialog allowing access. Add lock functionality in the
wallet settings section and title bar.

fix-3857

### DIFF
--- a/components/SettingsListItem.qml
+++ b/components/SettingsListItem.qml
@@ -7,6 +7,7 @@ import "../components" as MoneroComponents
 ColumnLayout {
     id: settingsListItem
     property alias iconText: iconLabel.text
+    property alias symbol: symbolText.text
     property alias description: area.text
     property alias title: header.text
     property bool isLast: false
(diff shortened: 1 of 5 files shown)


reemuru:
This should have a button like the `Close this wallet` and have the binding show up there when pressing `ctrl`.

selsta (approved):

PR #3096, opened 2020-09-20
AddressBook: description edits, remove hack

selsta:
Requires https://github.com/monero-project/monero/pull/6828

PR #3863, opened 2022-03-15
AddressBook: fix index 0 description bug

reemuru:


selsta:
any reason for this being WIP?

selsta:
I added that code here: https://github.com/monero-project/monero-gui/pull/3096/

How does editing a description work now with it removed? (I'm on my phone currently otherwise I would check myself)

selsta (commented):
Ok, was able to test it. This breaks editing descriptions.

selsta:
#3865 solved this issue for me, could you try it to confirm? :)

reemuru:
fixed by https://github.com/monero-project/monero-gui/pull/3865

PR #3865, opened 2022-03-16
AddressBook: fix adding new entry

### LINKED ISSUE #3866: adding entry to address book edits description of another entry
Adding an entry to the address book in Monero GUI version 0.17.3.1 will result in not only an entry being added, but it will also cause the description of the bottom entry in the address book to change to the description of the newly added entry.

selsta:
Closes #3866 

If `currentWallet.addressBook.addRow` was successful the code would go into the edit description branch which is not what we want.


Arimfexendrapus:
thanks for doing this guys, sorry I put it in the wrong place to begin with and then didn't look to see if it was here

PR #3878, opened 2022-04-06
wizard: redesign seed page

selsta:
- move mnemonic seed
- restore height into a separate page
- pdf template
- seed verification
- responsive UI
- accessibility

Rebased from here: #3677


GBKS:
Just got a notification on this. Sorry, have not kept with the previous issue. Design-wise do you need anything for the PDF template or otherwise?

selsta:
@GBKS I think we have everything we need :) If not, I'll follow up on your but as far as I can see the PDF we have looks good.

plowsof:
The usability of this gets a thumbs up from me, i printed the template out and wrote the 25 words on it. Its quite painless, and with the verify screen having the same layout as the pdf template, you can find the missing words quickly, and after you type correctly the green tick is shown and your cursor is focused onto the next word automatically. Mistakes can be corrected also.

Thanks for rebasing!

plowsof (approved):

plowsof (approved):

selsta:
@GBKS We would need a new version of the template for Polyseed, do you still have the the original file? If yes, could you create a new version for us that has the following

- 16 words
- Wallet creation date (this is not needed for restore, but it might be useful for organizing. Do Bitcoin seed templates usually have such a field)?

otherwise identical. Thank you!

PR #3883, opened 2022-04-08
Makefile: set CMAKE_BUILD_TYPE to debug for debug target

selsta:
https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

> Typical values include Debug, **Release**, RelWithDebInfo and MinSizeRel, but custom build types can also be defined.

> Depending on the situation, the value of this variable may be treated **case-sensitively** or case-insensitively.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ debug-static-win64:
 
 debug-static-mac64:
 	mkdir -p $(builddir)/debug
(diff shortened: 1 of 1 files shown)


selsta:
@mj-xmr please review

mj-xmr (approved):
It looks ok from the perspective of the task and its description, but could you explain why we are setting `Release` in the first place, while the directory name and the target name are both `debug`?

selsta:
Good question, I just checked for typos. Will update to debug.

selsta:
@mj-xmr updated

mj-xmr (approved):

PR #3885, opened 2022-04-11
main: don't async close wallet

### LINKED ISSUE #3840: Closing GUI After Wallet Open for a While Results in Infinite Loading Animation
OS: Windows

If I open a wallet at the start of the day, and close it at the end of the day (using Alt + F4), I get a "Closing wallet" loading screen that never goes away. I have to alt-f4 again to close the UI but I still see the process running in Task Manager
image

This issue cannot be reproduced by simply opening and closing the wallet. It happens if I have the wallet open for hours. It may be able to reproduce if you have the wallet open for 10 minutes, but I'm not sure.

Furthermore, when I log out the wallet, I get presented with the Welcome to Monero screen instead of the select wallet screen. I know it's a separate issue, but usually when people log out of one wallet, it's to open the other wallet, not to create a new wallet or something, because that's a one time thing, whereas the open another wallet happens everytime after.

Another feature request is to add a wallet_name[view-only].monero_gui file to the wallet directories that can be double clicked.

Another Bug, is clicking on monero: links opens a new instance of the GUI and doesn't even auto fill the data. What I expect is for the existing GUI to focus and auto fill the data or at least the new instance to autofill the data. I think you could use IPC to inform the existing instance of the data. In my application, I make a post request to 127.0.0.1 with the relevant data.

selsta:
Fixes #3840

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1123,21 +1123,20 @@ ApplicationWindow {
     // close wallet and show wizard
     function showWizard(){
         walletInitialized = false;
-        closeWallet(function() {
-            wizard.restart();
-            wizard.wizardState = "wizardHome";
-            rootItem.state = "wizard"
-            // reset balance, clear spendable funds message
-            clearMoneroCardLabelText();
-            leftPanel.minutesToUnlock = "";
-            // reset fields
-            middlePanel.addressBookView.clearFields();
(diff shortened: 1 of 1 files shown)


selsta:
Closing this now as it did not fix the issue.

PR #3888, opened 2022-04-22
FutureScheduler: delete unused function definitons

selsta:
I think they got replaced by

```
    QPair<bool, QFuture<void>> run(std::function<void()> function) noexcept;
    QPair<bool, QFuture<QJSValueList>> run(std::function<QJSValueList()> function, const QJSValue &callback);
```

and never removed.


mj-xmr (approved):
A small nitpick: these are declarations, not definitions.

The definitions aren't present in the `FutureScheduler` class anymore, so removing the declarations is a fine way to go.

selsta:
@mj-xmr fixed

mj-xmr (approved):

PR #3889, opened 2022-04-23
Mining: only update mining status when page is open

selsta:
Due to a bug the GUI was constantly checking the mining status every two seconds, no matter if the mining page is open or not.

### DIFF
--- a/pages/Mining.qml
+++ b/pages/Mining.qml
@@ -575,18 +575,10 @@ Rectangle {
 
     Timer {
         id: timer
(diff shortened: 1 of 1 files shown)


mj-xmr (approved):
The logic makes sense, plus there's additional sanity check `currentWallet !== undefined`, but I would need a couple of hours today to validate if it actually works, unless you're absolutely sure of this.

selsta:
I did use some extra async logging to confirm that everything works fine.

PR #3897, opened 2022-04-27
TxConfirmationDialog: warn high fees

selsta:
Screenshot 2022-04-27 at 17 42 26

### DIFF
--- a/components/TxConfirmationDialog.qml
+++ b/components/TxConfirmationDialog.qml
@@ -338,8 +338,9 @@ Rectangle {
                 spacing: 16
 
                 Text {
-                    color: MoneroComponents.Style.defaultFontColor
-                    font.pixelSize: 15
+                    property bool maliciousTxFee: parseFloat(root.transactionFee) > 0.01
+                    color: maliciousTxFee ? "red" : MoneroComponents.Style.defaultFontColor
+                    font.pixelSize: maliciousTxFee ? 20 : 15
                     text: {
                         if (currentWallet) {
                             if (!root.transactionFee) {
@@ -349,7 +350,7 @@ Rectangle {
                                     return qsTr("Calculating fee") + "..." +  translationManager.emptyString;
                                 }
                             } else {
-                                return root.transactionFee + " XMR"
+                                return root.transactionFee + " XMR" + (maliciousTxFee ? " (HIGH FEE)" : "")
                             }
                         } else {
                             return "";


reemuru (commented):
LGTM, compile tested and triggered high fee display

pr3897

reemuru (approved):

jeffro256 (approved):

jeffro256:
This shouldn't be the only long-term solution, but it's a nice quick fix for now!

reemuru:
> This shouldn't be the only long-term solution, but it's a nice quick fix for now!

@jeffro256 
agree, i can create and issue: 

`"Update Magic for high fee"`   ?

selsta:
The only long term solution will be to remove the remote node scanner from simple mode and use hardcoded trusted nodes.

jeffro256:
@selsta Or have nodes provide some kind of blockchain proof that their recommended fee is reasonable

SamsungGalaxyPlayer:
Apologies for being out of the loop here, but some sanity checks:

1. Does the current code show this warning if fees are > 0.01 XMR?
2. Does the current code check if the fee follows an appropriate multiplier set in the wallet defaults? We don't want wallets tricked to send unique fee values either.

@selsta

selsta:
> Does the current code show this warning if fees are > 0.01 XMR?

Yes, I changed it for the screenshot to demonstrate it. I'm open for different values.

> Does the current code check if the fee follows an appropriate multiplier set in the wallet defaults? We don't want wallets tricked to send unique fee values either.

As far as I understand it the wallet calculates the fee with data from the daemon. If the daemon tricks you like in this case I'm not sure what you want to do on wallet side.

plxkax:
> The only long term solution will be to remove the remote node scanner from simple mode and use hardcoded trusted community nodes.

It would be good to offer the choice between remote node or having the trusted community node. You don't want to make it feel like it's being forced, user should have the option to decide.

selsta:
You can either use simple mode, which automatically will connect you to a community node or you can use advanced mode which allows you to set your own remote node.

plxkax:
> 

Would there be any security compromises with automatically having all simple mode users onto community node only?

selsta:
Security compromises compared to now? No. Simple mode already connects to a random remote node.

plxkax:
To confirm, **Simple mode currently connects to a random remote node automatically** but with this change **Simple mode will connect to a community trusted node automatically** with no security compromises compared to now?

plxkax:
Do we have an any idea as to how big the community trusted node list will be?

selsta:
Correct.

selsta:
> Do we have an any idea as to how big the community trusted node list will be?

Around 10 maybe? Don't know yet.

nixt7:
It's GUI discussed here. But what about wallet RPC? It needs protection from high fees too.

selsta:
The GUI simple mode currently connects you to a random remote node that's why it was important to add the warning here quickly. The RPC wallet doesn't have such a feature so it is way more unlikely that you run into the high fee issue in the first place, unless you manually connect to a malicious remote node.

fugbixer:
maybe adding a won't proceed when the xmr fee is above a certain unreasonable amount like 0.5+ would be nice to make it safe. even when i don't like of adding a limit.

It could need something that's complete obvious for non technical users to see something is wrong and they should not continue.

PR #3902, opened 2022-04-30
WizardCreateDevice1: set default restore height to 1

selsta:
Due to API reasons height 0 means to approximate the current block height during hardware device restore. That's not what we want by default. Until this gets changed upstream we will set the default to 1.

https://github.com/monero-project/monero/blob/v0.17.3.2/src/wallet/api/wallet_manager.cpp#L145-L149


selsta:
@mj-xmr please review

mj-xmr (approved):

PR #3903, opened 2022-04-30
WizardCreateDevice1: add Ledger Nano S Plus

selsta:
Screenshot 2022-04-30 at 09 05 32

Waiting on feedback from Ledger if we are allowed to use this picture as I'm not sure about the license.

Edit: No reply from Ledger. All other pictures had the correct license so I say we can merge this. In the unlikely case they decide against it we can remove it again.


selsta:
@mj-xmr please review

guofuhua0930:
Version v0.17.3.2 of Monero GUI software still does not support Ledger Nano S Plus

selsta:
@guofuhua0930 What exactly is not supported? You have to select "Ledger Nano S" in the menu, but otherwise it should work fine. One person on Reddit got it working.

guofuhua0930:
ok, it worked, thanks

mj-xmr (approved):

PR #3905, opened 2022-04-30
js: minimize frameless window when clicking on taskbar icon

### LINKED ISSUE #3653: [Windows] GUI Does Not Minimize On Taskbar Icon Click

selsta:
Fixes #3653


selsta:
@elibroftw can you check if this works on Windows? https://github.com/monero-project/monero-gui/suites/6326753598/artifacts/227492115

elibroftw:
Nope

elibroftw:
As in it doesn't work

selsta:
@elibroftw patch was incomplete, next try: https://github.com/monero-project/monero-gui/suites/6337125518/artifacts/228213660

elibroftw:
works.

selsta:
ok, please approve it and I'll add it to the merge queue

elibroftw (approved):

PR #3913, opened 2022-05-06
p2pool: fix monerod sometimes restarting with invalid arguments

### LINKED ISSUE #3911: monerod second instance started by GUI
Linux x64
Ubuntu
Monero GUI 0.17.3.2 Mining monerod p2pool second daemon

The case is as follows:

The monerod is started a boot time. NO default directories used!

My lmdb is 100% synchronized. When I start the Monero GUI it connects to my local node. So far so good. 

The GUI connects to the already running monerod flawless and all is fine.

But when I start the mining by the GUI a second instance of the monero daemon seems to  initiated. 
I selected p2pool mining with main.
This second daemon is started with the wrong directory for the database. Default settings are used.
That brings it to the negative effect that the complete lmdb will be created (downloaded) again.

The problem I described above is repeatable.

My question is:

Why is the Monero GUI trying to establish a second daemon even it has connected successfully to daemon already running?

devhyper:
Resolves #3911

### DIFF
--- a/pages/Mining.qml
+++ b/pages/Mining.qml
@@ -292,6 +292,7 @@ Rectangle {
                                 var success;
                                 if (persistentSettings.allow_p2pool_mining) {
                                     if (p2poolManager.isInstalled()) {
+                                        daemonManager.getArgs() //updates arguments
                                         if (persistentSettings.allowRemoteNodeMining) {
                                             startP2Pool()
                                         }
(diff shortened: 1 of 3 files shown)


selsta:
https://github.com/monero-project/monero-gui/issues/3911#issuecomment-1120238901 should work as a workaround.

devhyper:
Just tested on linux, will test on windows too.

devhyper:
I'm trying to get the PID of monerod on Windows using tasklist, however whenever I run it using QProcess no output is ever written.
```
QProcess p;
QStringList tempArgs;
tempArgs << "/FI";  
tempArgs << "\"IMAGENAME eq monerod.exe\"";
tempArgs << "/FO";
tempArgs << "LIST";
p.setProgram("tasklist");
p.setArguments(tempArgs);
p.start();
p.waitForFinished();
QString tasklist = p.readAllStandardOutput();
```
tasklist is still an empty QString after running.
@selsta any ideas?

devhyper:
Running wmic using QProcess didn't produce any output, so I tried a powershell cmdlet, and interestingly that worked.

devhyper:
Tested on windows and linux.

devhyper:
#3936

PR #3914, opened 2022-05-07
main: disable setting -platformpluginpath

selsta:
Not clear if this also applies to monero-gui but it doesn't hurt to disable this option.

- https://achow101.com/2021/02/0.18-uri-vuln
- https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-3401


plowsof (approved):
confirmed that launch flags are case sensitive so `--PLATformpluginpath` does not work anyway

PR #3922, opened 2022-05-14
docker: Upgrade to Ubuntu 20.04 + add support for static Linux arm64 builds

zer0conf:
- Dockerfile.linux to use Ubuntu 20.04 & gcc-9
- Adds static linux arm64 support


selsta (commented):
So the problem here is that we intentionally used Ubuntu 16.04 for glibc backwards compatibility.

While it will result in code duplication I'd create a separate docker file for arm64 linux.

zer0conf:
Makes sense, will rework the PR

PR #3926, opened 2022-05-20
Mining: add compile time option to disable p2pool

selsta:
P2Pool doesn't work with all package manager setups, so it's best to add an option to disable it during compile time.

@devhyper @BigmenPixel0 @kpcyrd

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ option(STATIC "Link libraries statically, requires static Qt")
 option(USE_DEVICE_TREZOR "Trezor support compilation" ON)
 option(WITH_SCANNER "Enable webcam QR scanner" OFF)
 option(WITH_DESKTOP_ENTRY "Ask to install desktop entry on first startup" ON)
+option(WITH_P2POOL "Enable mining with P2Pool" ON)
 option(DEV_MODE "Checkout latest monero master on build" OFF)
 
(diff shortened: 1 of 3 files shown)


plowsof:
This would be great! i've seen / helped many users who are confused as to why their 'package manager' installed Monero GUI displays the 'p2pool failed to install message. UX ++

BigmenPixel0:
I seem to be able to add p2pool to flatpak. Just not tested yet.

kpcyrd:
Is there something that can be done to fix the packages instead? :) I wasn't aware of any issues yet.

selsta:
The GUI downloads the p2pool binary in the same directory as the monero-wallet-gui binary. I can imagine that package managers don't like this, as it's not the proper way to install something and also might cause permission issues.

Putting p2pool directly during the installation process in the same folder as monero-wallet-gui might work. Then you wouldn't need to disable it during compile time. But this has to be tested, as there are also log and config files written.

selsta:
Basically two options:

1) Disable during compile time
2) Install p2pool together with monero-wallet-gui, assuming everything was tested and works correctly

kpcyrd:
p2pool is now available in Arch Linux: https://archlinux.org/packages/community/x86_64/p2pool/.

`pacman -S p2pool` installs a binary to `/usr/bin/p2pool`, the same directory that also contains `monero-wallet-gui`.

selsta:
Closing as this got stale and no more requests that this should be added.

PR #3933, opened 2022-05-28
Use `p2pool` folder for p2pool on Windows

SChernykh:
And set the correct file permissions on it so it can be run without administrator privileges.


SChernykh:
@devhyper @selsta please review. I also plan to release P2Pool v2.1 in the end of the month and make a separate PR for it. These 2 PRs will be needed to support the coming v18 Monero network upgrade.

selsta (approved):
Tested on macOS. A second review from @devhyper would be great.

devhyper (approved):
Tested on Linux and Windows

PR #3954, opened 2022-06-26
Transfer: don't create offline tx before wallet is synced

selsta:
ping @reemuru you have used offline transaction signing before, could you review this?


plowsof:
with this pr the 'create' button is not clickable when the wallet isn't syncd. (currently this is allowed and causes the GUI to timeout). selsta helped me compile the GUI to test this PR (as i was the one who raised the issue) and here is it in action:
Not synced: ( create offline tx button not clickable )
sync_on
 Synced: (create offline tx button clickable)
sync_off
 
 the only unrelated reason for the button being clickable when not synced is when the gui believes it is fully synced at start up, and then begins syncing later.

plowsof (approved):

reemuru (approved):
Nice patch, I didn't notice that last time I was using stagenet. LGTM.

PR #3955, opened 2022-06-26
oshelper: fix openContainingFolder on Linux

selsta:


plowsof (approved):
In linux : the file (not the dir) is opened (e.g. when you go to settings > info and click on the log dir) . This PR fixes the issue.

PR #3957, opened 2022-06-28
DaemonManagerDialog: close dialog when GUI is connected

selsta:
This can be triggered when having a restricted local node running.

### DIFF
--- a/components/DaemonManagerDialog.qml
+++ b/components/DaemonManagerDialog.qml
@@ -32,6 +32,7 @@ import QtQuick.Dialogs 1.2
 import QtQuick.Layouts 1.1
 import QtQuick.Controls.Styles 1.4
 import QtQuick.Window 2.0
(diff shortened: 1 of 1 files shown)


plowsof:
**behaviour now:**    
./monerod --restricted-rpc
[...make sure its running/synching...]
./monero-wallet-gui
[...gui displays dialogue ' starting daemon 10...9..8...]
UI says it has not started / 120s timeout screen displays, after clicking ok, 'reality' is shown (that infact the daemon is connected / and synching)

**with this patch:**     
the 'starting daemon 10..9..' dialogue still appears but is 'almost' immediately closed (visible for less than half a second) and everything is displaying correctly (connected / synchronising) 

This patch would save alot of headaches, an acceptable quick fix , while the 'root cause' remains to be fixed (to remove the box entirely from displaying)

plowsof (approved):

PR #3958, opened 2022-07-04
WizardRestoreWallet1: trim whitespaces from seed

selsta:
Replaces #3748


plowsof:
after trying out #3748 i suggested accounting for 'copy and pasting a seed with a newline / whitespace ' at the end of it (which prevents you from clicking next, even though the seed is valid) , this patch handles this situation correctly , and carries over the useful features of 3748 (enter / return disabled)

plowsof (approved):

PR #3959, opened 2022-07-04
Updated to p2pool v2.2

SChernykh:
p2pool v2.2 has full support for the coming Monero network upgrade


selsta (approved):
Hashes verified.

PR #3965, opened 2022-07-09
convert nettype to int to return correct rpc port

### LINKED ISSUE #3962: local stagenet node can not connect
[in my version of reality] when you start gui and open a stagenet wallet with a `./monerod --stagenet` already running, it is unable to connect. however if you select `Remote node` and input `localhost:38081` it connects fine.. clicking on `Local` again to double check and yes unable to connect.

looking around in the source, i decided to hardcode the default port of testnet and 'default' to 38081 https://github.com/monero-project/monero-gui/blob/b26f38d37ed81e84e51199425a633bc517f7bec2/main.qml#L2255
```C++
    function getDefaultDaemonRpcPort(networkType) {
        switch (networkType) {
            case NetworkType.STAGENET:
                console.log("[DEBUG]: STAGENET");
                return 38081;
            case NetworkType.TESTNET:
                console.log("[DEBUG]: TESTNET");
                return 38081;
            default:
                console.log("[DEBUG]: DEFAULT");
                return 38081; //<-- we always end up here for stagenet wallet.. nettype == 2 and persistant settings == 2 
                //perhaps NetworkType.STAGENET is not set correctly?
        }
    }
```
I was unable to see these console.log messages but through process of elimination , when opening a stagenet wallet the networktype is neither stagenet or testnet, it uses default (18081) so unable to connect.

the correct value should be read from here but apparently not:
https://github.com/monero-project/monero-gui/blob/b26f38d37ed81e84e51199425a633bc517f7bec2/main.qml#L94

this looked weird how mainnet is set but the others not?

https://github.com/monero-project/monero/blob/424e4de16b98506170db7b0d7d87a79ccf541744/src/wallet/api/wallet2_api.h#L45

now i can review #3734 with a local stagenet node :smile:

plowsof:
closes #3962 , title could use a little work, its for the local daemon running in the background as `./monerod --stagenet`
selsta suggested a better 1 liner instead of a new line: thanks   
```C++
switch (parseInt(networkType))
```
solution # 2 is to change `var` to `int` but the root cause of this problem remains unknown 
https://github.com/monero-project/monero-gui/blob/b26f38d37ed81e84e51199425a633bc517f7bec2/main.qml#L1378

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -2253,7 +2253,7 @@ ApplicationWindow {
     }
 
     function getDefaultDaemonRpcPort(networkType) {
-        switch (networkType) {
+        switch (parseInt(networkType)) {
             case NetworkType.STAGENET:
                 return 38081;
             case NetworkType.TESTNET:


reemuru:
Ah yeah I always just closed and reopened stagenet wallet but it is an annoying bug. I think it has something to do with with the enum `NetworkType` definition of `uint8_t` in `wallet2_api.h`. Combine that with JS + qml and all kind of weirdness can happen. Nice patch.

Tested ack on [8199ef0](https://github.com/monero-project/monero-gui/pull/3965/commits/8199ef004d7597e3106dcc7bb856ed75b5e996e2)

reemuru (approved):

PR #3978, opened 2022-07-21
TxUtils: don't show resolve for float

### LINKED ISSUE #3975: Amount gets interpreted as openalias domain
Steps to reproduce

1. Paste `0.270867069989` into address field
2. Click resolve
3. Crash

reemuru:
Closes #3975

Add parseFloat() checks so that the 'Resolve' button
for OpenAlias remains hidden. Also add a length check
for one character with a dot. IP addresses will not be
respected, but the whole point of domain names and OA is
for facilitating human readable / user friendly information.

### DIFF
--- a/js/TxUtils.js
+++ b/js/TxUtils.js
@@ -71,9 +71,8 @@ function checkSignature(signature) {
 
 function isValidOpenAliasAddress(address) {
     address = address.trim()
-    var dot = address.indexOf('.')
-    if (dot < 0)
-        return false
     // we can get an awful lot of valid domains, including non ASCII chars... accept anything
-    return true
+    // there should be something after the .
+    // make sure it is not some kind of floating number
+    return address.length > 2 && isNaN(parseFloat(address)) && address.indexOf('.') >= 0
 }


js/TxUtils.js
```diff
@@ -71,9 +71,8 @@ function checkSignature(signature) {
...
 function isValidOpenAliasAddress(address) {
     address = address.trim()
-    var dot = address.indexOf('.')
-    if (dot < 0)
-        return false
     // we can get an awful lot of valid domains, including non ASCII chars... accept anything
-    return true
+    // there should be something after the .
+    // make sure it is not some kind of floating number
+    return address.length > 2 && !parseFloat(address) && parseFloat(address) !== 0 && address.indexOf('.') >= 0
```
selsta, line 77: Sorry, why parseFloat twice? I feel like I'm missing something obvious :P

reemuru: Hmm, iirc there was some interesting edge case here where normally the string would parse to 'NaN' which is falsy. Then something like `parseFloat("0...")` parses to 0 which should be falsy but falls through that first check? Idk let me double check I didn't really test this thoroughly.

selsta:
```suggestion
    return address.length > 2 && isNaN(parseFloat(address)) && address.indexOf('.') >= 0
```

Would this work too?

reemuru: Oh yeah, this looks much cleaner. Let me force push in a sec.

selsta (approved):

tobtoht (approved):

PR #3973, opened 2022-07-17
README: Update copyright to 2022

garth-xmr:

PR #3993, opened 2022-08-06
fix and update copyright notice

coquizen:
1. No need to use both Copyright and ©
2. No punctuation after the copyright section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Monero GUI
 
-Copyright (c) 2014-2019, The Monero Project
+© 2014-2022 The Monero Project
 
 ## Table of Contents
   * [Development resources](#development-resources)


selsta:
Please see https://en.wikipedia.org/wiki/BSD_licenses

`Copyright (c)` seems to be in line with how the BSD license is written, and how we do it in other repositories. Removing the comma is fine but we will merge #3973 first.

selsta:
Please rebase.

coquizen:
I wasn't aware there was already a written policy for BSD style licenses. The reference I was basing the commit on came from US Copyright Office:

> Copyright notice is a statement placed on copies or phonorecords of a work to inform the public that a copyright owner
> is claiming ownership of it.2
>  A notice consists of three elements that generally appear as a single continuous statement:
> • The copyright symbol © (or for phonorecords, the
> symbol ℗ ); the word “copyright”; **or** [emphasis added] the abbreviation
> “copr.”;
> • The year of first publication of the work; and
> • The name of the copyright owner.
> Example: © 2017 John Doe
(Source: [Circular 3 Copyright Notice](https://www.copyright.gov/circs/circ03.pdf)

But the suggestion is minor. This isn't the hill I am choosing to die on. :)

selsta:
I will merge this if you rebase and undo the © change.

selsta:
Hmm, seems we updated the year already.

PR #3994, opened 2022-08-06
libwalletqt: refresh once on wallet opening

selsta:
Found and fixed an issue that the wallet first waits 10 seconds before starting a refresh, turns out Feather fixed it 2 years ago. Added tobtoht as co-author for this reason and copied their variable naming.


plowsof:
wow, huge noticeable difference.. from 10s to 2~

plowsof (approved):

tobtoht (approved):

PR #4026, opened 2022-09-09
installer: allow to install on arm64

selsta:
Since Windows 11 ARM PCs can run x64 applications with emulation.

@rbrunner7


rbrunner7 (approved):
Checked the `ArchitecturesAllowed` syntax with the help file, looks good.

PR #4036, opened 2022-09-22
Fix behavior when switching nodes

j-berman:
Works together with the fixes in Monero PR https://github.com/monero-project/monero/pull/8585.

The most significant fix in this PR is that it prevents the GUI from getting stuck in an infinite "Connecting" loop when connecting to a non-existent (or sometimes offline) daemon.

It also fixes some jank (and *some* of the incorrect underlying behavior) when switching nodes. The underlying async behavior still isn't perfect, but it's a move in a better direction. See the demos below.

## GUI v0.18.1.1 + this PR + Monero PR 8585

*Not shown: connecting to `1.2.3.4:18081` takes 20 seconds to time out and display "Disconnected"*

PR behavior

## GUI v0.18.1.1 + Monero PR 8585 (this PR is excluded)

*Not shown: connecting to `1.2.3.4:18081` gets stuck infinite "Connecting"*

Only Monero repo PR behavior

### DIFF
--- a/src/libwalletqt/Wallet.cpp
+++ b/src/libwalletqt/Wallet.cpp
@@ -101,20 +101,30 @@ NetworkType::Type Wallet::nettype() const
 void Wallet::updateConnectionStatusAsync()
 {
     m_scheduler.run([this] {
+        qDebug() << "updateConnectionStatusAsync current status:" << m_connectionStatus;
         if (m_connectionStatus == Wallet::ConnectionStatus_Disconnected)
         {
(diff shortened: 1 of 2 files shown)


hinto-janaiyo (approved):
Tested this with https://github.com/monero-project/monero/pull/8585 on old nodes, offline nodes, bad ips

Error message + timeouts work correctly

selsta (approved):

01SMD (commented):
.

01SMD (commented):
.

PR #4047, opened 2022-10-06
DaemonManager: take dataDir into account when sending cmd

### LINKED ISSUE #4046: sendCommand does not take dataDir into account -> log file gets created
If someone changes their blockchain location they don't want any bitmonero.log file being created in the default location.

selsta:
Closes #4046


plowsof:
using this PR - no bitmonero.log file is created in the default dir when stopping/starting/re-opening the gui

plowsof (approved):

PR #4051, opened 2022-10-20
SettingsWallet: Update scan transaction for changes to behavior in monero repo PR 8566

j-berman:
EDIT: also ensures GUI simple mode and bootstrap mode users default to "untrusted" daemon

1. Update balance after calling `scanTransaction` in order to reflect correct balance along with https://github.com/monero-project/monero/pull/8566
2. Display clear error when user is connected to an untrusted daemon and requests to scan a tx that is older than their wallet's most recent tx.

### DIFF
--- a/pages/settings/SettingsWallet.qml
+++ b/pages/settings/SettingsWallet.qml
@@ -144,7 +144,7 @@ Rectangle {
                         console.error("Error: ", currentWallet.errorString);
                         if (currentWallet.errorString == "The wallet has already seen 1 or more recent transactions than the scanned tx") {
                             informationPopup.title = qsTr("Error") + translationManager.emptyString;
-                            informationPopup.text = currentWallet.errorString + ".\n\nIn order to rescan the transaction, you can re-sync your wallet by resetting the wallet restore height in the Settings > Info page. Make sure to use a restore height from before your wallet's earliest transaction." + translationManager.emptyString;
+                            informationPopup.text = qsTr("The wallet has already seen 1 or more recent transactions than the scanned transaction.\n\nIn order to rescan the transaction, you can re-sync your wallet by resetting the wallet restore height in the Settings > Info page. Make sure to use a restore height from before your wallet's earliest transaction.") + translationManager.emptyString;
                             informationPopup.icon = StandardIcon.Critical
                             informationPopup.onCloseCallback = null
                             informationPopup.open();
--- a/pages/settings/SettingsWallet.qml
+++ b/pages/settings/SettingsWallet.qml
@@ -138,6 +138,7 @@ Rectangle {
                 inputDialog.onAcceptedCallback = function() {
                     var txid = inputDialog.inputText.trim();
                     if (currentWallet.scanTransactions([txid])) {
+                        updateBalance();
                         appWindow.showStatusMessage(qsTr("Transaction successfully scanned"), 3);
                     } else {
                         appWindow.showStatusMessage(qsTr("Failed to scan transaction") + ": " + currentWallet.errorString, 5);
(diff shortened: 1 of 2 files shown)


selsta (approved):
Patch looks ok, why is this necessary now?

j-berman:
Assume a user receives tx0 then tx1. Assume the wallet has already scanned both txs.

The user then calls `scan_tx(tx0)` to re-scan tx0.

First the wallet will remove tx1 and tx0, then re-process tx0 then tx1 in order.

In https://github.com/monero-project/monero/pull/8566, when the wallet re-processes tx2, I chose to ignore the `on_money_received` callback to maintain a clean UX in the CLI. The callback only fires for the tx provided to `scan_tx`. So in this case it only fires for tx1.

The GUI updates the displayed balance inside this callback:

https://github.com/monero-project/monero-gui/blob/62878a042cc3cadb0650990270c5f7ccdb591ff3/main.qml#L770-L779

So without this PR here, and with the new PR in the Monero repo, the balance the GUI displays after calling `scan_tx(tx1)` is the balance after the wallet has processed just tx0.

This PR makes sure to update the balance after `scan_tx` (i.e. after all txs are processed) to reflect the correct balance.

Alternatively I could have re-fired the callback for all txs but that's weird in the CLI, since it will indicate feedback to the user that they received both tx0 and tx1 which is confusing to the user because the user only requested to scan tx0.

This PR wasn't necessary before my Monero repo PR because the balance wouldn't change at all when re-scanning a tx before.

j-berman:
Updated to catch the error when the wallet is connected to an untrusted daemon and the user attempts to scan a tx that is not their most recent tx. The wallet displays

GUI error untrusted daemon
:

pages/settings/SettingsWallet.qml
```diff
@@ -141,7 +141,16 @@ Rectangle {
                         updateBalance();
                         appWindow.showStatusMessage(qsTr("Transaction successfully scanned"), 3);
                     } else {
-                        appWindow.showStatusMessage(qsTr("Failed to scan transaction") + ": " + currentWallet.errorString, 5);
+                        console.error("Error: ", currentWallet.errorString);
+                        if (currentWallet.errorString == "The wallet has already seen 1 or more recent transactions than the scanned tx") {
+                            informationPopup.title = qsTr("Error") + translationManager.emptyString;
+                            informationPopup.text = currentWallet.errorString + ".\n\nIn order to rescan the transaction, you can re-sync your wallet by resetting the wallet restore height in the Settings > Info page. Make sure to use a restore height from before your wallet's earliest transaction." + translationManager.emptyString;
```
selsta, line 147: qsTr() missing

j-berman:
Spotted by @plowsof: the GUI unexpectedly assumes simple mode daemons are trusted. Updated to ensure they do not.

I don't see any issues with this today -- features that would be harmful when connected to an untrusted daemon are already not available in simple mode (rescanning spents, key image sync). However, without this change, and with the change to scan transaction in the monero repo, simple mode users would be left at added risk using the scan transaction feature, which is supposed to error out as in the screenshot above and described [here](https://github.com/monero-project/monero/pull/8566#issuecomment-1294255769).

plowsof:
@selsta advised me on what specific things should be tested. The new changes/fixes work perfectly now, 
- Simple mode users can scan their latest transaction / warning is displayed when scanning older transaction.
- Advanced mode users can scan an older transaction with a trusted node / see warning if untrusted.

Thank you! [my test script](https://gist.github.com/plowsof/c129e744e304d4e9879410d7d17238ba)

j-berman:
@selsta raised a valid point to me that bootstrap mode wallets don't need to be considered trusted when the user's daemon is fully synced. I don't see a particularly clean way to guarantee this behavior and figure it's safer to default assume the daemon is untrusted for bootstrap mode wallets

selsta (approved):

PR #4056, opened 2022-10-29
Bulgarian translation

TeoVibe:
fixed some syntax errors with "unfinished" left on completed translations and finished the translation


selsta:
Thank you, please submit the translations to https://translate.getmonero.org

We don't manually edit the .ts files, it has to be done over the platform.

PR #4062, opened 2022-11-02
workflows: verify p2pool hashes

plowsof:
selsta spotted "path" instead of "paths" in the workflow file - i missed this so it would run on every pull/push, and also offered suggestions on extra features (e.g. verify the pgp key fingerprint)

### DIFF
--- a/.github/verify_p2pool.py
+++ b/.github/verify_p2pool.py
@@ -1,0 +1,103 @@
+import requests
+import subprocess
+from urllib.request import urlretrieve
+import difflib
+
+sech_key = "https://p2pool.io/SChernykh.asc"
+sech_key_backup = "https://raw.githubusercontent.com/monero-project/gitian.sigs/master/gitian-pubkeys/SChernykh.asc"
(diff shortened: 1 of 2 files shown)


selsta (approved):

.github/workflows/verify_p2pool.yml
```diff
@@ -0,0 +1,15 @@
+on:
```
selsta, line 1: can you add something like

```
name: ci/gh-actions/verify
```

and rename the job to just

```
  p2pool-hashes:
```

selsta (approved):

.github/verify_p2pool.py
```diff
@@ -0,0 +1,98 @@
+import requests
+import subprocess
+from urllib.request import urlretrieve
+import difflib
+
+sech_key = "https://p2pool.io/SChernykh.asc"
```
SChernykh, line 6: You can add https://raw.githubusercontent.com/monero-project/gitian.sigs/master/gitian-pubkeys/SChernykh.asc as a back-up URL

selsta (approved):

SChernykh (approved):

PR #4071, opened 2022-11-20
Fixed ru translation

USBashka:
Replaced Е with Ё where needed


selsta:
Thank you, please submit them to https://translate.getmonero.org/ - we don't directly translate the source files.

PR #4085, opened 2022-12-18
main: skip proxy when setting localhost remote node

### LINKED ISSUE #4027: Allow local monerod server when using Socks5 proxy
Currently there is no way to connect to a local monerod instance and have all other connections go over the Socks5 proxy. When the Socks5 proxy is set, you lose connection to your local node. Many users will want to make sure price discovery is happening over Tor, and that is not currently possible when connecting to your own local monerod instance.

image

Ideally we can add a check box that allows it to bypass the proxy for local connections.

image

I wouldn't mind looking into adding it, as long as that is something that would be accepted.

selsta:
Closes #4027

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1447,8 +1447,15 @@ ApplicationWindow {
         function getWalletProxyAddress() {
             if (!useRemoteNode) {
                 return "";
+            } else {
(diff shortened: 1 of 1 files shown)


plowsof (approved):
Tested and works. Nice

PR #4088, opened 2022-12-23
open-wallet: set network type when using file browser

### LINKED ISSUE #4069: [UX] If User tries to open a stagenet wallet, let them
Suppose I open up the GUI and now select a wallet from the file. I know the wallet is a stagenet file, but when I try unlocking it, Monero hits me with the you can't open this in the mainnet wallet. 

Tell the user to switch it from the previous page or add an option to make it easier to switch the network from the unlock page itself.

plowsof:
closes #4069
 at open wallet file menu, if you click "Browse Filesystem" and open a .keys file - you will see an error if the networktype differs to the current one. Opening via the wallet icons adjust the network type automatically before opening.

to confirm: 
- all wallet icons are assigned their correct network type / open correctly as normal @ open wallet file menu.
- any wallet keys file opened via "Browse Filesystem" does not complain about a different network type.

(var networktype breaks things if i try "int" - although parseInt is still not required)

selsta is helping / offering advice for v1..2..- this is v5.0 >_<

### DIFF
--- a/src/main/oshelper.cpp
+++ b/src/main/oshelper.cpp
@@ -63,6 +63,7 @@
 
 #include "QR-Code-scanner/Decoder.h"
 #include "qt/ScopeGuard.h"
+#include "NetworkType.h"
 
 namespace
 {
@@ -279,4 +280,36 @@ bool OSHelper::installed() const
 #else
     return false;
 #endif
+}
+
+std::pair<quint8, QString> OSHelper::getNetworkTypeAndAddressFromFile(const QString &wallet)
+{
+    quint8 networkType = NetworkType::MAINNET;
+    QString address = QString("");
+    // attempt to retreive wallet address
+    if(QFile::exists(wallet + ".address.txt")){
+        QFile file(wallet + ".address.txt");
+        file.open(QFile::ReadOnly | QFile::Text);
+        QString _address = QString(file.readAll());
+        if(!_address.isEmpty()){
+            address = _address;
+            if(address.startsWith("5") || address.startsWith("7")){
+                networkType = NetworkType::STAGENET;
+            } else if(address.startsWith("9") || address.startsWith("B")){
+                networkType = NetworkType::TESTNET;
+            }
+        }
+
+        file.close();
+    }
+    return std::make_pair(networkType, address);
+}
+
+quint8 OSHelper::getNetworkTypeFromFile(const QString &keysPath) const
+{
+    QString walletPath = keysPath;
+    if(keysPath.endsWith(".keys")){
+        walletPath = keysPath.mid(0,keysPath.length()-5);
+    }
+    return getNetworkTypeAndAddressFromFile(walletPath).first;
 }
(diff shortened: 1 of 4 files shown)


rbrunner7 (approved):
Looks good to me.

plowsof:
Thank you for checking @rbrunner7 ! After showing my hand / proof of suffering/learning - I have been granted the 'answer' from selsta. The new changes move the functions inside `oshelper` (which has the advantage of already being accessible inside `WizardController.qml` (no need to import `WalletKeysFilesModel` there or add a strange declare{id: } thing) and is more appropriately named 'oshelper' as we're 'opening and reading a text file'.  Also dropping the need to declare the `var networktype` to save on diffs. (I have tested the new changes myself but the same functionality/method to confirm still stand)

rbrunner7 (approved):
Still looks good :)

I am sure the code has won from @selsta 's advice where to put what. (I myself couldn't properly judge where the best places are.)

src/main/oshelper.cpp
```diff
@@ -280,3 +281,33 @@ bool OSHelper::installed() const
...
+        }
+
+        file.close();
+    }
+    return std::make_pair(networkType, address);
+}
+
+quint8 OSHelper::getNetworkTypeFromFile(const QString &keysPath) const
+{
+    QString walletPath = keysPath.mid(0,keysPath.length()-5);
```
selsta, line 310: Maybe add a sanity check that the file passed does have a .keys extension before removing it.

src/main/oshelper.cpp
```diff
@@ -280,3 +281,33 @@ bool OSHelper::installed() const
...
+    std::pair<quint8, QString> networkTypeAndAddress = getNetworkTypeAndAddressFromFile(walletPath);
+    return networkTypeAndAddress.first;
```
selsta, line 312:
```suggestion
    return getNetworkTypeAndAddressFromFile(walletPath).first;
```

No need for an extra variable.

src/qt/KeysFiles.cpp
```diff
@@ -121,26 +122,9 @@ void WalletKeysFilesModel::findWallets(const QString &moneroAccountsDir)
...
-                if(address.startsWith("5") || address.startsWith("7")){
-                    networkType = NetworkType::STAGENET;
-                } else if(address.startsWith("9") || address.startsWith("B")){
-                    networkType = NetworkType::TESTNET;
-                }
-            }
-
-            file.close();
-        }
+        std::pair<quint8, QString> networkTypeAndAddress = OSHelper::getNetworkTypeAndAddressFromFile(wallet);
```
selsta, line 125:
```suggestion
        auto networkTypeAndAddress = OSHelper::getNetworkTypeAndAddressFromFile(wallet);
```

Easier than writing the whole type but personal preference.

selsta (commented):
Small suggestions, but basically approved.

selsta (approved):

PR #4113, opened 2023-02-01
Utils: simplify ago func

ufUNnxagpM:
Fix issue where the minutes display is shown when delta is 3600 (one hour) and the hour display is shown when delta is 86400 (one day).

### DIFF
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -57,12 +57,12 @@ function ago(epoch) {
 
     if(delta < 60)
         return qsTr("%n second(s) ago", "0", Math.floor(delta))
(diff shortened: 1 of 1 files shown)


plowsof:
i've made a fiddle using this PR to test here https://jsfiddle.net/h7r8y5ku/ , can change the input value on line 46 and run
```
Input: 86400

Output now: 24 hour(s) ago

Output #4113: 1 day(s) ago
```

```
Input: 3600

Output now: 60 minute(s) ago

Output #4113: 1 hour(s) ago
```

i have not tested any other values yet but on these 2 specifically it clearly works.

plowsof (approved):

PR #4123, opened 2023-02-21
workflows: update dependencies to fix warnings

selsta:


.github/workflows/build.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   build-macos:
     runs-on: macOS-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
```
hinto-janai, line 9: There's `actions/checkout@v4` now.

hinto-janai (approved):
There's an actions/checkout@v4 at this point btw.

Issue #4122, opened 2023-02-20
monerod runs as dbus.service 


plowsof:
the goal is to detect if monerod is 'running as a service'

i hope its an issue on my end because sanity checks earlier did not show this behaviour with release. edit* i did not 'double click the binary' 
to check this: (when monerod is running)
`ps -eo pid,cgroup | grep $(pgrep monerod)`

running as ./monerod
8742 0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-67ffb31d-2952-4656-94a4-1dc3a46b

after monerod gets restarted by the gui or started by gui normally
9017 0::/user.slice/user-1000.slice/user@1000.service/app.slice/dbus.service

actual service
10492 0::/system.slice/monerod.service

the gui checks if it ends in .service 

will look at this tomorrow but the fix to ignore "dbus.service" will work

issue happens when i start monero-gui by clicking the appimage / or binary - instead of `./monero-wallet-gui` 

### pstree ?
`pstree -s $(pgrep monerod)`
./monerod ran from terminal
systemd---systemd---gnome-terminal----bash---monerod---34*[{monerod}]
monerod actual systemd service
systemd---monerod---25*[{monerod}]
monerod from gui 
systemd───systemd───monerod───28*[{monerod}]

if there is only one occurrence of systemd then return true?

PR #4124, opened 2023-02-21
DaemonManager: remove systemd check

plowsof:
#4122 for me, 'double clicking' the monero gui binary leads to monerod being launched as a dbus.service. running it from the terminal. i did not notice this on Mint but i would sleep easier if this check was disabled for now until a better method of detecting if monerod is running under systemd or not.


selsta (approved):
To be reverted once a better method is found.

tobtoht (approved):

PR #4140, opened 2023-03-27
readme: tweak build instructions to include ARM64 target

Botspot:
ARM64 Linux is growing rapidly in popularity, both due to advances in SBC design and Asahi Linux for M1 Macs. This PR makes the process of compiling monero-gui slightly easier for beginners and seasoned users alike.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,11 @@ The following instructions will fetch Qt from your distribution's repositories i
     ```
     make release -j4
     ```
+    If on ARM64:
+
+    ```
+    make release-linux-armv8 -j4
+    ```
 
     If on ppc64le:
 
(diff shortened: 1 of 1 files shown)


selsta:
Did you try building with just

```
make release
```

It should work too.

Botspot:
> Did you try building with just
> 
> ```
> make release
> ```
> 
> It should work too.

I just now tried that, and yes it does indeed work. I had simply taken the readme at face value that `make release` would only work on x86_64.
This is good to know, but I still think the readme should be adjusted. Please see my updated proposal.

selsta:
I have opened #4145 and set you as the author, `make release` should work everywhere so I simplified the README instead.

PR #4151, opened 2023-04-16
flatpak: Add workflow

BigmenPixel0:
A new release will trigger the workflow. It builds the flatpak package in 3 parts (because of slow QEMU), prints hashes and pushes to Flathub via the secret token.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -1,0 +1,170 @@
+name: Flatpak
+
+on:
+  push
+
+jobs:
+  part1:
+    name: Part 1/3
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:kde-5.15-22.08
+      options: --privileged
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+
+    - name: Install deps
+      run: dnf -y install docker
+
+    - name: Setup QEMU
+      uses: docker/setup-qemu-action@v2
+      with:
+        platforms: arm64
+
+    - name: Build flatpak
+      uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+      env:
+        FLATPAK_BUILDER_N_JOBS: 3
+      with:
+        manifest-path: share/org.getmonero.Monero.yaml
+        arch: ${{ matrix.arch }}
+        cache: false
+        stop-at-module: boost
+
+    - name: Tar flatpak-builder
+      run: tar -cvf flatpak-builder.tar .flatpak-builder
+
+    - name: Save flatpak-builder
+      uses: actions/upload-artifact@v3
+      with:
+        name: flatpak-builder-${{ matrix.arch }}
+        path: flatpak-builder.tar
+
+  part2:
+    name: Part 2/3
+    needs: part1
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:kde-5.15-22.08
+      options: --privileged
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install deps
+      run: dnf -y install docker
+
+    - name: Setup QEMU
+      uses: docker/setup-qemu-action@v2
+      with:
+        platforms: arm64
+
+    - name: Restore flatpak-builder
+      uses: actions/download-artifact@v3
+      with:
+        name: flatpak-builder-${{ matrix.arch }}
+
+    - name: Untar flatpak-builder
+      run: tar -xvf flatpak-builder.tar
+
+    - name: Build flatpak
+      uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+      env:
+        FLATPAK_BUILDER_N_JOBS: 3
+      with:
+        manifest-path: share/org.getmonero.Monero.yaml
+        arch: ${{ matrix.arch }}
+        cache: false
+        stop-at-module: monero-gui
+
+    - name: Tar flatpak-builder
+      run: tar -cvf flatpak-builder.tar .flatpak-builder
+
+    - name: Save flatpak-builder
+      uses: actions/upload-artifact@v3
+      with:
+        name: flatpak-builder-${{ matrix.arch }}
+        path: flatpak-builder.tar
+
+  part3:
+    name: Part 3/3
+    needs: [part1, part2]
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:kde-5.15-22.08
+      options: --privileged
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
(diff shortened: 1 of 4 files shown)


selsta:
Have you thought about using upload-artifact / download-artifact instead of the cache to pass data between the jobs?

https://github.com/actions/upload-artifact
https://github.com/actions/download-artifact

At the end you can also use

https://github.com/marketplace/actions/delete-artifact

to delete the temporary data. My worry with using a cache is that other jobs might delete data in between the steps due to the 10GB limit.

https://github.com/monero-project/monero-gui/actions/caches

BigmenPixel0:
> Have you thought about using `upload-artifact` / `download-artifact` instead of the cache to pass data between the jobs?

Oh, that would be much better!

selsta:
What if you .tar the dir manually before uploading the .tar?

selsta:
```
tar -cvf flatpak-builder.tar .flatpak-builder

...

    - name: Save flatpak-builder
      uses: actions/upload-artifact@v3
      with:
        name: flatpak-builder-${{ matrix.arch }}
        path: flatpak-builder.tar
```

```
    - name: Restore flatpak-builder
      uses: actions/download-artifact@v3
      with:
        name: flatpak-builder-${{ matrix.arch }}
        path: flatpak-builder.tar

...

tar -xvf flatpak-builder.tar
```

BigmenPixel0:
x86_64 flatpak-builder.tar [was uploaded](https://github.com/BigmenPixel0/monero-gui/actions/runs/4723069372/jobs/8378511632) succesfully, so now I'm waiting for X hours. :)

share/org.getmonero.Monero.metainfo.xml
```diff
@@ -0,0 +1,74 @@
...
+    <url type="bugtracker">https://github.com/monero-project/monero-gui/issues</url>
+    <url type="donation">https://ccs.getmonero.org/donate/index.html</url>
+    <url type="faq">https://getmonero.org/get-started/faq</url>
+    <url type="help">https://getmonero.org/resources/user-guides</url>
+    <url type="translate">https://translate.getmonero.org</url>
+    <url type="contact">https://getmonero.org/community/hangouts</url>
+    <url type="contribute">https://getmonero.org/get-started/contributing</url>
+
+    <releases>
+        <release version="0.18.2.2" date="2023-04-10"/>
```
selsta, line 59: Does this have to be changed in advanced? Sometimes we don't know the exact release date.

BigmenPixel0: > Does this have to be changed in advanced?

Yes. 

> Sometimes we don't know the exact release date.

Even, for example, the day before the release or the day? I could create a PR that will be accepted before release, or the problem still is not solved so?

BigmenPixel0: In [OBS](https://github.com/obsproject/obs-studio/blob/72759ae69da9d211c849b3d16cc549d9ab8e3c1b/UI/xdg-data/com.obsproject.Studio.appdata.xml.in#L44) it's done interesting.

selsta: So this can be changed after the release tag is set?

BigmenPixel0: > So this can be changed after the release tag is set? Or can it be automatically set by a script before the workflow is running?

It will be need to add some [CMake files](https://github.com/obsproject/obs-studio/commit/6b9da9b9127f066e782d0fe8be5cc703da570d1f), so it won't be need to change the appdata file every release.

> Also seems OBS doesn't have any historical releases. Are they necessary?

Desirable, but not required.

BigmenPixel0:
For now there are 2 problems: metainfo updating and CI checks before flatpak building (or wait for a new release of flat-manager or make it manually).

BigmenPixel0:
> > CI checks before flatpak building (or wait for a new release of flat-manager or make it manually).
> 
> Can you expand on this?

That's [appdata](https://github.com/obsproject/obs-studio/blob/a0fbdc3c3469c2069f2a06c169180b4539f30fe8/.github/workflows/flatpak.yml#L92) and [manifest](https://github.com/flathub/flatpak-builder-lint) checkers.

BigmenPixel0:
It will push to the beta branch because we should test the beta token.

BigmenPixel0:
I changed the token `FLATHUB_TOKEN_BETA` to `FLATHUB_` because exactly `FLATHUB_` has been added to the secrets. :)

.github/workflows/flatpak.yml
```diff
@@ -0,0 +1,171 @@
+name: Flatpak
+
+on:
+  release:
+    types: push
```
selsta, line 5: Is this documented somewhere? I don't see `push` here, but maybe I'm looking at the wrong place or missing something.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

.github/workflows/flatpak.yml
```diff
@@ -0,0 +1,171 @@
+name: Flatpak
+
+on:
+  release:
```
johnnyjeannatasha, line 4: Lines 1-7 addhh

.github/workflows/flatpak.yml
```diff
@@ -0,0 +1,171 @@
...
+      run: tar -xvf flatpak-builder.tar
+
+    - name: Build flatpak
+      uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+      env:
+        FLATPAK_BUILDER_N_JOBS: 3
+      with:
+        manifest-path: share/org.getmonero.Monero.yaml
+        arch: ${{ matrix.arch }}
+        cache: false
```
johnnyjeannatasha, line 87: Change to true cache? Needing cashe now

BigmenPixel0: Already via artifacts.

johnnyjeannatasha: Ok. Thanks.

johnnyjeannatasha: Let

adrelanos:
> flathub has a verified blue checkmark similar to Twitter. It shows that a flathub project is official and not by random third-party. Please consider applying for it.

* https://github.com/monero-project/monero-gui/issues/4206

.github/workflows/flatpak.yml
BigmenPixel0: Now resolved?

BigmenPixel0: OK, so is it better to push before other prs now or at the final (to not create a lot of workflows)?

selsta: Probably best to merge this last.

selsta (approved):
I think this can be tested now.

johnnyjeannatasha:
Adding link to open site.

Get Outlook for Android<https://aka.ms/AAb9ysg> donate to that..thank you, byebye....ANATASHA B. DELGADO~FIERRO,  "have a ™great afanstacical¢ %nice day% •to you✓"
________________________________

johnnyjeannatasha:
Resisting API key token

Get Outlook for Android<https://aka.ms/AAb9ysg> donate to that..thank you, byebye....ANATASHA B. DELGADO~FIERRO,  "have a ™great afanstacical¢ %nice day% •to you✓"
________________________________

PR #4157, opened 2023-04-19
DaemonManager.cpp: disable JIT on macOS ARM

selsta:


plowsof:
removed the "if" to force the changes. compiles fine. monerod is able to function as normal and sync blocks. mining starts only in the UI while status shows 0h/s

plowsof (approved):

PR #4158, opened 2023-04-21
TxUtils: add handleOpenAliasResolution func, simplify code

selsta:
Simplifies code and removes code duplication.

### DIFF
--- a/js/TxUtils.js
+++ b/js/TxUtils.js
@@ -76,3 +76,34 @@ function isValidOpenAliasAddress(address) {
     // make sure it is not some kind of floating number
     return address.length > 2 && isNaN(parseFloat(address)) && address.indexOf('.') >= 0
 }
+
+function handleOpenAliasResolution(address, descriptionText) {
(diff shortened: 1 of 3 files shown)


plowsof:
sanity checks by changing line 81: (expected error = pass)
```js
    const result = "false|888tNkZrPN6JsEgekjMnABU4TBzc2Dt29EPAvkRxbANsAnjyPbb3iQ1YBRk1UXcdRsiKc9dhwMVgN5S9cQUiyoogDavup3H" 
```
:heavy_check_mark: 
```js
    const result = "false|n0tr34l"
```
:heavy_check_mark:  
```js
    const result = "true|n0tr34l"
```
:heavy_check_mark: 
```js
    const result = "|n0tr34l"
```
:heavy_check_mark: 

turning wifi off before clicking resolve = no crash, it seems the result == "|" so i see no address found / dnsec not valid :heavy_check_mark:

plowsof (approved):

PR #4162, opened 2023-04-29
Add VSCode IDE Settings

elibroftw:
This will improve productivity for Windows users using VS Code as they just need to run a task or a command to build and run the wallet.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,40 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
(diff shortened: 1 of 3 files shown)


plowsof:
Is this out of scope? (forgive me/ignore if other IDE's are supported) but it seems like a hassle (what about my ide? my ide is older and needs other settings etc) perhaps this is best suited for some kind of tutorial

elibroftw:
I don't think this is out of scope. The project should support all IDEs of its contributors. It doesn't impede anyone else and it improves usability and onboarding.

RecRanger:
+1 for adding features like this. It's very helpful when new devs are looking to get started, and don't want to fight the basic project setup. 

The common argument against this sort of PR is that the choices are highly-controversial. Devs can easily override these settings with user/profile settings if that is the case. In reality though, there's really nothing non-standard in here.

PR #4176, opened 2023-05-16
Translations from Weblate

Monero-Weblate:


selsta (commented):
Why are the `pages/Mining.qml` related translations being removed?

selsta (approved):

PR #4192, opened 2023-07-02
fix overlap in certain languages

SNeedlewoods:
Overlapping elements fixed for every translation that is currently available.
Future changes of the translations (if translated string is longer than source string) may break this.


jpk68 (approved):
Tested and working.

selsta (approved):

PR #4244, opened 2023-11-14
DaemonManager: prevent GUI from hanging on hanging `monerod` launch

### LINKED ISSUE #4240: GUI hangs if `monerod` cannot bind
When starting `monero-wallet-gui` and starting a local node, it will hang with no indication/error/log if `monerod` itself cannot bind, e.g:
```
2023-11-07 17:44:20.109	E ZMQ bind failed: Address already in use
2023-11-07 17:44:20.121	E Exception in main! Failed to add TCP socket(127.0.0.1:18082) to ZMQ RPC Server
```

## Reproduce
1. Start something on ZMQ/RPC port
2. Start `monero-wallet-gui` and launch local node
3. `monero-wallet-gui` hangs forever attempting to launch `monerod` which cannot bind

image

Something could be logged if `monerod` exits with non-zero.

hinto-janai:
Fixes https://github.com/monero-project/monero-gui/issues/4240

## Problem 1
`monerod` hangs on this line when creating an http connection (for local `json-rpc`):
https://github.com/monero-project/monero/blob/ac02af92867590ca80b2779a7bbeafa99ff94dcb/src/common/rpc_client.h#L125

using the generic 3 minutes and 30 seconds connection timeout:
https://github.com/monero-project/monero/blob/ac02af92867590ca80b2779a7bbeafa99ff94dcb/src/common/http_connection.h#L42-L45

This causes direct `json-rpc` invocation (`./monerod sync_info`) to hang for 3m30s before exiting if `monerod` can't bind.

## Problem 2
Monero GUI relies on direct `json-rpc` invocation to tell if `monerod` is "running":
https://github.com/monero-project/monero-gui/blob/e9cd4588aef3f0808ce153957a504f43dcdbeb26/src/daemon/DaemonManager.cpp#L239-L245

This causes the GUI to hang and the [`Watcher`](https://github.com/monero-project/monero-gui/blob/e9cd4588aef3f0808ce153957a504f43dcdbeb26/src/daemon/DaemonManager.cpp#L160) is not quite in sync so it'll hang (in 3m30s chunks) until the 2 miraculously sync up.

## Change
When launching `monerod`, launch it with a timeout of `5 seconds` before assuming something has gone wrong and return `false`, as in we failed to launch.

This lowers the "is monerod running?" poll-rate to every 5 seconds which makes the error screen surface around the `120s` mark as intended.

### DIFF
--- a/src/daemon/DaemonManager.cpp
+++ b/src/daemon/DaemonManager.cpp
@@ -146,7 +146,7 @@ void DaemonManager::stopAsync(NetworkType::Type nettype, const QString &dataDir,
 {
     const auto feature = m_scheduler.run([this, nettype, dataDir] {
         QString message;
-        sendCommand({"exit"}, nettype, dataDir, message);
+        sendCommand({"exit"}, nettype, dataDir, message, /* no timeout */ -1);
 
         return QJSValueList({stopWatcher(nettype, dataDir)});
     }, callback);
@@ -239,9 +239,21 @@ void DaemonManager::printError()
 bool DaemonManager::running(NetworkType::Type nettype, const QString &dataDir) const
 { 
     QString status;
-    sendCommand({"sync_info"}, nettype, dataDir, status);
+    bool started = sendCommand(
+        {"sync_info"},
+        nettype,
+        dataDir,
+        status,
+        // milliseconds before the command
+        // timesout and returns `false`.
+        5000
+    );
+
     qDebug() << status;
-    return status.contains("Height:");
+
+    // Did `monerod` start successfully
+	// and print out its height?
+    return started && status.contains("Height:");
 }
 
 bool DaemonManager::noSync() const noexcept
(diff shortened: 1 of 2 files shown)


plowsof:
could you help me reproduce the issue? i had first tried to bind a monero-wallet-rpc client to the same port the 'launched by monero-gui' monerod would attempt to use for its --zmq-port, but i noticed the 120 second timeout

hinto-janai:
1. Block port 18081/18083
2. Start `monerod` from Monero GUI with no special arguments
3. See no error after 120 second timeout

https://github.com/monero-project/monero-gui/assets/101352116/fe073827-4bb2-4f62-9704-0a26b6258546

plowsof:
```
$ nc -lp 18081 > /dev/null
```
i still see the failed to start after 120 seconds, but im half certain i experienced a loop longer than 120 , strange
[Screencast from 15-01-24 08:16:14.webm](https://github.com/monero-project/monero-gui/assets/77655812/084a4812-6632-4d93-96e9-e872a92ebe95)

plowsof:
```  
nc -lk 127.0.0.1 18081 > /dev/null
```
without `k` netcat exits after the first connection attempt , so it no longer holds the port, to explain my above issue with not reliably reproducing it

plowsof:
reproduced and this PR fixes it, will need a rebase, lgtm!

nahuhh:
@plowsof  i think 5 seconds might be too short? May be fine of a faster device, or with a small txpool

plowsof:
my first thought from looking at the gui was this is just updating the ui , that its a soft fail, update the ui every 5 seconds, after 120 display the error page and when you click ok on the error - the loop is still continuing

PR #4284, opened 2024-03-09
Dockerfile: bump zlib version to 1.3.1

selsta:


tobtoht:
Suggestion to download from https://github.com/madler/zlib/releases instead. The zlib.net link gets invalidated every time there is a new release it seems.

selsta:
Updated

tobtoht (approved):

PR #4263, opened 2024-01-21
I2p integration (WIP)

preland:

PR #4337, opened 2024-08-18
Implement I2P integration

matthias-hoste:
This is a basic I2P integration with I2Pd. I went off of a seemingly abandoned PR(https://github.com/monero-project/monero-gui/pull/4263)
Right now after selecting a wallet mode you have the option to use or not use I2P as a tx-proxy.

Left to do for me:
- [ ] Advanced mode option to only use I2P(create a tunnel and set an anonymous inbound)
- [ ] Add option for pre-existing I2P routers(in case you are running java I2P)
- [ ] Add I2P network stats page? Right now it's not even obvious if it's running or what it's doing
- [ ] Clean up code, daemon args look very unclean

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,6 @@
 [submodule "external/quirc"]
 	path = external/quirc
 	url = https://github.com/dlbeer/quirc/
+[submodule "external/i2pd"]
+	path = external/i2pd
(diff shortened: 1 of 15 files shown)


selsta:
Closing, please see #4578

Issue #4341, opened 2024-08-22
[GUI][Restore wallet] 'Create wallet' stays disabled until monero-wallet-gui is closed


b4n6-b4n6:
Kindly observe video @ 00:37 where 'Create wallet' button stays disabled even though mnemonic seed has been corrected
[restore-bug-2.webm](https://github.com/user-attachments/assets/4ec8af69-94a4-41c2-aaec-9f8dda5d2d51)

nahuhh:
Confirmed

plowsof:
newlines / extra whitespaces can be stripped from the users seed to pass verification and sent to the wallet api. there would be concerns with merging this as we are modifying the structure of the users seed and potentially creating a different wallet. there is always 1 person who will enter words 1, 6, 11, 16, 21 on the first line lol

nahuhh:
The whitepaces and lines seem to be stripped properly

the problem is that the "create wallet" button remains greyed out indefinitely (until restarting the wallet)

b4n6-b4n6:
> newlines / extra whitespaces can be stripped from the users seed to pass verification and sent to the wallet api. there would be concerns with merging this as we are modifying the structure of the users seed and potentially creating a different wallet. there is always 1 person who will enter words 1, 6, 11, 16, 21 on the first line lol
> 
> to avoid that concern, the restore wallet from seed screen would be the same window as the seed preview with numbers next to each word.

Kindly
Whitespace in seed is not relevaant to this issue

the problem is that the "create wallet" button remains greyed out indefinitely (until restarting the wallet)

b4n6-b4n6:
Fix [PR](https://github.com/monero-project/monero-gui/pull/4346)

PR #4346, opened 2024-08-31
Fix 'Create wallet' button getting stuck until restart when invalid s…

b4n6-b4n6:
Fixes https://github.com/monero-project/monero-gui/issues/4341

Please review and feedback

### DIFF
--- a/wizard/WizardRestoreWallet4.qml
+++ b/wizard/WizardRestoreWallet4.qml
@@ -82,11 +82,13 @@ Rectangle {
                     btnNext.enabled = false;
                     wizardController.wizardStateView.wizardRestoreWallet2View.pwField = "";
                     wizardController.wizardStateView.wizardRestoreWallet2View.pwConfirmField = "";
-                    wizardController.recoveryWallet();
-                    wizardController.writeWallet(function() {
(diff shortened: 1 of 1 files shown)


wizard/WizardRestoreWallet4.qml
```diff
@@ -82,11 +82,13 @@ Rectangle {
...
-                        wizardController.useMoneroClicked();
-                        btnNext.enabled = true;
-                    });
+                    var recoveryResult = wizardController.recoveryWallet();
+                    if (recoveryResult) {
+                        wizardController.writeWallet(function () {
+                            wizardController.useMoneroClicked();
+                        });
+                    }
+                    btnNext.enabled = true;
```
b4n6-b4n6, line 91: `wizard/WizardCreateWallet5.qml` also uses `btnNext.enabled = true` outside of the `wizardController.writeWallet(...)` scope

plowsof:
i can check this one later / EOD

selsta:
@b4n6-b4n6 can you test if https://github.com/monero-project/monero-gui/pull/4351 also solves the issue? if yes, I would closse this here because the other PR is simpler

PR #4362, opened 2024-10-01
Dockerfile.linux: freetype.git mirror

plowsof:
https://savannah.gnu.org/ apears to be temp offline


plowsof:
may require freetype2 > freetype, will let workflow run first

https://github.com/monero-project/monero-gui/blob/08e2eafb7cfc18fc08f4b74885fa682001f12bca/Dockerfile.linux#L206

nahuhh (approved):

PR #4407, opened 2025-01-31
[WIP] Migrate to Qt6

Malinero:
https://github.com/monero-project/monero-gui/issues/4348

Still a number of issues to be fixed, but the first CI passes

PR #4412, opened 2025-02-19
Win deploy: Boost Regex is header-only since 1.77

Malinero:
same as https://github.com/monero-project/monero/blob/master/CMakeLists.txt#L1048

### DIFF
--- a/cmake/Deploy.cmake
+++ b/cmake/Deploy.cmake
@@ -58,7 +58,6 @@ if(APPLE OR (WIN32 AND NOT STATIC))
             libboost_filesystem-mt.dll
             libboost_locale-mt.dll
             libboost_program_options-mt.dll
-            libboost_regex-mt.dll
             libboost_serialization-mt.dll
             libboost_thread-mt.dll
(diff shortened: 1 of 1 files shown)


selsta:
While the patch itself looks good, why is this needed? Is deploy currently broken?

1.77 came out a while ago.

Malinero:
I think it's currently broken. I don't have a windows, but while trying to debug the windows CI on #4407, I saw this https://github.com/Malinero/monero-gui/actions/runs/13379636601/job/37365795118#step:4:761

Result of : 

```
        foreach(ITEM ${WIN_DEPLOY_DLLS})
            find_file(${ITEM}_VAR ${ITEM} REQUIRED)
        endforeach()
```

selsta (approved):
Not sure why this only broke now.

tobtoht:
>Not sure why this only broke now.

MSYS2 used to ship a static compatibility library and stopped doing that a few months ago.

tobtoht (approved):

PR #4419, opened 2025-03-10
Enforce whonix opengl renderer

everoddandeven:
This PR contains a simple hack to force video rendering mode to `opengl` when the wallet is running in a Whonix environment, fixing a bug described here:

https://bounties.monero.social/posts/182/1-123m-make-monero-wallet-gui-whonix-friendly

> GUI would clear screen improperly leading to shredding and unreadable GUI - to workaround this one had to click away and then it would clear screen and render GUI properly

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -68,6 +68,7 @@
 #include "qt/updater.h"
 #include "qt/utils.h"
 #include "qt/TailsOS.h"
+#include "qt/WhonixOS.h"
 #include "qt/KeysFiles.h"
 #include "qt/MoneroSettings.h"
 #include "qt/NetworkAccessBlockingFactory.h"
@@ -154,6 +155,7 @@ bool isWindows = false;
 bool isMac = false;
 bool isLinux = false;
 bool isTails = false;
+bool isWhonix = false;
 bool isDesktop = false;
 bool isOpenGL = true;
 bool isARM = false;
@@ -175,6 +177,7 @@ int main(int argc, char *argv[])
 #elif defined(Q_OS_LINUX)
     bool isLinux = true;
     bool isTails = TailsOS::detect();
+    bool isWhonix = WhonixOS::detect();
 #elif defined(Q_OS_MAC)
     bool isMac = true;
 #endif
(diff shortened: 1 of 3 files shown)


selsta:
This fails to compile on some CI jobs, could you take a look?

Also which renderer is used by defaut, if not opengl? Software renderer?

src/main/main.cpp
```diff
@@ -175,6 +176,7 @@ int main(int argc, char *argv[])
 #elif defined(Q_OS_LINUX)
     bool isLinux = true;
     bool isTails = TailsOS::detect();
+    bool isWhonix = WhonixOS::detect();
```
plowsof, line 179: note how isTails is defined as a bool on line 157 to stop none linux OS throwing the not defined error, same is needed for isWhonix

everoddandeven:
> This fails to compile on some CI jobs, could you take a look?

It should be fine now. I see macos build is failing but I think it's not my fault.

> Also which renderer is used by defaut, if not opengl? Software renderer?

Yes, precisely `softwarecontext`.

b4n6-b4n6:
I can confirm this branch fixes the issue: triggers the openGL graphics mode and monero-wallet-gui can be used without issue on whonix

before:

Screenshot from 2025-03-11 21-53-37

image

after:

Screenshot from 2025-03-11 21-51-18

selsta:
Why does it use `softwarecontext` in the first place? Default should be `opengl`.

everoddandeven:
> Why does it use `softwarecontext` in the first place? Default should be `opengl`.

I don't know why it was `softwarecontext` at first place in some cases on whonix. But with this fix we make sure that it is always `opengl`.

b4n6-b4n6:
> Why does it use `softwarecontext` in the first place? Default should be `opengl`.

I don't know either, but I reproduced this on 4 machines I believe on Virtualbox+Whonix

plowsof:
i confirm the bug is seen when using virtualbox. If i attempt to enable 3d acceleration in virtualbox settings, then the Whonix VM fails to start.

PromptPunksFauxCough:
When `QMLSCENE_DEVICE=opengl` is applied at launch:

1) The following errata is returned:
```
W Could not create scene graph context for backend 'opengl' - check that plugins are installed correctly in /usr/plugins
```
2) Undesirably Usable

3) Introduces noticeable lag coupled with occasional, minor graphics tearing

---

For further clarity, tagging [@adrelanos](https://github.com/adrelanos) (former package maintainer).

Issue #2878, opened 2020-05-01
Automatic fallback to softwarecontext renderer


selsta:
This wasn't fixed. Please re-open.

I also have 100% CPU using monero GUI `v0.15.0.4` in a VirtualBox VM (Whonix-Workstation). It's just the first screen (language selection screen). No monero installed ever in that VM before.

`QMLSCENE_DEVICE=softwarecontext ./monero-wallet-gui` helps as a workaround but it's not a solution. Users shouldn't be required to figure that out the hard way. Should be the default or some other fix.

Also my CPU use is now down to 30% which still seems to much for just an idle language selection screen.

_Originally posted by @adrelanos in https://github.com/monero-project/monero-gui/issues/2238#issuecomment-622497250_

selsta:
The best solution would be automatic fallback to softwarecontext in systems which have trouble with OpenGL. If this is not possible then I guess this will be a wontfix.

> Should be the default

Setting the softwarecontext renderer as default will result in a worse experience for the majority of users. The renderer has worse quality, supports less features AFAIK also uses more CPU. See https://doc.qt.io/QtQuick2DRenderer/qtquick2drenderer-performance.html

adrelanos:
Btw as for Whonix integration (which runs mostly in VMs), I guess would be best if some script [1] checked if hardware rending is enabled. And if not, set `QMLSCENE_DEVICE=softwarecontext` as global environment variable by default.

How to test from command line if hardware acceleration is available?

selsta:
>> Should be the default
> 
> Setting the softwarecontext renderer as default will result in a worse experience for the majority of users. The renderer has worse quality, supports less features AFAIK also uses more CPU. See https://doc.qt.io/QtQuick2DRenderer/qtquick2drenderer-performance.html

I agree. softwarecontext renderer should if anything only be default when needed. Not for everyone that doesn't need it.

----

[1] Has nothing to do with monero. The name of package is [vm-config-dist](https://github.com/Whonix/vm-config-dist). I don't suggest to implement such a script in monero that changes environment variables globally. Just wondering for a good solution for Whonix / VMs.

adrelanos:
[Test command from inside VM to detect if VirtualBox 3D acceleration is enabled or disabled?](https://forums.virtualbox.org/viewtopic.php?f=3&t=97983)

adrelanos:
I guess this needs a separate ticket.
30% CPU in Debian buster based VirtualBox VM
[...]

PR #4424, opened 2025-03-25
Enforce renderer to opengl if virtualbox 3D acceleration is detected

everoddandeven:
This PR contains a simple fix to allow setting the renderer to opengl when running monero-wallet-gui under virtualbox with 3D acceleration enabled.

Related #2878 

Closes #4419

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -68,6 +68,7 @@
 #include "qt/updater.h"
 #include "qt/utils.h"
 #include "qt/TailsOS.h"
+#include "qt/VirtualBox.h"
 #include "qt/KeysFiles.h"
 #include "qt/MoneroSettings.h"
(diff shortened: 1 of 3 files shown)


plowsof:
sadly i can not start the virtual machine with 3d acceleration enabled in virtualbox, even after maxing out the available RAM, likely a me problem

everoddandeven:
I updated the code, and tested on a lubuntu 24.04 vm, because whonix crashes with 3D acceleration enabled. The wallet is now able to choose the correct renderer, and I no longer encounter the glitches. What do you think @adrelanos ?

adrelanos:
I am unavailable as a reviewer.

Changes on Kicksecure, Whonix side:
https://github.com/Kicksecure/vm-config-dist/blob/master/etc/profile.d/20_software_rendering_in_vms.sh

PR #4425, opened 2025-03-27
[Libwalletqt] TransactionHistory adjustment for Monero Issue 9866

Tzadiko:
Amendment to the GUI as a result of a change made here to src/wallet/api/wallet2_api.h

https://github.com/monero-project/monero/pull/9868

### DIFF
--- a/src/libwalletqt/TransactionHistory.cpp
+++ b/src/libwalletqt/TransactionHistory.cpp
@@ -80,7 +80,7 @@ void TransactionHistory::refresh(quint32 accountIndex)
 
         m_pimpl->refresh();
         for (const auto i : m_pimpl->getAll()) {
-            if (i->subaddrAccount() != accountIndex) {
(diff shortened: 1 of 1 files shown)


src/libwalletqt/TransactionHistory.cpp
```diff
@@ -80,7 +80,7 @@ void TransactionHistory::refresh(quint32 accountIndex)
 
         m_pimpl->refresh();
         for (const auto i : m_pimpl->getAll()) {
```
SChernykh, line 82: I think you need to add `&` to `const auto`, or it will make a copy of each transaction in this loop.

Tzadiko: Did not notice this, good point - changed.

Tzadiko:
Closed with https://github.com/monero-project/monero/pull/9868

Issue #4466, opened 2025-06-28
Change the word 'background' in GUI mining to 'pause when active'


PyXMR2025:
Change the word 'background' in GUI mining to 'pause when active'

In GUI, the term 'background' does not intuitively imply that it refers to mining when the user is idle. Perhaps we need to change it.

After discussion in the Monero GUI channel.

PyXMR2025:
newbie20250619:newbie20250619
@day20250619:matrix.org
|
05:24 PM

Jackie

@day20250619:matrix.org If possible, register a Github account and open it https://github.com/monero-project/monero-gui/pull/4467/files Click on 'Approve' in the 'Review changes' section and then click on' Submit review '.

I have a Github account.
259 - text: qsTr("Background mining (experimental)") + translationManager.emptyString
259 + text: qsTr("Pause when active (experimental)") + translationManager.emptyString
At least twice I pointed that I don't like "pause when active" phrase. I'd rather prefer "Mine only when idle".
Maybe somebody else can approve "pause when active" instead of me.

PR #4467, opened 2025-06-29
Mining.qml: Change 'Background mining' to 'Pause when active'

PyXMR2025:
Mining.qml: Change 'Background mining' to 'Pause when active' 
https://github.com/monero-project/monero-gui/issues/4466


skzxc-fz:
There are two voices in the community now: 1. pause when active 2. mine only when idle

PR #4482, opened 2025-07-24
p2pool: add nano sidechain

nahuhh:
@devhyper @SChernykh 

(also added missing flags)


SChernykh:
I'm not sure that using `nano` by default is a good idea. It finds blocks every 4-5 days on average, so it can spend 3 weeks without finding a block. `mini` is still the best default choice.

nahuhh:
I changed it so nano is just appended to the list, keeping mini as the default

selsta (approved):

PR #4497, opened 2025-09-01
workflows: remove unmaintained docker caching action

selsta:
Does not work anymore. Action hasn't been updated since 2020. There's a fork but it also appears unmaintained.


plowsof (approved):

plowsof:
We can still use githubs generic cache to zip / extract docker images and rebuild on tag etc which iirc is what i was doong in https://github.com/monero-project/monero-gui/pull/4367

tobtoht (approved):

PR #3868, opened 2022-03-20
StandardDropdown: Add Flickable and ScrollBar to droplist

reemuru:
This commit adds a Flickable and ScrollBar to the
StandardDropdown component. The use case is for the list
of fiat currencies which is flowing off-screen. Tested on
ASUSTek laptop touchpad.

Fixes: 302490e ("Add additional fiat currencies")

It is not perfect since the droplist slightly slides above the
selector, however, that is as close as I could get it.

PR #4501, opened 2025-09-04
feat: add pagination on the Receive page (optional)

### LINKED ISSUE #4500: Bug: Receive page lags or hangs for large wallets
Opening a wallet with many subaddresses causes severe lag and impact usability.

To reproduce the issue, create a wallet with many addresses in monero-wallet-cli.  For example, using:
```
address mnew 1000
```
30+ times in monero-wallet-cli will create a wallet with 30,001 addresses (like a very active BTCPay wallet).  With this many addresses, the app will lag on a performant PC.

sneurlax:
Closes https://github.com/monero-project/monero-gui/issues/4500.

Without this feature, opening a wallet with many subaddresses causes severe lag and impacts usability.  This change introduces optional pagination to the subaddresses in the Receive tab.

To reproduce the issue, create a wallet with many addresses in monero-wallet-cli.  For example, using:
```
address mnew 1000
```
30+ times in monero-wallet-cli will create a wallet with 30,001 addresses (like a very active BTCPay wallet).  Before this feature and without it enabled (disabled by default), the app will lag on a performant PC when navigating to the Receive page.  With this feature, it shows a "Loading addresses..." message while loading the first "page" of addresses.  As a Quality of Life (QoL) feature, a "Scroll to bottom" button allows loading all remaining addresses at once.

  - Add pagination support to the Subaddress backend.
  - Add "Enable subaddress pagination" setting to Interface section.
  - When pagination is disabled (default): behaves exactly like original GUI.
  - When pagination is enabled:
    - Load first 100 subaddresses (the default page size) initially.
    - "Scroll to bottom" button loads all remaining addresses.
    - Loading footer shows progress during fetch operations.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -1521,6 +1521,8 @@ ApplicationWindow {
         property bool autosave: true
         property int autosaveMinutes: 10
         property bool pruneBlockchain: false
+        property bool enableSubaddressPagination: false
+        property int subaddressPageSize: 100
 
         property bool fiatPriceEnabled: false
         property bool fiatPriceToggle: false
--- a/pages/Receive.qml
+++ b/pages/Receive.qml
@@ -551,13 +551,61 @@ Rectangle {
                 }
             }
 
+            Rectangle {
+                id: primaryAddressContainer
+                Layout.fillWidth: true
+                Layout.preferredHeight: 50
+                Layout.topMargin: 6
+                color: MoneroComponents.Style.titleBarButtonHoverColor
+                radius: 3
+                visible: persistentSettings.enableSubaddressPagination && (!subaddressListView.model || subaddressListView.count === 0)
+                
+                Rectangle {
+                    anchors.top: parent.top
+                    anchors.bottom: parent.bottom
+                    width: 2
+                    color: MoneroComponents.Style.accountColors[0]
+                }
+                
+                MoneroComponents.Label {
+                    anchors.left: parent.left
+                    anchors.leftMargin: 12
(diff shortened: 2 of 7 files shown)


selsta:
> Load first 100 subaddresses (the default page size) initially.

Is it possible to make this dynamic? Because on a smaller screen you now have two different ways to scroll, basically I would have to scroll both on the outer and inner layer to the bottom.

Screensho

sneurlax:
> > Load first 100 subaddresses (the default page size) initially.
> 
> Is it possible to make this dynamic? Because on a smaller screen you now have two different ways to scroll, basically I would have to scroll both on the outer and inner layer to the bottom.
> 
> Screensho

Yes, I noticed that but accepted it.  It's definitely possible to make it dynamic to try and fit the window size to try to only get one scrollable section

I wonder how or if to try to resize that dynamically as the window is resized, though.  I'd appreciate any more details / specificity you can provide if you'd be so kind as to expound upon what you'd like to see

plowsof:
Reg "dynamic" - i am no QT dev, but i also attempted through blood sweat and tears to add some kind of dynamic resizing and failed. https://github.com/monero-project/monero-gui/pull/3868 is related as this would also benefit from that.

selsta:
I have opened an alternative here: https://github.com/monero-project/monero-gui/pull/4667

It avoids a new pagination setting while still being performant.

sneurlax:
Thanks selsta, this can be closed!

PR #4520, opened 2025-11-06
slightly shortened human readable date time

HannesGitH:
e.g. 
- `1 day(s) ago` -> `1 day ago`
- `2 day(s) ago` -> `2 days ago`

### DIFF
--- a/translations/monero-core.ts
+++ b/translations/monero-core.ts
@@ -3333,32 +3333,36 @@ For the case with Spend Proof, you don&apos;t need to specify the recipient addr
         <location filename="../js/Utils.js" line="59"/>
         <source>%n second(s) ago</source>
         <comment>0</comment>
-        <translation type="unfinished">
-            <numerusform></numerusform>
+        <translation>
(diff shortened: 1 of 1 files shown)


selsta (commented):
Does this take into account that delta is a `float`?

HannesGitH:
honestly, I had no idea how plurals would work here (never used qt), but now I dug into it and I think its now fixed the 'correct way'

selsta:
The .ts are machine generated and should not be touched manually, the code change needs to be in the .qml itself. I don't think we got plurals working correctly in other places.

HannesGitH:
> I don't think we got plurals working correctly in other places

At least you got the same setup I put in the english file already in ther german one, so I suppose it might already be working, didnt check yet tho.

> The .ts are machine generated and should not be touched manually,

then this changes shall probably be made in weblate.. Guess you are right w that

selsta:
Closing as we can't accept changes to the translation files directly. Weblate is still work in progress in getting the platform back up running.

PR #4550, opened 2026-01-14
cmake: find windeployqt with qt5 suffix

selsta:
Should also fix Windows CI.


selsta:
CI is still failing because mingw64 `qmake` has been renamed to `qmake-qt5`, which can't be found by windeployqt. Not sure yet how to solve this in a clean way.

PR #4552, opened 2026-01-18
DaemonManager: use accurate blockchain size estimate

selsta:
It seems it always used the inaccurate 70GB number.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -94,7 +94,7 @@ ApplicationWindow {
     readonly property string localDaemonAddress : "localhost:" + getDefaultDaemonRpcPort(persistentSettings.nettype)
     property string currentDaemonAddress;
     property int disconnectedEpoch: 0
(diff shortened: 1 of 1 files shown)


src/daemon/DaemonManager.cpp
```diff
@@ -314,7 +314,7 @@ QVariantMap DaemonManager::validateDataDir(const QString &dataDir) const
 
         // Make sure there is 75GB storage available
         storageAvailable = storage.bytesAvailable()/1000/1000/1000;
-        if (storageAvailable < 75) {
+        if (storageAvailable < estimatedBlockchainSize) {
```
plowsof: :+1: nice. can also edit the comment here above e.g. `// Make sure there is enough storage available`

plowsof:
A follow up PR or here to bump the blockchain estimate (which says 100GB) https://docs.getmonero.org/technical-specs/#current-blockchain-size

plowsof (approved):

Issue #4286, opened 2024-03-10
Monero-wallet-gui won't quit on Mac-ARM-Sonoma


kirk-kerekes:
Monero-wallet-gui won't quit on Mac-ARM (specifically MacOS 14.4 on M2 Pro)

After stopping Mining,

When Quit is invoked, the GUI just goes unresponsive. 

Ultimately force-quit is required.

selsta:
This doesn't happen when you don't mine?

kirk-kerekes:
Good question. 
So I --
* force quit the app, 
* restarted the app, 
* waited until it was synchronized but...
* did not start mining and ...
* it once again hung at quit.

Since it appears to be the GUI itself that is locking up, I assume that the problem is in whatever UI framework(s) manage that. When it is hung, _nothing_ in the Monero gui works -- menu bar, cursor updating, nothing. And it is (in human terms) instantaneous. No delay between selecting Quit and the lockup. Because this is rather hard to miss, I assume that the issue is a conflict, but other than the VPN, I have nothing installed that is supposed to be breaking the rules of app isolation. No UI modifiers, no jailbreaking. The VPN uses Apple's VPN scheme. 

Variances from the norm:

* I am running a VPN (WireGuard, using the official Wireguard app)
* I am also running Apple's Private Relay
* WiFi is on, but not connected to anything. (probably a rare state)
* I am controlling the Mac via ARD, from an iPad Pro running Screens.
* I am running iStat Menus.

selsta:
v0.18.3.2 has updated Qt from 5.15.9 to 5.15.12, can you test with it once it is out later today?

kirk-kerekes:
I tried v0.18.3.2, no change in the hang-on-quit. 

But the mining is 6X faster! (From approximately 500H/s to 3000+H/s)

I preserved the crash-dump generated by the force quit. If you can use it, let me know how best to send it.

selsta:
I think you can drag and drop it in a GitHub comment.

kirk-kerekes:
[Monero Gui Hang Dump.txt](https://github.com/monero-project/monero-gui/files/14565477/Monero.Gui.Hang.Dump.txt)

Done!

selsta:
Can you also open Console.app, and select `monero-wallet-gui` under `Crash report` and share that? It's easier for me to read.

kirk-kerekes:
[...]

PR #4555, opened 2026-01-19
FutureScheduler: use custom thread pool

selsta:
Might help with app freezing on shutdown.


plowsof:
solves an issue here: https://github.com/monero-project/monero-gui/issues/4286#issuecomment-3770147593

just waiting on others to test, but - the GUI behaves as normal on Ubuntu so i can approve on nothing changing - start/stop gui while node is/isn't running

selsta:
Replaced by #4558

PR #4556, opened 2026-01-19
WizardController: set password for temp wallet

selsta:


plowsof:
without: (log level 2 or 3 -> open existing wallet , set log level -> close wallet -> create new one)
```
Creating temporary wallet /tmp/monero-core.yCraNL
```
```
/tmp$ ls
monero-core.yCraNL
monero-core.yCraNL.address.txt
monero-core.yCraNL.keys      
```
copy the file to home folder, then it opens.

with:
```
Creating temporary wallet /tmp/monero-core.ySvjRE
```
couldn't open requires password.

now testing the normal flow of creating wallet, no problem, adding a password to the new wallet , fine.

ideally this process would occur 100% in ram writing no temp wallet artefacts to disk

PR #4558, opened 2026-01-20
filter: intercept quit event to avoid deadlock

selsta:
Resolves https://github.com/monero-project/monero-gui/issues/4286


plowsof:
no changes on Ubuntu linux :+1:

PR #4182, opened 2023-05-30
p2pool: Detailed statistics

devhyper:
Display for full p2pool statistics from stats folder.
stats

PR #4566, opened 2026-02-08
mining: detailed dashboard for p2pool

rssoc:
This PR implements a [long requested community feature](https://bounties.monero.social/posts/68/6-026m-statistical-mining-dashboard-for-p2pool-in-monero-gui-wallet-2m) intended to educate newcomers on mining concepts and expose them to p2pool.

This adds:
- **A New Dashboard View**: Accessible from "Advanced > Mining" when "P2Pool" is selected as the mining mode. Implements buttons with transitions and theme compatibility, as well as a user interface and experience matching the rest of the program.
- **Educational Resources**: An on-demand explanation for each dashboard field to help educate newcomers on p2pool and mining terminology and concepts.
- **Refined Data Visualization**: Friendly readouts for hashrate, with rolling averages, and mining effort, alongside an "Advanced" section for those who wish to view the raw API data.
- **Synchronization Gating**: Waits until the side and blockchains are synced to address some consistency issues. The initial synchronization takes some time, and the reporting from the API is a bit eager since we're using P2Pools built-in miner which itself doesn't wait.

This is intended to supersede #4182 which was a partial implementation lacking most of the important points above.


src/p2pool/P2PoolManager.cpp
```diff
@@ -130,25 +130,23 @@ bool P2PoolManager::isInstalled() {
 }
 
 void P2PoolManager::getStatus() {
```
thomasbuilds, line 134: **Issue: `getStatus()` has hidden side effects on EMA / effort state**

`getStatus()` is called on its own cadence by `Mining.qml`, but it now invokes `fetchLocal()`, which mutates `m_hashrate_ema15m/1h/24h`, `m_effort`, `m_round_hashes` and `m_shares_found`. However, `m_lastUpdate` is only assigned in `update()` never in `fetchLocal()`.

Consequences:

- While the user sits on the Mining page, every `getStatus()` tick advances the EMAs but `m_lastUpdate` keeps growing stale, so `secs = m_lastUpdate.secsTo(now)` blows up and `alpha → 1.0` each new sample fully replaces the EMA, defeating the smoothing.
- When the dashboard is then opened, `update()` sees state already corrupted by `getStatus()` calls.
- Two callers (Mining tick + Dashboard 5 s timer) interleave and double-count samples within a window.

**Suggested fix:** split `fetchLocal()` into a *pure* formatter (no member writes) used by `getStatus()`, and keep the stateful EMA/effort step inside `update()` only. Roughly:

```cpp
// pure: just reads files and formats; no member writes
QVariantMap fetchLocalSnapshot() const;

// stateful: snapshot + advance EMAs + set m_lastUpdate
void recomputeAggregates(const QVariantMap& local, const QVariantMap& pool);
```

Then `getStatus()` calls `fetchLocalSnapshot()` and `update()` calls both.

src/p2pool/P2PoolStatsProvider.h
thomasbuilds, line 1: **Issue: holding `const T&` to `P2PoolManager`'s members is a footgun**

`P2PoolStatsProvider` keeps `const bool& m_started` and `const QString& m_p2poolPath` referencing fields owned by `P2PoolManager`. It works today only because the parent owns the provider via `unique_ptr` and outlives it. Any future refactor that:

- moves the provider onto a worker thread,
- copies/moves it,
- or reorders member destruction in `P2PoolManager`,

will silently produce UB the references will dangle or race with writes in `start()`/`exit()` with no compiler warning.

**Suggested fix:** take a back-pointer to the parent and add small accessors, or pass values + setters. For example:

```cpp
explicit P2PoolStatsProvider(P2PoolManager* parent);
// ...
private:
    P2PoolManager* m_manager;  // non-owning, parent guarantees lifetime
    bool started() const { return m_manager->isStarted(); }
    QString poolPath() const { return m_manager->p2poolPath(); }
```

This also keeps the threading story explicit if you ever move stats fetching off the GUI thread.

src/p2pool/P2PoolStatsProvider.cpp
```diff
@@ -0,0 +1,353 @@
...
+     QVariantMap network =
+          fetch("network/stats");
+
+     quint64 height =
+          network["height"].toULongLong();
+
+     return height > 0;
+}
+
+void P2PoolStatsProvider::update()
```
thomasbuilds, line 320: **Issue: each `update()` re-reads & re-parses the same JSON files ~10×**

Within a single tick:

- `fetchLocal()` reads `local/miner` + `pool/stats`, then calls `isSidechainReady()` which reads them **again**.
- `fetchPool()` reads `pool/stats` + `local/miner` and again calls `isSidechainReady()`.
- `fetchNetwork()` reads `network/stats` and `isNetworkReady()` reads it again.
- `fetchRaw()` reads everything one more time.

That's roughly 10 file opens + JSON parses every 5 s, all redundant.

**Suggested fix:** parse each file once at the top of `update()` and pass the maps down:

```cpp
void P2PoolStatsProvider::update()
{
    const QVariantMap localRaw   = fetch("local/miner");
    const QVariantMap poolRaw    = fetch("pool/stats");
    const QVariantMap poolStats  = poolRaw["pool_statistics"].toMap();
    const QVariantMap networkRaw = fetch("network/stats");

    const bool sidechainReady = poolStats["sidechainHeight"].toULongLong() > 1000
                              && localRaw["block_reward_share_percent"].toInt() != 100;
    const bool networkReady   = networkRaw["height"].toULongLong() > 0;

    auto local   = buildLocal(localRaw, poolStats, sidechainReady);
    auto pool    = buildPool(poolStats, localRaw, sidechainReady);
    auto network = buildNetwork(networkRaw, networkReady);
    auto raw     = buildRaw(localRaw, poolRaw, networkRaw);

    m_lastUpdate = QDateTime::currentDateTime();
    emit p2poolUpdateStats(local, pool, network, raw);
}
```

thomasbuilds:
also there are way too many commits imo

src/p2pool/P2PoolManager.cpp
rssoc: Hmm...
> While the user sits on the Mining page, every getStatus() tick advances the EMAs but m_lastUpdate keeps growing stale, so secs = m_lastUpdate.secsTo(now) blows up and alpha → 1.0 each new sample fully replaces the EMA, defeating the smoothing.

There is a degradation in the accuracy in the EMA, depending on how long you're not on the dashboard. I don't think most users would notice it, but yeah it does bother me that it is happening. Though I don't like the proposed fix because EMA should be modeled as a property on the miner. Given that nothing's using the `m_lastUpdate` property besides the `fetchLocal()` for EMA calculations, it's probably better if we just rename it to  `m_lastEMAUpdate` and push the maintenance of that into `fetchLocal()`.

> Two callers (Mining tick + Dashboard 5 s timer) interleave and double-count samples within a window.

I don't think this was actually an issue because only one timer is active at a time (depending on your view).

src/p2pool/P2PoolStatsProvider.h
rssoc: This was an intentional choice by me because:

> Any future refactor that: ...

I don't see these happening., and it's much simpiler to hold a `const T&` when we know it's going to be correct by construction. Writing wrappers to an object that we know will always exist together with the `P2PoolManager` seems a bit premature to me. 

If that changes in the future for any reason, I'd say it's better handled then anyway.

src/p2pool/P2PoolStatsProvider.cpp
rssoc: The operating system already caches the files for you, so it's not really re-read each time. This was intentional; I don't think what you'd get out by doing this is worth the complication or the cost in clarity (especially given that this function is only called once every 5 seconds, or so).

rssoc:
> also there are way too many commits imo

I broke the commits up in a way s.t. things that a reviewer might be interested in seeing would have good visibility. 

I appreciate the feedback though, and I would like to hear what the thoughts are from the maintainers. I know some things have been going on in the background regarding `P2Pool` internals, and I'm just waiting for the next point release so that I can check the branch against that and have an upstream merge in the near future, hopefully. 

I've gotten positive feedback from others who have used the dashboard thus far.

src/p2pool/P2PoolManager.cpp
rssoc: > it's probably better if we just rename it to `m_lastEMAUpdate` and push the maintenance of that into `fetchLocal()`.

Done in https://github.com/monero-project/monero-gui/pull/4566/commits/49cb3bffa9896edfc064d97ac23fd4504a195d4d.

nahuhh:
Should drop the merge commits. Its not an official rule on the gui repo, but it is on other monero-project repos

if you need to update the branch, use rebase

rssoc:
Thanks, yeah that makes sense. @selsta, would you prefer that I rebase onto current master and drop the merge commits now? My only hesitation was that the resulting force-push might disrupt any review already underway. I’m happy to follow the preferred workflow.

Otherwise this is pretty much ready for review, if there are any other changes you’d like, please lmk and I can have them in. 

cc: @plowsof, in case you have any feedback w.r.t. functionality or anything.

nahuhh:
> Thanks, yeah that makes sense. @selsta, would you prefer that I rebase onto current master and drop the merge commits now? My only hesitation was that the resulting force-push might disrupt any review already underway. I’m happy to follow the preferred workflow.
> 
> Otherwise this is pretty much ready for review, if there are any other changes you’d like, please lmk and I can have them in. 
> 
> cc: @plowsof, in case you have any feedback w.r.t. functionality or anything.

There are a few oddities in the commits, such as, why are you reverting the copyright back to 2024? Also using 2014-2024 as date on new files

Before merge, the commits should be squashed, merge commits dropped / rebased onto master instead. But even before any of that, im wondering why you decided to revert the copyright? Or was that decision made by an llm?

pages/MiningDashboard.qml
```diff
@@ -0,0 +1,304 @@
+// Copyright (c) 2014-2024, The Monero Project
```
nahuhh: Example, this is a brand new file, it cant have a copyright from 2-12 years before it was created

src/p2pool/P2PoolManager.h
```diff
@@ -46,17 +48,21 @@ class P2PoolManager : public QObject
...
 
     enum DownloadError {
-        BinaryNotAvailable,
-        ConnectionIssue,
-        HashVerificationFailed,
-        InstallationFailed,
+	BinaryNotAvailable,
+	ConnectionIssue,
+	HashVerificationFailed,
+	InstallationFailed,
```
nahuhh: Don't know what happened here, but you seemo have replaced spaces with tabs.

the rest of the file uses spaces. Ensure that this wasnt done elsewhere

nahuhh (changes requested):

pages/MiningDashboard.qml
rssoc: Yes exactly :-)

I initially put a copyright up to 2026, but as I was looking through the conventions in the rest of the codebase, for one reason or another, everything had their copyright up to 2024 (even if they were created after that year). 

If that was some convention that I did not know about, I didn't want to stray from it - Hence why I made a separate commit bringing them down from "2026" -> "2024" since this was something I wanted some attention on ;-)

src/p2pool/P2PoolManager.h
rssoc: Apologies, this was from my Emacs. Good catch as I thought I had already cleaned these out from the commits when I prepared them.

pages/MiningDashboard.qml
nahuhh: IIUC, new files should just have the current year, edited files should have the year bumped

src/p2pool/P2PoolManager.h
rssoc: Addressed, thanks. This was the only instance that got through.

pages/MiningDashboard.qml
```diff
@@ -0,0 +1,304 @@
...
+                qsTr("Mining Dashboard") + translationManager.emptyString
+            description:
+                qsTr("View detailed statistics about your hardware, the pool, and the overall network.") + translationManager.emptyString + "<br><br>" +
+                qsTr("With P2Pool mining, payouts happen in bursts. You are paid only when the pool finds a block, and when you have a share in the 'window'. It can be normal to go long periods without seeing a payout.") + translationManager.emptyString + "<br><br>" +
+                qsTr("For more consistent payouts, use the calculators available for <a href='%1'>nano</a>, <a href='%2'>mini</a>, and <a href='%3'>main</a>, and choose the chain where your average share time most closely matches the pool's average block time.")
+                  .arg("https://nano.p2pool.observer/calculate-share-time")
+                  .arg("https://mini.p2pool.observer/calculate-share-time")
+                  .arg("https://p2pool.observer/calculate-share-time") + translationManager.emptyString + "<br><br>" + "<b>" +
+                qsTr("Learn more about each value by mousing over the label.") + translationManager.emptyString
```
jpk68: Nit: it would probably sound better to say "hovering over the label"

pages/MiningDashboard.qml
rssoc: I generally agree with you, so I went ahead and dropped that commit & bumped all created and modified files to 2026. Thanks.

pages/MiningDashboard.qml
```diff
@@ -0,0 +1,304 @@
...
+                            text: model.label
+                            tooltip: model.tooltip
+                        }
+                        DashboardValue { text: model.value }
+                    }
+                }
+            }
+
+            DashboardColumn {
+                DashboardHeader { text: qsTr("The Pool") + translationManager.emptyString }
```
jpk68: Nit: something like "P2Pool network" rather than "The Pool" might be more clear

rssoc:
> Before merge, the commits should be squashed, merge commits dropped / rebased onto master instead. 

Thanks, since I ultimately agree with you I went ahead and dropped off the merge commits - though I rebased off the last release instead of master (generally as a maintainer I find this to be smoother?). 

Not that there would have been conflicts anyway, but I at least have everything verified against the latest P2Pool we package.

> But even before any of that, im wondering why you decided to revert the copyright? 

See the reply I made to your review, there was a technicality. Though anyways I agree that we should probably be bumping the year, so I did that.  

> Or was that decision made by an llm?

Believe it or not, this PR is from a little bit before those times, at least, from my perspective ;-)

Thanks for all the feedback.

pages/MiningDashboard.qml
rssoc: I do like the consistency of "The" in "Your PC", "The Pool", "The Network", maybe: "Your PC", "The P2Pool Network", "The Monero Network"? 

Though "P2Pool Network" is a bit ambiguous in itself, since P2Pool consists of multiple side-chains (mini, nano, etc). The culmination of these could be considered the "network"... 

To the end user, maybe "The Pool" does make the most sense since those statistics are from the actual P2Pool Pool they're actually part of? lmk..

pages/MiningDashboard.qml
rssoc: Yeah agreed. Done.

pages/MiningDashboard.qml
jpk68: Personally, I think "Your machine", "P2Pool network", and "Monero network" would sound the best, but that's just me.

rssoc: After thinking about it I can see how it makes sense. I also much prefer "Your machine" to "Your PC", so I've adopted it as you say. Also I changed the header text to be less ambiguous as well:

```
qsTr("View detailed statistics about your hardware, the P2Pool, and the overall Monero network.") 
```

nahuhh: id probably still say "the P2Pool network" (instead of "the P2Pool"), but thats just me

nahuhh:
> Thanks, since I ultimately agree with you I went ahead and dropped off the merge commits - though I rebased off the last release instead of master (generally as a maintainer I find this to be smoother?).

GUI releases are cut from master, not from tags. 
Also, master is where this will be merged. You should rebase onto master (aka the base branch)

rssoc:
> GUI releases are cut from master, not from tags. Also, master is where this will be merged. You should rebase onto master (aka the base branch)

Yeah not a problem, I've rebased us onto `master`; and apologies for the extra force-push noise that was involved there as well (first force-push was rebased against a stale `master`).
 
Thank you everyone again for your comments!

PR #4575, opened 2026-03-21
installer: move p2pool to LocalAppData

selsta:


selsta:
@rbrunner7 could you review this?

src/p2pool/P2PoolManager.cpp
```diff
@@ -241,9 +243,14 @@ P2PoolManager::P2PoolManager(QObject *parent)
     started = false;
     // Platform dependent path to p2pool
 #ifdef Q_OS_WIN
-    m_p2poolPath = QApplication::applicationDirPath() + "/p2pool";
+    if (OSHelper().installed()) {
+        m_p2poolPath = QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation) + "/p2pool";
+    } else {
+        m_p2poolPath = QApplication::applicationDirPath() + "/p2pool";
+    }
+
```
rbrunner7: After quite some reading of the code, I am still not sure what exactly this "if" is good for: What exactly does the `installed` method check? Does this code try to keep an already existing p2pool installation in the "old" place, and if yes, why? As far as I can see, the .iss file tells the installer to delete both possible locations first on an upgrade, so is there something to keep at all?

rbrunner7 (approved):
Looks good to me.

src/p2pool/P2PoolManager.cpp
Alghwl: الغولي

PR #4581, opened 2026-04-22
DaemonManager, P2PoolManager: supervise spawned processes

thomasbuilds:
Both managers launched their child with startDetached(), monerod additionally with --detach on Unix, so the QProcess object never tracked it: the output and state connections never fired, leaving the log page console without daemon output and daemon exit undetected, and shutdown had to fall back to pkill/taskkill by image name, which also killed monerod and p2pool instances the GUI did not start.

Track the child with start() + waitForStarted() and signal the owned process on shutdown, escalating from SIGTERM to SIGKILL and giving up after 20 seconds when the daemon cannot be killed because this process does not own it. Once the RPC is down, wait for the owned process to actually exit, and refuse to start while the previous daemon is still running: destroying the tracking QProcess of a live daemon would kill it. Windows kills directly, as taskkill /F did, since console applications ignore the WM_CLOSE posted by terminate().

Drop --detach; --non-interactive already prevents stdin reads. monerod runs in its own session on Unix, so a node kept running in the background still survives the launching terminal closing, and ~DaemonManager disowns a running node instead of letting ~QProcess kill it on exit.

Cap the log page console now that daemon output reaches it, skip empty output lines, and do not mirror stdout into the GUI log; monerod writes its own log file. daemonStopped() is only emitted after the daemon actually ran.

### DIFF
--- a/src/p2pool/P2PoolManager.cpp
+++ b/src/p2pool/P2PoolManager.cpp
@@ -209,7 +209,8 @@ bool P2PoolManager::start(const QString &flags, const QString &address, const QS
     m_p2poold->setWorkingDirectory(m_p2poolPath);
 
     // Start p2pool
-    started = m_p2poold->startDetached();
+    m_p2poold->start();
+    started = m_p2poold->waitForStarted();
 
     if (!started) {
         qDebug() << "P2Pool start error: " + m_p2poold->errorString();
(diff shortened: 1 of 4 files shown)


selsta:
Based on quick testing on macOS, I did not notice any issues with this PR and starting / stopping daemon.

plowsof:
i like this so far. on debian it restarts monerod with correct p2pool args and can stop both monerod / p2pool

plowsof:
selsta asked me to confirm if we can exit the gui and allow monerod to keep running. no.
- start the daemon
- ui/logs and `ps aux | grep 'monerod'` shows its running
- close gui and select "keep it running" button
- maybe the log line is relevant `QProcess: Destroyed while process ("/monero-gui/build/bin/monerod") is still running.`
- `ps aux | grep 'monerod'` its gone

thomasbuilds:
Thanks should be fixed now

plowsof:
yep, stays running now. and the gui can still restart monerod with p2pool or close it (after gui is started after letting monerod keep running)

src/p2pool/P2PoolManager.cpp
jpk68, line 224: It looks like this function can block the GUI thread for ~7 seconds instead of returning almost instantly like it did before. IMO, it would be better to do something similar to the changes in `DaemonManager` to avoid this.

thomasbuilds:
Closing, this needs to be rewritten against Qt6 rather than adapted. The `setupChildProcess()` override here doesn't compile under Qt 6.8 minimum and re-review turned up other issues.

I also opened this from my `master` branch, which has made my fork bad to maintain. I'll resubmit later.

PR #4591, opened 2026-05-09
settings: fix pluralization

SNeedlewoods:
submitted by jpk68
https://codeberg.org/jpk68/monero-gui/pulls/2

(I have not reviewed this code, merely forwarding it)

### DIFF
--- a/pages/settings/SettingsLayout.qml
+++ b/pages/settings/SettingsLayout.qml
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2024, The Monero Project
+// Copyright (c) 2014-2026, The Monero Project
 // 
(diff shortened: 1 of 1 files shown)


selsta:
as far as I know the correct way to do this in QML is

```
text: qsTr("Every %n minute(s)", "", value) + translationManager.emptyString
```

`%n` triggers lupdate to consider this to be a plural, and then the platforms like weblate would allow to translate the different cases depending on the language, but I never tried this

selsta (approved):
Approving this but would prefer to merge it once weblate is back up.

plowsof (approved):

SChernykh (approved):

PR #4596, opened 2026-05-22
network: require SSL for HTTPS requests

selsta:


plowsof:
download of a new update works. also price api.

plowsof (approved):

PR #4597, opened 2026-05-26
QR: skip QImage padding when filling quirc buffer

selsta:
also update quirc submodule


plowsof:
looks good to me, tested with screen grab, an address + description (bunch of zeros to make the QR denser) + amount , noticed a problem with Wayland unrelated to this PR #4607

plowsof (approved):

PR #4601, opened 2026-06-03
Replace simple mode bootstrap with remote nodes

selsta:


js/RemoteNodes.js
```diff
@@ -0,0 +1,196 @@
+.pragma library
+
+var addresses = [
+    "nodes.hashvault.pro:18081",
```
jeffro256, line 4: How was this list compiled? And what would be the processes for addition and removal?

selsta: It was taken from https://monero.fail API, I'm open for different way to get a list of nodes. I've just compiled a random list now to have a working PR.

js/RemoteNodes.js
```diff
@@ -0,0 +1,196 @@
...
+    "kuk.fan",
+    "152.53.150.12:18081",
+    "xmr.diyarciftci.xyz:18081",
+    "xmr.diyarciftci.xyz:18089",
+    "79.110.49.111:18089",
+    "node.vern.cc:18081",
+    "xmr.saxemberg.com:18089",
+    "91.224.92.48:18089",
+    "82.153.138.19:18081",
+    "node.monero.mbaraa.com",
```
plowsof, line 156: i was unlucky to select this node, we need ports set to test this afaict , when in this state , there is no 'change node' button available , and its not trying another node, just an ~~infinite loop~~ *it seems like it did eventually cycle to another node, but unlucky to land on no port again xD

it did find a working node and begin syncing.

plowsof:
so far im happy with the functionality, nodes are cycled quickly, the entire list is tried (random shuffled on each launch), default port is attempted when one isnt provided, and the good node we where last connected to persists.  a good node persisting is good for ux but maybe the same node each each time allows one node to see when you are syncing / using your wallet. using a tor proxy also fine.

tiny nit: in testnet/stagenet the simple mode option is greyed out but the pointer changes as though it where clickable 

and the node list should be ip's only, [no urls](https://www.rastersec.com/blog/chainanalysis-malicious-xmr)

js/RemoteNodes.js
```diff
@@ -0,0 +1,197 @@
...
+    "xmr.r0cket.net:18081",
+    "xmr-node.ddns.net:18081",
+    "monero.definitelynotafed.com",
+    "89.234.140.131:18089",
+    "monerod.foolsump.com:18089",
+    "176.123.6.130:18081",
+    "kuk.fan:18089",
+    "139.144.169.163:18081",
+    "89.233.107.164:18081",
+    "139.59.89.227:18089",
```
plowsof, line 145: i know this is a random list, maybe a CI to ensure no ip's here are on the current ban list xD
https://github.com/Boog900/monero-ban-list/blob/22ddf399a8dc7cbfc33519b2affafaca31b2510b/ban_list.txt#L278

js/RemoteNodes.js
```diff
@@ -0,0 +1,197 @@
...
+    "xmr-au-1.cryptovps.io",
+    "xmr-full.p2pool.uk:18089",
+    "xmr-de.boldsuck.org:18081",
+    "monero.anycolo.net:18081",
+    "monero.dyni.net:18081",
+    "owl.lc:18089",
+    "monero.10z.com.ar:18089",
+    "xmr.visnova.pl",
+    "static.85.202.181.135.clients.your-server.de:18089",
+    "static.63.34.245.188.clients.your-server.de:18089",
```
plowsof, line 40: resolved to one, sneaky

https://github.com/Boog900/monero-ban-list/blob/22ddf399a8dc7cbfc33519b2affafaca31b2510b/ban_list.txt#L401

```
dig static.63.34.245.188.clients.your-server.de`
;; ANSWER SECTION:
static.63.34.245.188.clients.your-server.de. 7014 IN A 188.245.34.63
```

js/RemoteNodes.js
plowsof: for addition ideally no proxies, this could be someone wanting to host a node at home and to obscure their ip by having nginx sit between them but we can't know if its just a malicious party being a MITM. not on any official public dns/mrl ban list. running the reference release/master binaries. 

maybe require that they are also serving p2p traffic on 18080?

plowsof: edit to "no proxies" : 1 ip mirroring another monero node shouldnt be allowed, but sitting behind a reverse proxy can provide ddos protection, so that alone cant exclude nodes from the list.

PR #4609, opened 2026-06-18
TransactionHistory: prevent CSV formula injection in writeCSV

thomasbuilds:
writeCSV wrote the transaction note and subaddress label into the CSV stripping only the quote character. A cell beginning with =, +, - or @ can be interpreted as a formula by spreadsheet software on open, which CSV quoting does not prevent.

The transaction note can be attacker-controlled: a payment request's tx_description is stored as the note when the payment is sent, so a crafted note can run a spreadsheet formula when the user later exports and opens their history, potentially enabling data exfiltration or command execution.

Prefix affected fields with a single quote so they are treated as text; fields beginning with whitespace or a control character are prefixed too.

### DIFF
--- a/src/libwalletqt/TransactionHistory.cpp
+++ b/src/libwalletqt/TransactionHistory.cpp
@@ -36,6 +36,31 @@
 #include <QWriteLocker>
 #include <QtGlobal>
 
+namespace {
+    QString sanitizeCSVField(QString field)
+    {
+        field.remove(QChar('"'));
+
+        if (field.isEmpty()) {
+            return field;
+        }
+
+        const QChar first = field.at(0);
+        const bool needsPrefix =
+            first == QChar('=') ||
+            first == QChar('+') ||
+            first == QChar('-') ||
+            first == QChar('@') ||
+            first.isSpace() ||
+            first.category() == QChar::Other_Control;
+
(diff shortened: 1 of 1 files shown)


src/libwalletqt/TransactionHistory.cpp
```diff
@@ -36,6 +36,16 @@
...
+namespace {
+    QString sanitizeCSVField(QString field)
+    {
+        field.remove(QChar('"'));
+        if (!field.isEmpty() && QString("=+-@\t\r").contains(field.at(0))) {
+            field.prepend(QChar('\''));
+        }
+        return field;
+    }
+}
```
selsta, line 63:
```suggestion
namespace {
    QString sanitizeCSVField(QString field)
    {
        field.remove(QChar('"'));

        if (field.isEmpty()) {
            return field;
        }

        const QChar first = field.at(0);
        const bool needsPrefix =
            first == QChar('=') ||
            first == QChar('+') ||
            first == QChar('-') ||
            first == QChar('@') ||
            first.isSpace() ||
            first.category() == QChar::Other_Control;

        if (needsPrefix) {
            field.prepend(QChar('\''));
        }

        return field;
    }
}
```

maybe like this?

selsta (approved):

PR #4612, opened 2026-06-22
wizard: remove temporary wallet files when wallet creation is abandoned

thomasbuilds:
removeTemporaryWallet() was only called on the successful save path, leaving wallet key files behind in the temp directory whenever a wallet creation or restore was abandoned, failed, or retried.

Consolidate cleanup into closeTemporaryWallet(), which closes the wallet before removing the files (the wallet stores its cache on close, which would recreate the file) and skips deletion while createWalletFromDeviceAsync() is writing the files on another thread. It runs on wizard restart, before creating a replacement wallet, and when quitting mid-wizard; failed restore and device creation remove their files directly after closing the wallet.

Block wizard input during asynchronous device wallet creation: the processing splash does not block the wizard, so navigating back to the wizard home could delete the files while they were being created. deviceWalletCreationInProgress tracks the creation because the device passphrase dialog re-enabled the wizard when closed, and restart() keeps the walletManager connections while it is set so a late walletCreated still cleans up.

### DIFF
--- a/components/PasswordDialog.qml
+++ b/components/PasswordDialog.qml
@@ -110,7 +110,7 @@ FocusScope {
     function close() {
         leftPanel.enabled = true
         middlePanel.enabled = true
-        wizard.enabled = true
+        wizard.enabled = !wizard.deviceWalletCreationInProgress
         if (rootItem.state == "wizard") {
             titleBar.state = "essentials"
         } else {
--- a/main.qml
+++ b/main.qml
@@ -288,6 +288,7 @@ ApplicationWindow {
         if (isQuitting)
             return;
         isQuitting = true;
+        wizard.restart();
         closeWallet(function() {
             gracefulShutdownComplete();
         })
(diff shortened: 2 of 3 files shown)


jpk68 (approved):
This looks fine to me, but it might be good to have another person check as well.

selsta:
@thomasbuilds did you confirm this works correctly with `createWalletFromDeviceAsync()`? This cleanup might race with hardware wallet file creation.

thomasbuilds:
This should fix it ~~entirely~~

selsta:
1. Hardware wallet creation starts.
2. Wizard is disabled.
3. Passphrase dialog opens.
4. User submits or cancels the passphrase.
5. The dialog re-enables the wizard.
6. Hardware wallet creation may still be running.

This edge case might need a separate `deviceWalletCreationInProgress` variable.

selsta:
I have opened https://github.com/monero-project/monero-gui/pull/4654 as an alternative, as this current PR still has unresolved edge cases with async wallet cleanup + graceful exit. Fixing this would add too much code complexity, instead it's easier to keep everything in memory until final save.

thomasbuilds:
closing in favor of #4654

PR #4613, opened 2026-06-22
Dockerfiles: fetch dependencies over https instead of git:// / http://

thomasbuilds:
The release Dockerfiles cloned Qt (code.qt.io) and libgpg-error/libgcrypt (git.gnupg.org) over unauthenticated git://, and fetched libiconv (ftp.gnu.org) over http://. Qt is tag-pinned only, so a MITM on its clone could substitute source into the build. Switch them all to https, like every other dependency in these files. The gnupg commit pins and the libiconv sha256sum are unchanged.

### DIFF
--- a/Dockerfile.android
+++ b/Dockerfile.android
@@ -56,7 +56,7 @@ RUN wget -q https://github.com/madler/zlib/releases/download/v${ZLIB_VERSION}/zl
     && make -j${THREADS} install \
     && rm -rf $(pwd)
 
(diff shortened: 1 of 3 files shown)


selsta (approved):

thomasbuilds:
For a follow up PR, should we pin Qt to commit hashes or is tracking the v5.15.19-lts-lgpl tag intentional?

PR #4614, opened 2026-06-23
Dockerfiles: pin Qt to commit hashes

thomasbuilds:
Qt was cloned with -b ${QT_VERSION} --depth 1 but, unlike every other dependency in these files, never reset to a commit, so a moved upstream tag would silently change the Qt source built into the release. Pin each module to its v5.15.19-lts-lgpl tag commit with git reset --hard, matching the existing pattern (e.g. libgpg-error/libgcrypt). linux and windows pin every module explicitly; android pins the qt5 superproject that init-repository resolves the submodules from.

Follow-up to #4613.


selsta:
Is it possible to change this so only one hash needs to be hardcoded instead of each submodule?

thomasbuilds:
Yes good suggestion, implemented

selsta (approved):

PR #4616, opened 2026-06-25
linux: use screenshot portal for Wayland QR scanning

### LINKED ISSUE #4607: Grab QR from screen does not work in Wayland
on Wayland, the Transfer page's "Grab QR code from screen" button does nothing

`OSHelper::grabQrCodesFromScreen() captures via QScreen::grabWindow(0), which doesn't work under Wayland.`

i request displaying a warning under Wayland that it's not supported when the button is clicked (most likely to be accepted) or fix it under Wayland `Capture via org.freedesktop.portal.Screenshot` (which would bring a prompt up where the user is asked if they want to share the screenshot with the application (unlikely to be merged)

waozixyz:
Fixes https://github.com/monero-project/monero-gui/issues/4607

 ### What changed

  This updates `OSHelper::grabQrCodesFromScreen()` so the Transfer page's "Grab QR code from screen" action can work on Wayland.

  On Wayland, `QScreen::grabWindow(0)` does not capture the desktop, so the
  existing QR screen scan path silently failed. This PR detects Wayland on Linux
  and uses `org.freedesktop.portal.Screenshot` instead. The existing
  `QScreen::grabWindow(0)` path is still used for non-Wayland platforms.

  The portal path:
  - asks the desktop portal for a screenshot using the standard user-consent
  prompt
  - waits for the portal response
  - loads the returned local screenshot URI
  - runs the existing QR decoder on that image
  - returns no QR codes if the request is cancelled, denied, times out, or the
  portal is unavailable

  ### Why

  Wayland intentionally blocks direct screen capture APIs like
  `QScreen::grabWindow(0)`. The XDG screenshot portal is the desktop-supported
  way for applications to request a screenshot under Wayland while keeping user
  consent in the flow.

  This keeps the existing Transfer page behavior where possible, instead of
  leaving the button doing nothing on Wayland.

  ### Notes

 This adds `Qt5DBus` on Linux desktop builds so the application can communicate with the XDG desktop portal.


nahuhh:
What happened here

Edit: oh. Lots of merge commits

in future, dont `git merge` on monero-project repos. Use `git rebase`

PR #4627, opened 2026-07-05
build: switch to Ninja

selsta:
This is part of the upcoming Qt 6 migration, where switching to Ninja fixes QML cache generation on Windows.


plowsof:
didn't notice anything different under ubuntu wayland

plowsof (approved):

PR #4639, opened 2026-07-09
settings: add config page for i2p

jpk68:
Adds the long-awaited I2P config page to the GUI. Depends on [#10523](https://github.com/monero-project/monero/pull/10523).


jpk68:
Now moved from its own page to `Settings > Node`; I also removed the I2P logo and used a dropdown menu instead of radio buttons to save space.

PR #4641, opened 2026-07-09
wizard: warn when adjusting kdf rounds

### LINKED ISSUE #4076: Adjusting KDF Rounds Invalidates Current Password
After creating a wallet, if the number of KDF rounds is changed, the wallet password becomes invalid.

If this is the expected behavior, I think there should be a warning when changing it. A tooltip when the password is invalid may help too.

jpk68:
Closes #4076

### DIFF
--- a/wizard/WizardHome.qml
+++ b/wizard/WizardHome.qml
@@ -222,6 +222,8 @@ Rectangle {
                     id: kdfRoundsText
                     Layout.maximumWidth: 180
 
+                    property bool kdfWarningShown: false
+
                     labelText: qsTr("Number of KDF rounds:") + translationManager.emptyString
                     labelFontSize: 14
                     fontSize: 16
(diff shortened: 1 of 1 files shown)


wizard/WizardHome.qml
```diff
@@ -230,7 +232,28 @@ Rectangle {
...
+                        var wasDefault = persistentSettings.kdfRounds === 1;
+                        var newValue = parseInt(kdfRoundsText.text) >= 1 ? parseInt(kdfRoundsText.text) : 1;
+                        persistentSettings.kdfRounds = newValue;
+
+                        if (newValue === 1) {
+                            kdfRoundsText.kdfWarningShown = false;
+                        } else if (wasDefault && !kdfRoundsText.kdfWarningShown) {
+                            kdfRoundsText.kdfWarningShown = true;
+                            confirmationDialog.title = qsTr("Warning") + translationManager.emptyString;
+                            confirmationDialog.text = qsTr("You have selected a non-standard number of KDF rounds.\n\n"
```
selsta: Please change this to a single string literal, I don't remember if runtime expression is allowed for `qsTr()`

jpk68: Thanks, done.

selsta: You can split it up over multiple lines

```
qsTr("text"
     "text2");
```

jpk68: Would you prefer if I did this, or is it fine as is?

selsta (approved):

PR #4643, opened 2026-07-11
makefile: remove unused variable

jpk68:
Removes the unused `deldirs` variable in the `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,10 +22,6 @@ ifeq ($(USE_SINGLE_BUILDDIR), OFF)
     builddir := $(builddir)/$(branch)
     topdir := $(topdir)/..
   endif
-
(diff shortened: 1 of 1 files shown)


selsta:
This appears less relevant for monero-gui because unlike CLI it does not generate wallets in the same directory as the binary.

jpk68:
> This appears less relevant for monero-gui because unlike CLI it does not generate wallets in the same directory as the binary.

Maybe this is a dumb reason, but I just like that it's more consistent between repos this way. Otherwise, I guess the unused `deldirs` variable can be removed?

jpk68:
Done.

SNeedlewoods (approved):

selsta (approved):

PR #4644, opened 2026-07-11
components: make daemon manager dialog non-modal

jpk68:
This essentially just allows you to still interact with the rest of the GUI while the "Starting local node in 10 seconds" dialog is active. The window blocking the rest of the GUI is a bit of a pain.

### DIFF
--- a/components/DaemonManagerDialog.qml
+++ b/components/DaemonManagerDialog.qml
@@ -38,7 +38,7 @@ import "../components" as MoneroComponents
 
 Window {
     id: root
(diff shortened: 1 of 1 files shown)


SNeedlewoods:
Sorry, but I NACK.
With this patch the dialog pop-up can get hidden behind the GUI without the user noticing. If people don't have the patience to wait 10 seconds they can press "Use custom settings"->"Start daemon", IMO.

jpk68:
Looks like a decent number of people are against this change; closing.

PR #4645, opened 2026-07-11
WizardController: set password for restored temp wallet

plowsof:


selsta (approved):
Depends on https://github.com/monero-project/monero/pull/10901, so this has to be merged after a submodule update.

PR #4647, opened 2026-07-14
sync: show human-readable estimation of remaining sync time

### LINKED ISSUE #4646: Add option to show date of current block height while syncing
It may be useful to (have an option to) show the date/timestamp of the most recent block height while syncing the blockchain, similar to how the GUI in Bitcoin Core does it. For example:

- `Blocks remaining: 17,502 (24 days behind)`
- `Blocks remaining: 3,717,015 (2014-04-18)`

This is already done in info logs in `monerod`, IIRC. I guess it could either be an option or the default behaviour, depending on whether there's space (?)

jpk68:
Adds a text label to display an estimation of remaining blockchain sync time. Also formats the number of remaining blocks with comma separators.

### DIFF
--- a/js/Utils.js
+++ b/js/Utils.js
@@ -65,6 +65,10 @@ function ago(epoch) {
         return qsTr("%n day(s) ago", "0", Math.floor(delta / 86400))
 }
 
+function formatThousands(num) {
+    return num.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, "$1,");
+}
+
 function netTypeToString(){
     // 0: mainnet, 1: testnet, 2: stagenet
     var nettype = appWindow.persistentSettings.nettype;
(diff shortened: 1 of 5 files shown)


js/Utils.js
```diff
@@ -65,6 +65,10 @@ function ago(epoch) {
         return qsTr("%n day(s) ago", "0", Math.floor(delta / 86400))
 }
 
+function formatThousands(num) {
+    return num.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, "$1,");
```
SChernykh: `num.toLocaleString("en-US");` should do the same without a regex and it's easier to read. But it's not critical.

jpk68: Thanks, fixed.

selsta:
There appears to be a mismatch between what the PR implements and what it claims to resolve.

PR #4648, opened 2026-07-14
qr: add missing null check

jpk68:


jpk68:
Upon discussing this with @selsta, it seems mostly unnecessary; closing.

PR #4649, opened 2026-07-14
Fix precision loss when passing amounts from QML

selsta:


jpk68 (approved):

plowsof:
i was too slow, LGTM : creating a payment request (and in merchant mode) keeps the precision monero:address?tx_amount=10000000.000000000111

PR #4651, opened 2026-07-14
main: add system tray icon

### LINKED ISSUE #2038: [UX] Add a tray icon for the GUI
On Windows-64 there is no tray icon in the system tray for the GUI. Currently if I completely close the GUI I can’t easily open the program again, I have to go into the file I downloaded and launch it from the .exe file.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

### LINKED ISSUE #3316: Feature Request:  minimize to system tray on Windows.
I didn't see this suggested anywhere, so here goes.
When running on Windows in GUI mode, it would be nice if the application could be banished to the system tray.  This gets it out of the taskbar and makes it a little less intrusive. Out of sight, out of mind and all that.

Since the application is meant to run in the background for long periods of time, this makes sense to me.

### LINKED ISSUE #4205: Add the minimize to Tray setting.
When the wallet is launched and syncs with the network, it takes up space on the taskbar. Is it possible to add a setting so that the program is minimized to the tray?

jpk68:
Adds a tray icon that can be used to (un)minimize the main window. Also includes a `Quit` option to close the app from the tray.

Resolves #4205

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -33,6 +33,7 @@ import QtQuick.Controls 1.1
 import QtQuick.Controls.Styles 1.1
 import QtQuick.Dialogs 1.2
 import QtGraphicalEffects 1.0
+import Qt.labs.platform 1.1 as PlatformLabs
 
 import FontAwesome 1.0
 
@@ -1511,6 +1512,7 @@ ApplicationWindow {
         property int lockOnUserInActivityInterval: 10  // minutes
         property bool blackTheme: MoneroComponents.Style.blackTheme
         property bool checkForUpdates: true
+        property bool trayIconEnabled: false
         property bool autosave: true
         property int autosaveMinutes: 10
         property bool pruneBlockchain: false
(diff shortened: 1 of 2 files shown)


selsta:
I don't think this solves all the linked issues, the common request is a system tray for the daemon running in the background when the GUI quit, which as far as I can tell is not solved by this PR, but I did not compile it yet.

jpk68:
> I don't think this solves all the linked issues, the common request is a system tray for the daemon running in the background when the GUI quit, which as far as I can tell is not solved by this PR, but I did not compile it yet.

Apologies, I should probably remove some of the linked issues. AFAIK it does not solve that problem.

selsta:
It also does not seem to solve the other issues since the common request is for the app to basically hide from the task bar, which is not what minimized does. This might be solvable with setting it to `Window.Hidden`, but I did not try this yet.

https://doc.qt.io/qt-6/qwindow.html#Visibility-enum

jpk68:
> It also does not seem to solve the other issues since the common request is for the app to basically hide from the task bar, which is not what minimized does. This might be solvable with setting it to `Window.Hidden`, but I did not try this yet.
> 
> https://doc.qt.io/qt-6/qwindow.html#Visibility-enum

Fixed; this should work now.

plowsof:
is this for windows only?

jpk68:
> is this for windows only?

No, it should be cross-platform. It works as expected on KDE as well.

main.qml
```diff
@@ -2498,4 +2500,40 @@ ApplicationWindow {
...
+
+        onVisibleChanged: {
+            if (!visible && appWindow.visibility === Window.Hidden) {
+                appWindow.visibility = restoredVisibility;
+                appWindow.raise();
+                appWindow.requestActivate();
+            }
+        }
+
+        onActivated: {
```
selsta, line 2537: On macOS, this appears broken UI wise. A separate `PlatformLabs.MenuItem` for Hide / Show is preferred.

main.qml
```diff
@@ -2498,4 +2500,56 @@ ApplicationWindow {
...
+
+        menu: PlatformLabs.Menu {
+            PlatformLabs.MenuItem {
+                text: appWindow.visibility === Window.Hidden ? qsTr("Show") : qsTr("Hide")
+                onTriggered: trayIcon.toggleWindow()
+            }
+
+            PlatformLabs.MenuItem {
+                text: qsTr("Quit")
+                onTriggered: Qt.quit()
```
selsta:
```suggestion
                onTriggered: closeAccepted()
```

main.qml
jpk68: Fixed (I think).

main.qml
jpk68: Done.

PR #4657, opened 2026-07-16
wizard: only validate custom wallet locations

selsta:
Also show the wallet location selector in all wallet modes and use the selected path when creating a wallet.


plowsof:
dir selector shown in simple mode now, good, justified as an easy fix to display the error. 

some notes as i run through this PR. im just worried about the default dir not being writeable scenario, but this is how it plays out:
default dir not writeable:
- simple mode > accept disclaimer > create new wallet -> next -> seed > seed confirm > password >.create wallet 'failed to store wallet'

re-reading https://github.com/monero-project/monero/issues/8429 
"if you specify a path outside of the allowed directory" which is a _custom dir_ - the original checks run on it . i like this PR so far. the checks run on the custom dir

plowsof (approved):

PR #4659, opened 2026-07-17
Migrate to Qt6

### LINKED ISSUE #3257: Possible conflict between "Inno Setup: BlockchainDir" setting and main.qml: MoneroSettings { property string blockchainDataDir: "" }
When I originally installed Monero GUI Wallet v0.14.1.0, I specified the BlockchainDir in the Inno Setup to "C:\ProgramData\Cryptocurrencies\Monero", rather than the default "C:\ProgramData\bitmonero", but after upgrading to Monero v0.17.1.5 and following a power failure - #3256 - the [HKCU\SOFTWARE\monero-project\monero-core] registries were re-initialised to the same configuration as MoneroSettings, including blockchainDataDir and wallet_path.

The Monero GUI Wallet should not have the ability to automatically and silently reinitialise the blockchainDataDir or any other monerod configurations (daemon{Flags,Password,Username}), especially if settings were configured by the installer and are present under [HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Monero GUI Wallet_is1].
Monero GUI Wallet should configure monerod via an API - monero-project/monero#7083

Such re-configuration of monerod poses a serious security threat on Multi-user workstations - monero-project/monero#7079.

I noticed that a change to the PersistantSettings for languages was implemented between v0.17.1.3 and v0.17.1.5, which if re-initialisation of MoneroSettings were triggered, I assume would cause all of the existing PersistentSettings to reset, including account_name, blockchainDataDir and wallet_path - https://github.com/monero-project/monero-gui/compare/v0.17.1.3...master

Comparison of MoneroSettings and current monero-core registries:
MoneroSettings:
```
        property string language: 'English (US)'
        property string language_wallet: 'English'
        property string locale: 'en_US'
        property string account_name
        property string wallet_path
        property bool   allow_background_mining : false
        property bool   miningIgnoreBattery : true
        property var    nettype: NetworkType.MAINNET
        property int    restore_height : 0
        property bool   is_trusted_daemon : false
        property bool   is_recovering : false
        property bool   is_recovering_from_device : false
        property bool   customDecorations : true
        property string daemonFlags
        property int logLevel: 0
        property string logCategories: ""
        property string daemonUsername: ""
        property string daemonPassword: ""
        property bool transferShowAdvanced: false
        property bool receiveShowAdvanced: false
        property bool historyShowAdvanced: false
        property bool historyHumanDates: true
        property string blockchainDataDir: ""
        property bool useRemoteNode: false
        property string remoteNodeAddress: ""
        property string bootstrapNodeAddress: ""
        property bool segregatePreForkOutputs: true
        property bool keyReuseMitigation2: true
        property int segregationHeight: 0
        property int kdfRounds: 1
        property bool hideBalance: false
        property bool askPasswordBeforeSending: true
        property bool lockOnUserInActivity: true
        property int walletMode: 2
        property int lockOnUserInActivityInterval: 10  // minutes
        property bool blackTheme: true
        property bool checkForUpdates: true
        property bool autosave: true
        property int autosaveMinutes: 10

        property bool fiatPriceEnabled: false
        property bool fiatPriceToggle: false
        property string fiatPriceProvider: "kraken"
        property string fiatPriceCurrency: "xmrusd"

        property string proxyAddress: "127.0.0.1:9050"
        property bool proxyEnabled: isTails
```

Registry:
```
[HKEY_CURRENT_USER\SOFTWARE\monero-project\monero-core]
"hideBalance"="false"
"daemonPassword"=""
"remoteNodeAddress"=""
"auto_donations_enabled"="false"
"fiatPriceProvider"="kraken"
"nettype"=dword:00000000
"bootstrapNodeAddress"=""
"historyHumanDates"="true"
"kdfRounds"=dword:00000001
"receiveShowAdvanced"="false"
"is_trusted_daemon"="false"
"transferShowAdvanced"="false"
"blackTheme"="true"
"is_recovering_from_device"="false"
"lockOnUserInActivityInterval"=dword:0000000a
"segregationHeight"=dword:00000000
"logCategories"=""
"lockOnUserInActivity"="true"
"daemonFlags"=""
"is_recovering"="false"
"allow_background_mining"="false"
"walletMode"=dword:00000002
"fiatPriceCurrency"="xmrusd"
"historyShowAdvanced"="false"
"auto_donations_amount"=dword:00000032
"language"="English (US)"
"payment_id"=""
"customDecorations"="true"
"fiatPriceEnabled"="false"
"remoteNodeRegion"=""
"logLevel"=dword:00000000
"daemonUsername"=""
"showPid"="false"
"keyReuseMitigation2"="true"
"useRemoteNode"="false"
"segregatePreForkOutputs"="true"
"miningIgnoreBattery"="true"
"remoteNodeService"=""
"blockchainDataDir"=""
"fiatPriceToggle"="false"
"account_name"=""
"locale"="en_US"
"wallet_path"=""
"restore_height"=dword:00000000
"askPasswordBeforeSending"="true"
"checkForUpdates"="true"
"autosave"="true"
"autosaveMinutes"=dword:0000000a
"proxyAddress"="127.0.0.1:9050"
"proxyEnabled"="false"
"language_wallet"="English"

[HKEY_CURRENT_USER\SOFTWARE\monero-project\monero-core\QQControlsFileDialog]
"sidebarSplit"="111.60000000000001"
"favoriteFolders"=hex(7):00,00
"sidebarWidth"="80"
"width"=dword:00000000
"sidebarVisible"="false"
"height"=dword:00000000
```

### LINKED ISSUE #3948: Monero GUI forgets some of its configuration
Monero GUI v0.17.3.2 today greeted me with a dark theme (ouch!) and started to sync at block 0 (double ouch!).
I have used 0.17.3.2 before, so there shouldn't be any changes upon a restart.
It seems to have forgotten some configuration settings, but not all. The wallet location was still correct and it was properly opened. The blockchain location was reset to default, where it couldn't be found, obviously. I don't know if/what other settings were affected.
There's plenty of space left on the drive, so this shouldn't be the cause as suggested in issue #3929.
OS is Debian 11.

### LINKED ISSUE #4348: Migrate to Qt 6
Qt 5.15 LTS commercial extended lifetime ends in [May 2025](https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders). Due to an agreement with KDE, commercial releases become available under LGPL with a one year delay. However, we can't expect patches for critical issues or support for new platforms after official support ends.

Qt 6.8 is the latest LTS version, set for release in September 2024.

The required build system changes (CMake / Dockerfiles) are relatively straightforward. It's unclear to me at this time how much code needs to be modified to be compatible with Qt 6, and how much breakage there will be in the QML / graphical department.

### LINKED ISSUE #4712: Gui freezes after RDP disconnect
I'm using Monero Gui to mine on a remote machine that I connect to via RDP.
When I close the RDP session and reconnect it, the GUI is frozen, it does not accept any input and can't be closed.
Mining apparently continues in the background, at least according to CPU usage.
When I open a new Gui in parallel and close it again and accept to forcibly stop the node, the old Gui instance can be closed once the node has been stopped, but nothing else can be clicked.

OS: Windows 10
Gui version: Fluorine Fermi 5.2

selsta:
This PR migrates Monero GUI from Qt 5 to Qt 6, including the necessary QML, C++, CMake, deployment, CI, and packaging changes.

@Malinero started this effort and provided much of the foundational work that made this migration possible.

The migration is kept as one large commit because its individual parts are tightly coupled and intermediate commits would not compile. Splitting it further would therefore create an artificial history rather than a useful sequence of working changes.

Everyone who commited code changes to this PR is listed as an author, and also thank you to @jpk68 and RavFX for helping with testing.

Closes #4348
Resolves #3257
Resolves #3948
Resolves #4712


pages/settings/SettingsInfo.qml
```diff
@@ -41,6 +41,28 @@ Rectangle {
...
+        case GraphicsInfo.OpenVG:
+            return "OpenVG"
+        case GraphicsInfo.Software:
+            return "Software"
+        case GraphicsInfo.Null:
+            return "Null"
+        default:
+            return "Unknown"
+        }
+    }
```
jpk68: Something like this looks cleaner and is a lot more concise.

```suggestion
    readonly property string rendererName: ({
        [GraphicsInfo.OpenGL]: "OpenGL",
        [GraphicsInfo.Vulkan]: "Vulkan",
        [GraphicsInfo.Metal]: "Metal",
        [GraphicsInfo.Direct3D11]: "Direct3D 11",
        [GraphicsInfo.Direct3D12]: "Direct3D 12",
        [GraphicsInfo.OpenVG]: "OpenVG",
        [GraphicsInfo.Software]: "Software",
        [GraphicsInfo.Null]: "Null"
    })[GraphicsInfo.api] ?? "Unknown"
```

wizard/WizardLanguage.qml
```diff
@@ -238,7 +238,4 @@ Rectangle {
         id: versionTimer
     }
```
jpk68: The blank line here can likely be removed

jpk68 (changes requested):

cmake/FixupMacBundle.cmake
jpk68, line 1: Is this file ever included/used anywhere?

components/StandardDialog.qml
```diff
@@ -50,8 +47,6 @@ Rectangle {
     property alias cancelText: cancelButton.text
     property alias closeVisible: closeButton.visible
 
-    property var icon
```
jpk68, line 53: Despite this being removed, it looks like the property is still called in a few other places (such as [here](https://github.com/selsta/monero-gui/blob/dfd2f1bfe9c5001d6a99ef356720d6ebec949a0c/pages/AddressBook.qml#L553)).

components/IconButton.qml
jpk68: This line produces the following warning message:

`qrc:/components/IconButton.qml:65:9 Parameter "mouse" is not declared. Injection of parameters into signal handlers is deprecated. Use JavaScript functions with formal parameters instead.`

I'm pretty sure this can be changed to `onClicked: (mouse) => button.clicked(mouse)`. A search using `grep` suggests a similar change is also needed in a few other locations.

wizard/WizardWalletInput.qml
jpk68: It looks like calling `.folder` in Qt6 will throw an `AttributeError`.

https://doc.qt.io/qt-6/qfiledialog.html

pages/Sign.qml
```diff
@@ -61,12 +59,12 @@ Rectangle {
         if (result) {
             signatureVerificationMessage.title = qsTr("Good signature") + translationManager.emptyString
             signatureVerificationMessage.text  = qsTr("This is a good signature") + translationManager.emptyString
-            signatureVerificationMessage.icon = StandardIcon.Information
+            signatureVerificationMessage.type = MessageDialog.Information
```
jpk68: According to a quick web search, it looks like `.type` is deprecated as well (unless Google's AI is hallucinating).

src/main/main.cpp
```diff
@@ -227,17 +192,6 @@ int main(int argc, char *argv[])
 
     MainApp app(argc, argv);
 
-#if defined(Q_OS_WIN)
```
jpk68, line 230: As this is now removed, does that mean the GUI will always try to use native graphics on Windows, unless it's explicitly specified otherwise? As in, will it still try to use native rendering even when it can't/shouldn't, so long as the user hasn't forced software rendering?

pages/settings/SettingsInfo.qml
thomasbuilds: yours is shorter for sure, the OG one is easier to read though

main.qml
thomasbuilds:
```suggestion
if(!saveTxDialog.selectedFile) {
currentWallet.disposeTransaction(transaction)
return;
}

var path = walletManager.urlToLocalPath(saveTxDialog.selectedFile)
```

Qt6 `FileDialog` has no `fileUrl` property, it was renamed to `selectedFile`, it was missed here.
Reading a nonexistent property returns `undefined`, so `!saveTxDialog.fileUrl` is always true: when a view-only wallet confirms a transaction and picks a save location, the unsigned transaction is silently disposed instead of written to file.

thomasbuilds (changes requested):

pages/settings/SettingsInfo.qml
selsta: Updated to the more concise version, it is a bit more difficult to read but this shouldn't be code that needs change anytime soon.

src/main/main.cpp
selsta: I updated the code to correctly detect software renderer.

> does that mean the GUI will always try to use native graphics on Windows, unless it's explicitly specified otherwise?

As far as I understand it, yes. We can't do the OpenGL trick anymore to detect fallback.

selsta:
Force pushed to implemented all the suggested changes. @jpk68 @thomasbuilds

thomasbuilds:
QR scanner is still Qt5?

selsta:
Likely yes, I did not compile with SCANNER enabled, will put it on my to-do.

jpk68:
The code looks good to me, however I have not yet gone through Docker-related files. I'll wait for the QR code thing to be resolved before approving.

thomasbuilds:
List of things my AI said (not verified - cherry pick if you want)
- Restore the sacrificial SOCKS5 proxy line in NetworkAccessBlockingFactory (defense-in-depth; the main block is intact)
- ProcessingSplash animator: running: parent.visible on the RotationAnimator resolves through the document root (Animators have no parent), so it spins forever even when hidden
- GradientBackground strokeWidth: 0 → -1 (0 doesn't disable the stroke; default stroke color is white)
- Leftover console.log in LanguageSidebar (lines 88, 177)
- PortableSettings::setPortable copies from the file, racing the QML Settings ~500 ms flush; portableFilePath() is CWD-relative and uncached
- CI: two jobs still on checkout@v1, upload-artifact name: dead under archive: false, ccache installed but never cached, NDK pinned by SHA-1
- Cosmetic: white-theme close/minimize SVGs are dark so colorization can't tint them, InlineButton's white-theme shadow was dropped, MoneroCalendar hover overrides the selected-day highlight, QtQuickControls1 import alias is misleading

src/qt/PortableSettings.h
```diff
@@ -32,70 +26,43 @@
...
-#include <QDebug>
-#include <qsettings.h>
+#include <QString>
+#include <QUrl>
 
-static const int settingsWriteDelay = 500; // ms
+class QSettings;
 
-class MoneroSettings : public QObject, public QQmlParserStatus
+class PortableSettings : public QObject
```
jpk68, line 37: Does this new settings class still guarantee that disk writes won't occur in portable mode?

src/qt/PortableSettings.cpp
```diff
@@ -0,0 +1,125 @@
...
+    QSettings source = makeSettings(m_portable);
+    QSettings destination = makeSettings(enabled);
+
+    source.sync();
+    if (source.status() != QSettings::NoError)
+        return false;
+
+    // The destination should be an exact copy. Keeping keys that no longer
+    // exist in the source can otherwise resurrect stale settings.
+    destination.clear();
```
jpk68: It looks like this clears the string before confirming that `sync()` succeeds. If the sync fails, would this not wipe the destination's prior settings?

src/qt/PortableSettings.h
selsta: Portable mode is so that the data is stored in the same directory as the binary. It still writes to disk.

src/qt/PortableSettings.cpp
selsta: Latest version contains a more secure design.

selsta:
Addressed most of the points mentioned above. Also implemented SCANNER support, was only able to test it on macOS.

Tomorrow I will try to extract a couple more changes that are not strictly Qt6 related as separate PRs.

> Restore the sacrificial SOCKS5 proxy line in NetworkAccessBlockingFactory (defense-in-depth; the main block is intact)

This was intentional to reduce the amount of dependencies we need when compiling Qt.

thomasbuilds (approved):

jpk68 (approved):
Looks good to me; again, I have not tested Docker-related "code" or the QR scanner functionality.

jpk68 (approved):

selsta:
I will have to do one more force push in the next days to make this Qt 6.8.4 ready (Qt does not offer pre-built binaries anymore due to their LTS licensing agreement with KDE), so it will require changes to CI and Docker build scripts.

selsta:
Updated Docker to `6.8.4`. This should soon be merge ready unless we receive new bug reports.

jpk68 (approved):

selsta:
Rebased, fixed issues brought up by @thomasbuilds LLM review, enabled unity builds to speed up CI, added more tests and various other small improvements.

thomasbuilds (approved):

jpk68 (approved):

jpk68 (approved):

thomasbuilds (approved):

PR #4660, opened 2026-07-19
wizard: fix restore from QR code

thomasbuilds:
updateFromQrCode lived in Wizard.js, a .pragma library script with no access to QML scope, so every scan aborted on its first statement and filled nothing. Broken since the wizard redesign (f329a710), the function also still referenced pre-redesign items that no longer exist.

Move it into WizardRestoreWallet1.qml next to the fields it fills, switch to the keys form so the scanned values are visible, and handle the null extra_parameters emitted for bare-address QR codes.


selsta (approved):
I'll open follow up PRs to get restore from QR fully working.

PR #4661, opened 2026-07-19
workflows: update checkout and upload-artifact to v7

selsta:


thomasbuilds (approved):

.github/workflows/build.yml
plowsof: another here

selsta: Fixed.

plowsof (approved):

PR #4663, opened 2026-07-19
oshelper: wait for hidden window before taking screenshot on macOS

selsta:
Workaround for macOS. In the future I'll switch to ScreenCaptureKit but that will require larger changes.

### DIFF
--- a/src/main/oshelper.cpp
+++ b/src/main/oshelper.cpp
@@ -28,6 +28,7 @@
 
 #include "oshelper.h"
 
+#include <algorithm>
 #include <unordered_set>
 
 #include <QCoreApplication>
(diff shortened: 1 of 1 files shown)


src/main/oshelper.cpp
```diff
@@ -144,7 +155,7 @@ QList<QString> OSHelper::grabQrCodesFromScreen() const
     }
     catch (const std::exception &e)
     {
-        qWarning() << e.what();
+        qWarning() << "QR screen capture: decoder exception:" << e.what();
```
jpk68, line 158: Nit: spacing before error text

selsta: https://doc.qt.io/qt-6/qdebug.html#formatting-options

spaces get added automatically

jpk68 (approved):

PR #4667, opened 2026-07-21
qml: improve performance with many subaddresses

selsta:
Depends on #4659.

Supersedes #4501 with a virtualized `ListView` implementation. Thanks to @sneurlax for identifying, reproducing, and investigating the performance issue.


sneurlax:
ACK & LGTM, thanks for tackling this in a better way, selsta!

PR #4669, opened 2026-07-30
chore: update zxcvbn-c

jpk68:
Updates `zxcvbn-c`, used for estimating password strength. It appears to have accumulated numerous bugfixes over the past few years.

An included header seemingly meant for MSVC was removed from the updated code to fix the build.


jpk68:
Updated to fix the constant defined in `CMakeLists.txt`, since I was looking at the wrong number before (oops). It would not have caused any memory-safety issues regardless.

selsta (approved):

PR #4670, opened 2026-08-03
settings: add System/Light/Dark theme selector

### LINKED ISSUE #3660: "Use System Theme"
currently we have a "Light theme" on/off toggle. 

I suggest we make "Theme" a drop-down or a segmented control with 3 options: 

1. Light theme 
2. Dark theme
3. System theme 

Thanks!

atraxsrc:
Closes #3660.

Replaces the "Light theme" checkbox in Settings → Interface with a System / Light / Dark selector. When set to System, the wallet follows the desktop's colour-scheme preference and updates live, without a restart.

### Approach

`Style.blackTheme` becomes a read-only property derived from a new `Style.theme` tri-state (`"system"`, `"light"`, `"dark"`). All 34 files that read `Style.blackTheme` are untouched — only the four sites that wrote to it are updated. Making the derived property `readonly` means any future write is a load-time error rather than a silently broken binding.

System detection is a small `ColorScheme` QObject exposed to QML:

- **Linux**: reads `org.freedesktop.appearance color-scheme` from  xdg-desktop-portal and subscribes to `SettingChanged` for live updates. 
- **Fallback**: compares the `Window` and `WindowText` lightness of the  application palette, which covers macOS and Linux setups with a Qt platform  theme plugin.
- **Windows**: Qt 5 does not reflect the preference in the palette, so it reads  `AppsUseLightTheme` from the registry.

`Qt5DBus` is an **optional** build dependency — when it is not found, the palette fallback is used and nothing else changes. This avoids touching the depends/Guix setup.

The portal path exists because the palette heuristic is not sufficient on its own: on desktops that do not theme Qt applications (COSMIC, and any setup where the Qt palette comes from a static qt5ct profile), the palette never reflects the user's actual choice.

### Migration

The existing `blackTheme` setting is migrated to the new `theme` key on first run, and the legacy key is kept in sync afterwards so downgrading to an earlier release preserves the user's choice. The default for new profiles is unchanged (dark).

### Testing

Pop!_OS 24.04, COSMIC, Wayland, Qt 5.15.13:

- Dark theme is visually unchanged from before.
- Selector persists across restarts; light and dark both apply correctly.
- With System selected, toggling COSMIC → Appearance switches the wallet live  in both directions (screenshots below).
- Migration from a config containing only `blackTheme=false` resolves to  `theme=light`, verified with the migration path instrumented.

Not verified: the migrated value being written to disk on the first upgrade run specifically. Settings writes are disabled on the wizard path (`WizardLanguage.qml` calls `setWritable(false)`), so this needs a profile with
an existing wallet to exercise end to end.

macOS and Windows are untested — I don't have access to either.

### Screenshots
dark
light
system toggle dark
system toggle light

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,19 @@ if(Qt5QmlModels_PKG_CONFIG_FOUND)
   list(APPEND QT5_LIBRARIES Qt5QmlModels)
 endif()
 
+# Optional. Lets the "System" theme setting read the desktop colour-scheme preference
+# from xdg-desktop-portal. Without it we fall back to inspecting the Qt palette.
+if(UNIX AND NOT APPLE AND NOT ANDROID)
+  find_package(Qt5DBus ${QT_MIN_VERSION} QUIET)
+  if(Qt5DBus_FOUND)
+    message(STATUS "Found Qt5DBus, system colour-scheme detection will use xdg-desktop-portal")
+    list(APPEND QT5_LIBRARIES Qt5DBus)
+    add_definitions(-DHAVE_QT_DBUS)
+  else()
+    message(STATUS "Qt5DBus not found, system colour-scheme detection will use the Qt palette")
+  endif()
+endif()
+
 foreach(QT5_MODULE ${QT5_LIBRARIES})
   find_package(${QT5_MODULE} ${QT_MIN_VERSION} REQUIRED)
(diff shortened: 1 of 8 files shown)


selsta:
Thank you for the PR, due to the use of dbus I will wait until the Qt6 migration PR is merged before taking a look at this.

PR #4677, opened 2026-08-05
security: remove sensitive data from logs and escape HTML in tx details

Ismail-Benali:
## Summary

Two low-severity security hardening changes in the QML UI.

### 1. Stop logging sensitive wallet data
Several `console.log` calls wrote transaction IDs, wallet addresses, payment IDs and payment-proof signatures to the debug output:
- `pages/History.qml`: proof generation logged txid + address
- `pages/TxKey.qml`: proof generation and verification logged txid + address + signature
- `pages/AddressBook.qml`: sending logged address + paymentId
- `main.qml`: failed wallet open logged wallet.address

A payment-proof signature is a signed proof of ownership; writing it to the log alongside the txid weakens the proof's usefulness if the log ever leaks.

### 2. Escape HTML in the transaction details dialog
`buildTxDetailsString` in `pages/History.qml` builds a rich-text table (`StandardDialog` uses `TextEdit.AutoText`) with values that originate from the user or the chain. Only tx_note was escaped; the remaining fields (tx_id, address, paymentId, integratedAddress, tx_key, rings, destinations) were inserted raw. All fields are now passed through `Utils.htmlEscape()`. The intentional `<br>` line separators inside the destinations field are restored after escaping so the display is unchanged.

## Files changed
- main.qml
- pages/AddressBook.qml
- pages/History.qml
- pages/TxKey.qml

## Testing
Not built; changes are limited to removing log lines and HTML-escaping string interpolation. No functional behavior is intended to change.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -545,7 +545,6 @@ ApplicationWindow {
                         errorMessage += "\n" + qsTr("Make sure your Trezor is connected and unlocked.");
                     }
                     passwordDialog.showError(errorMessage);
-                    console.log("closing wallet async : " + wallet.address)
                     closeWallet();
                     return;
             }
@@ -1075,10 +1074,8 @@ ApplicationWindow {
             var result = currentWallet.getReserveProof(false, currentWallet.currentSubaddressAccount, amount, message)
             txProofComputed(null, result)
         } else {
-            console.log("Getting payment proof: ")
-            console.log("\ttxid: ", txid,
-                        ", address: ", address,
-                        ", message: ", message);
+            // Don't log txid/address/message; payment-proof signatures are
+            // sensitive and must not end up in the logs.
             function spendProofFallback(txid, result){
                 if (!result || result.indexOf("error|") === 0) {
                     currentWallet.getSpendProofAsync(txid, message, txProofComputed);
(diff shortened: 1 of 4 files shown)


nahuhh:
Not 100% sure (worth checking), but i think active wallet primary address is logged on the default lvl (from wallet2) when you open the wallet, so removing when you close the wallet wouldnt help there.

the logging of the address is also normal in other programs, such as monero-wallet-rpc

jpk68:
Yes, there is still some logging of sensitive data, namely [here](https://github.com/monero-project/monero-gui/blob/master/main.qml#L1073-L1075) and [here](https://github.com/monero-project/monero-gui/blob/master/main.qml#L1105-L1108).

Ismail-Benali:
Thanks for the review @jpk68 and @nahuhh.

@jpk68: you're right — I had missed those two spots. I've now removed the console.log calls in handleGetProof and handleCheckProof (txid, address, message and the payment-proof signature) in commit 5efbc6d, so the same class of sensitive data is no longer written to the logs there either.

@nahuhh: good point about wallet2 logging the primary address at the default level on open. That logging lives in the monero-core submodule (wallet2.cpp), which is outside this repository, so it can't be changed in this PR. Our changes here only remove the additional GUI-side logging of addresses/txids/signatures on top of what the core already emits.

jpk68:
Would be great if you could squash these commits, and maybe remove the commit description :)

Ismail-Benali:
Done — squashed into a single commit and removed the commit description. Thanks!

jpk68 (approved):

PR #4679, opened 2026-08-10
cmake: add option to skip compilation of optional monero binaries

jpk68:
This speeds up build times by a decent amount when compiling the GUI locally.


selsta:
Should be default behavior in my opinion, maybe you can change the option name to BUILD_EXTRAS or something with OFF as default?

jpk68:
Updated to have the option on by default.

selsta (approved):

PR #4680, opened 2026-08-22
Wizard/Settings: Add password strength bar to password changing interface

EntropyHoover:
This PR intends to add a password strength bar when user is changing wallet password in settings. The following work is done in this PR:

1.Move the bar implementation out of WizardAskPassword.qml, making it an independent component
2.Include the password strength bar component in PasswordDialog.qml, making the bar interface appear when user is changing password.
3.Refactor WizardAskPassword.qml to let it use the new bar component instead, in the same time keeping its public interface unchanged.
4.Move the translation assets related to password strength from WizardAskPassword to PasswordStrengthBar

### DIFF
--- a/components/PasswordStrengthBar.qml
+++ b/components/PasswordStrengthBar.qml
@@ -1,0 +1,123 @@
+// Copyright (c) 2026, The Monero Project
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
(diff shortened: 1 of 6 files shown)


nahuhh:
Lol?. Oh. Probably whitespace changes in the translation files?

edit, nvm, its all the moves

jpk68:
Hey, thanks for working on this. I think it would be better to leave the translation files untouched, as they're machine-generated by Weblate (which our instance of is down right now). This would also reduce the overall diff by quite a bit.

selsta:
Yes, please undo the translation file changes.

EntropyHoover:
@selsta The translation files are reverted. The pr should be clean now

components/PasswordStrengthBar.qml
```diff
@@ -0,0 +1,123 @@
...
+        } else if (strengthLevel === 1) {
+            return MoneroComponents.Style.blackTheme ? "#FFFF00" : "#FFCC00";
+        }
+        return MoneroComponents.Style.blackTheme ? "#00FF00" : "#008000";
+    }
+
+    visible: passwordStrengthAvailable
+    spacing: 0
+
+    onPasswordChanged: calcPasswordStrength()
```
jpk68, line 62: It looks like the password strength is now calculated twice per keystroke, once by the component's `onPasswordChanged`, and once by the pre-existing `btnNext.enabled`. This should probably be changed so it's done only once.

components/PasswordStrengthBar.qml
```diff
@@ -0,0 +1,123 @@
...
+
+        // getPasswordStrength returns a value from 0 to... lots
+        var strength = walletManager.getPasswordStrength(password);
+        // consider anything below 10 bits as dire
+        strength -= 10;
+        if (strength < 0) {
+            strength = 0;
+        }
+        // use a slight parabola to discourage short passwords
+        strength = strength ^ 1.2 / 3;
```
jpk68: This seems to be a no-op due to operator precedence (`^` binds looser than `/`, so it's `strength ^ 0.4` and then `int32` cast to `0`). This does appear to be a pre-existing bug, but should probably be fixed regardless, especially considering this component is now used in two places.

jpk68 (commented):
Just a few nitpicks listed below - it would be good to squash the commits before merging as well :)

EntropyHoover:
Both issues are now fixed, and commits are squashed.

jpk68 (approved):

PR #4685, opened 2026-08-30
qml: fix broken leading-zero stripping in amount fields

munzzyy:
The leading-zero branch in the amount input handlers declares:

```js
if (match) {
    const cursorPosition = cursorPosition;
    text = match[1];
    cursorPosition = Math.max(cursorPosition, 1) - 1;
}
```

The local `const` reuses the property name `cursorPosition` for the new
binding and for its own initializer. Under let/const scoping the
right-hand side resolves to the not-yet-initialized local binding
instead of the property on the item. The declaration throws. The block
never gets past that line. `text = match[1]` and the cursor position
update never run.

In practice this means typing a redundant leading zero followed by a
digit (007, 05, ...) into any of these fields does not get corrected.
The same exception fires again on every keystroke after that, as long
as the leading zero is still there.

Same block, same bug, in four places: the send amount field in
Transfer, both payment request amount fields in Receive, and the
reserve proof amount field in TxKey.

Fix: rename the local variable so it stops shadowing the property. The
rest of the logic is unchanged.

I have no build of the GUI to click through the actual pages. So I
isolated the exact code from the unpatched and patched handlers into a
standalone TestCase, not part of this diff, just local verification,
and ran it with qmltestrunner-qt5 against Qt 5.15.19:

```
QDEBUG : ...test_buggy_throws_on_leading_zero() qml: buggy() threw: true (ReferenceError: cursorPosition is not defined)
PASS   : ...test_buggy_throws_on_leading_zero()
PASS   : ...test_fixed_strips_leading_zero()
Totals: 4 passed, 0 failed, 0 skipped, 0 blacklisted, 11ms
```

Same result under Qt 6.11.1 qmltestrunner. qmllint is clean on all
three modified files before and after this change. Same output, none,
both times.

### DIFF
--- a/pages/TxKey.qml
+++ b/pages/TxKey.qml
@@ -35,6 +35,7 @@ import "../components" as MoneroComponents
 import moneroComponents.Clipboard 1.0
 
 import "../js/TxUtils.js" as TxUtils
+import "../js/Utils.js" as Utils
 
 Rectangle {
     color: "transparent"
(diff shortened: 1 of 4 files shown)


jpk68 (approved):
LGTM, though at some point it may be worth refactoring this helper into `js/Utils.js`.

munzzyy:
Makes sense. I kept this one to the minimal in-place fix so it stays easy to review and backport, but moving the leading-zero handling into a shared helper in js/Utils.js would be a clean follow-up. I can do that as a separate PR, or fold it in here if you would rather keep it to one change.

jpk68:
Your choice, but doing it now would likely be more convenient ;)

munzzyy:
Folded it in. All four spots now call a stripLeadingZeros(text, cursorPosition) helper in js/Utils.js instead of carrying their own copy of the branch.

pages/TxKey.qml
```diff
@@ -112,19 +113,11 @@ Rectangle {
...
-                        }
-                        error = walletManager.amountFromString(text) > appWindow.getUnlockedBalance();
-                    }
+                    const corrected = Utils.stripLeadingZeros(text, cursorPosition);
+                    text = corrected.text;
+                    cursorPosition = corrected.cursorPosition;
+                    error = walletManager.amountFromString(text) > appWindow.getUnlockedBalance();
+                }
                     validator: RegExpValidator {
                     regExp: /^\s*(\d{1,8})?([\.,]\d{1,12})?\s*$/
```
jpk68, line 122: Nit: indentation

munzzyy: Fixed, thanks.

jpk68:
Please squash your commits :)

jpk68 (approved):

plowsof (approved):

PR #4696, opened 2026-09-06
src: add support for multisig

jpk68:
Adds multisignature functionality to the GUI.


jpk68:
Note: to build this, you'll need [this](https://github.com/jpk68/monero/tree/multisig-patches) checked out for the `monero` submodule. I'm not too sure why it doesn't work normally, without manually changing the branch :/

PR #4713, opened 2026-09-17
components: make standard dialog less intrusive

jpk68:
Makes `StandardDialog` less intrusive by not taking up the entire screen.

Mutually exclusive to #4702.

screen


PyXMR2025 (approved):
Please check whether dialogs containing long text have broken scrolling or any content display issues.

PR #4592, opened 2026-05-09
main: add more fiat currencies, deprecate kraken api

SNeedlewoods:
submitted by jpk68
https://codeberg.org/jpk68/monero-gui/pulls/3

(I have not reviewed this code, merely forwarding it)

PR #4714, opened 2026-09-20
main: add more fiat currency options, rework api logic

jpk68:
Continuation of #4592.


SChernykh (approved):

selsta (approved):

plowsof:
is there a simple way to make the drop down box draw upward as hack to make all the options visible when in windowed mode?

components/StandardDropdown.qml
```diff
@@ -152,72 +153,80 @@ ColumnLayout {
             width: dropdown.width
             y: head.y + head.height
```
plowsof:
```suggestion
            y: head.y - droplist.height
```

plowsof (approved):

selsta (approved):

plowsof (approved):

